PR #19, opened 2016-04-23
Add bitcube seed node

btscube:
Added bitcube seed node.


theoreticalbts:
This seed node is not accessible to me, closing.

PR #45, opened 2016-05-02
add liondani's seed node located in Switzerland

liondani:


theoreticalbts:
Seed node is up, adding to README.md.

PR #50, opened 2016-05-11
Fix out-of-source builds (Ignore this one)

arhag:
Building from a directory other than the source directory should now work with these changes.


arhag:
I'm closing this one because it is accidentally merging into master.

The corrected pull request can be found here: https://github.com/steemit/steem/pull/51

PR #52, opened 2016-05-11
Update LICENSE.md

ryanRfox:
Update LICENSE.md to meet Graphene library copyright requirement to include the MIT license within derived works.


theoreticalbts:
Closing, the license stands as-is. AFAIK Cryptonomex (as the Graphene copyright holder) signed off on the license terms of Steem, so essentially Steemit doesn't have to comply with the same requirements, or distribute Graphene code under the same license, as a member of the general public would (since Steemit has a different license agreement with Cryptonomex than the general public does).

PR #61, opened 2016-05-17
Update BUILD_OSX.md

tchap:
Add missing dependency: `libtool`


mvandeberg:
BUILD_OSX.md has since been updated and includes libtool dependency.

PR #67, opened 2016-05-25
Quick but inelegant fix for UTF8 patch bugs

arhag:
This seems to do it. No more "invalid utf8" messages when replaying blockchain.

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -4,25 +4,20 @@
 
 #ifndef IS_LOW_MEM
 #include <diff_match_patch.h>
-#include <locale>
-#include <codecvt>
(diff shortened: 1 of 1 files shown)


arhag:
Darn. It turns out this doesn't build using the MinGW-w64 compiler because it doesn't support codecvt.

A way around this is use Boost.Locale instead since Steem already has a Boost dependency.

arhag:
Okay. I switched over to Boost.Locale for my UTF8 string conversions. It now builds with both gcc and the MinGW-w64 compiler. And it runs properly (no "invalid utf8" errors) on both Linux and Windows.

I can do
```
get_state "/news/@pfunk/april-breaks-global-temperature-record-marking-seven-months-of-new-highs"
```
and I get the correctly edited result.

PR #72, opened 2016-05-30
Use old cat-parts on Windows build to avoid Python3 dependency

arhag:
Worked with bitcube to fix my changes to the build process so that Steem builds on Windows using MSVC and without requiring Python3.


arhag:
This is no longer necessary now that @btscube has included the changes in his pull request: https://github.com/steemit/steem/pull/138

Issue #88, opened 2016-06-10
Divide by zero error when paying curators


arhag:
https://github.com/steemit/steem/blob/v0.5.0/libraries/chain/database.cpp#L1635

If everyone that voted for a comment updates their votes, and no new votes are added, then `total_weight` will be 0 in the `pay_curators` function which will lead to a divide by zero error.

The function should check if `total_weight == 0` and if so, skip the while loop entirely.

Also, Line 1644 in that function also has a problem. The if statement should just read:
```
if( unclaimed_rewards.value )
```

arhag:
The bug above is not just a theoretical problem. It actually exists in the live blockchain as of block ????. I ran the my script to calculate content payouts by running v0.5.0 steemd in debug mode, and when it came time to push the July 4th block I got:
```

**Comment by arhag** [comment]
The bug above is not just a theoretical problem. It actually exists in the live blockchain as of block ????. I ran the my script to calculate content payouts by running v0.5.0 steemd in debug mode, and when it came time to push the July 4th block I got:
```

arhag:
The bug above is not just a theoretical problem. It actually exists in the live blockchain. The bug is triggered (probably from all the votes of a comment being updated) as some block between block number 1997811 and block number 2199289. 

I ran my script (pushing real blocks from the live chain up to block number 2199289) to calculate content payouts by running v0.5.0 steemd in debug mode, and when it came time to push the July 4th block I got:
```
steemapi.steemnoderpc.RPCError: 13 N5boost16exception_detail10clone_implINS0_19error_info_injectorISt14overflow_errorEEEE: Integer Division by zero.
rethrow Integer Division by zero.:
    {"what":"Integer Division by zero."}
    th_a  database.cpp:2255 _apply_block

    {"new_block":{"previous":"00218efacd69f60e9d8343a1d4dba14aeac9eb0d","timestamp":"2016-07-04T00:00:00","witness":"au1nethyb1","transaction_merkle_root":"0000000000000000000000000000000000000000","extensions":[],"witness_signature":"20775b87692cfb0dc00f81a477994f072949c91399d1aac1f04ced4e44a792450d2cc27b96d8fa406998b3d80088bab90447fab0b96746ea2bb9662a1647dad26d","transactions":[]}}
    th_a  database.cpp:508 operator()

    {"witness_owner":"au1nethyb1"}
    th_a  database.cpp:663 operator()
```

[...]

PR #91, opened 2016-06-10
Fix divide by zero in pay_curators #88

arhag:
See issue #88.


arhag:
This was fixed by https://github.com/steemit/steem/commit/2533404d9e4f42d81d6748db2dd6ff516009ee1c so this pull request is no longer necessary.

PR #170, opened 2016-07-16
Update BUILD_OSX 

metasyn:
It was a bit outdated.


metasyn:
This builds on top of #61  as well as #164, 3 pull requests in one !
@mvandeberg

ilovezfs:
I have no trouble building with the bottled libtool. What is
```
There is a known bug with libtool due to a hardcoded SED path. Fix this by uninstalling & reinstalling:
```
referring to exactly? Do you have an issue link?

metasyn:
@ilovezfs which version of OSX are you using?

I, alongside someone else, was running into this on 10.11.5. I didn't open an issue. Should I ?

https://steemit.com/osx/@metasyn/fix-steemd-on-osx-homebrew-error-fix

ilovezfs:
@metasyn I think that issue you linked to is unrelated. The path `/usr/local/Library/ENV/4.3/sed` is no longer valid. We just moved it to `/usr/local/Library/Homebrew/shims/super/sed ` three days ago: https://github.com/Homebrew/brew/commit/a02be9eea204264ce48827ac365c8b53b986d66a

ilovezfs:
Also I see the hard coded path you were talking about in the 2.4.6 bottle, but for the 2.4.6_1 bottle actually that is no longer the case:
```
iMac-TMP:~ joe$ grep 'SED=' /usr/local/Cellar/libtool/2.4.6/bin/glibtool
SED="/usr/local/Library/ENV/4.3/sed"
  SED=$func_path_progs_result
iMac-TMP:~ joe$ grep 'SED=' /usr/local/Cellar/libtool/2.4.6_1/bin/glibtool
SED="sed"
  SED=$func_path_progs_result
```

So I'd suggest you brew update and brew upgrade.

metasyn:
@ilovezfs I see - so do you think we should add brew upgrade to the instructions list then?

I'll remove the references to rebuilding libtool - thanks for the help !

bmann:
Yes, brew update / brew upgrade at the beginning of all instructions. Most people that are less common CLI users on Mac will have used brew, but it will be often out of date.

Just found this thread myself and was making notes of the stale-ness of the instructions. I hit some openssl snags, too.

PR #289, opened 2016-08-16
Easier to paste into config.ini

wmbutler:
Just thought it might be easier for users to add to the config.ini in this format.


theoreticalbts:
Superseded by #295

PR #292, opened 2016-08-16
Allow sending identical transactions

emfrias:
workaround for the restriction that we can't send two identical copies of the same transaction out during the same block due to same transaction id


theoreticalbts:
Including multiple completely identical transfers in the same block sounds an awful lot like spam. What exactly are you trying to do?

abitmore:
Want to transfer to one receiver twice with same amount and memo? You can just pack 2 operations into one transaction, or use different TaPOS headers with 2 transactions.

emfrias:
We send out Steem when someone sends us BTC.  Sometimes they will send us two identical BTC amounts in the same Bitcoin block.  We process each transaction individually, but sometimes they still end up in the same block on Steem.

We can't combine them into a single transfer of twice the amount, that would be confusing.

If we added a ton of logic on our exchange, we could combine them into a single Steem transaction, but then everyone else who wants to do this would have to implement the same logic.

Tweaking the TaPOS headers/expiration times is another hack that we considered, but the wallet would have to keep more state to pull this off and it's a rare event so the few extra bytes of custom operation don't make me lose any sleep.

mvandeberg:
I don't think this opens us up to more spam. Rate limiting is based on transaction size, so utilizing this behavior is more expensive to the user. Besides, an attacker could use the techniques abit mentioned to create unique transaction IDs.

abitmore:
@emfrias is it possible to include the incoming Bitcoin trx_id into the memo field of outgoing op? I guess the trx ids are different even if 2 transactions have same `[sender,receiver,amount]` and in same bitcoin block. And it's better for you and your customers to track.

dnotestein:
@abitmore your change would be specific to our business, whereas we're looking for a general purpose solution to what is IMO an actual flaw in the current operation of the client.

dnotestein:
just out of curiosity, was this closed as incorporated or closed as rejected?

sneak:
The `develop` branch no longer exists.  Please base PRs to `master`.

https://steemit.com/steem/@steemitblog/4b7tnh-steem-tech-advisory-repository-branches-to-be-renamed

mvandeberg:
A year old and needs rebasing.

PR #293, opened 2016-08-16
Fix error reading compression flag from config.ini

emfrias:
allow logfile gzipping of rotated-out log files to be controlled from config.ini.  

**This requires a change to FC that enables compression: https://github.com/steemit/fc/pull/4**


mvandeberg:
A year old and needs rebasing.

Issue #283, opened 2016-08-16
Create updated build instructions for Ubuntu 14.04, 16.04


theoreticalbts:
Build from scratch in a brand-new VM for each of these platforms and take careful notes about what needs to be done, then put it in root of this repo.

theoreticalbts:
When new build instructions have been written, #225 #238 should be closed.

theoreticalbts:
Notes to self:

- Build with and without `programs/build_helpers/configure_build.py`
- Remember Python is now a dependency because of `cat-parts.py`

theoreticalbts:
Added new build instructions to develop.

PR #296, opened 2016-08-17
Instructions for re-compiling after hardfork

wmbutler:


abitmore:
1. Should checkout special tag but not master branch. Sometimes master branch is not in sync (unfortunately).
2. Should add `-DCMAKE_BUILD_TYPE=Release` to `cmake` command by default

abitmore:
`--rebuild-blockchain` should be `--replay-blockchain`?

wmbutler:
I added the -DCMAKE_BUILD_TYPE but don't know what to do regarding the proper branch to check out.

abitmore:
>I added the -DCMAKE_BUILD_TYPE but don't know what to do regarding the proper branch to check out.

I don't know either.

By the way I updated the original comment, please check.

theoreticalbts:
Basically you want to put build instructions in the tree, which makes this a duplicate of #283.

PR #376, opened 2016-09-12
0.14.1 changes

mvandeberg:
Change versions for 0.14.1
Revert target votes per day change
Change hardfork 14 time to 2016-9-20 15:00:00 UTC


noisy:
what was the reason to change date of hf?

mvandeberg:
It was for today. We need to give witnesses time to review the changes and upgrade.

PR #400, opened 2016-09-14
387 one touch testing

sneak:
This change should be discussed before merge - it removes the top-level Makefile and the build products from the `.gitignore` and expects `cmake` and `make` to be run from an ephemeral `build` subdirectory.

Requesting feedback @steemit/steem-developers - would this be a major inconvenience?

It significantly simplifies building/testing scripts and processes, as well as the `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,36 +4,19 @@
 *.cmake
 CMakeCache.txt
 CMakeFiles
-Makefile
 compile_commands.json
 moc_*
 *.moc
 hardfork.hpp
 
+build/
 lcov
 
 libraries/utilities/git_revision.cpp
-
 libraries/wallet/Doxyfile
 libraries/wallet/api_documentation.cpp
 libraries/wallet/doxygen
 
-programs/cli_wallet/cli_wallet
-programs/js_operation_serializer/js_operation_serializer
-programs/steemd/steemd
-programs/steemd/test
-programs/delayed_node
-programs/build_helpers/cat-parts
-programs/size_checker/size_checker
-programs/util/get_dev_key
-programs/util/inflation_model
-
-tests/app_test
-tests/chain_bench
-tests/chain_test
-tests/plugin_test
-tests/intense_test
-tests/performance_test
 
 doxygen
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,9 @@ FROM phusion/baseimage:0.9.19
 
 #ARG STEEMD_BLOCKCHAIN=https://example.com/steemd-blockchain.tbz2
 
+ARG STEEM_BUILD_TYPE=web
+
+# install essential build-time packages and cleanup after installer
 RUN \
     apt-get update && \
     apt-get install -y \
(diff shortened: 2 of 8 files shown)


abitmore:
It's a good change IMO. I have been using out-of-source building for a long time, and it works well.

sneak:
@abitmore The new root-level Makefile has non-cmake-related stuff in it for us.  The cmake-generated Makefile goes in the build dir, and is invoked separately.  When you say overwritten, what do you mean?

abitmore:
I mean, if I run `cmake .` in the root directory, will it generate a new make file and overwrite the existed one? Sorry if it's a stupid question, I'm not so familiar with make files.

sneak:
It will, I believe; however if you're not using it, no harm.  The only issue would be if someone tries to commit that changed Makefile - which git will note it changed (before it was in `.gitignore` because I guess people are used to building in the root of the repo, so it would be ignored).

The nice thing about the *new* Makefile is `make build` in the top-level directory will make the `build` subdir and `cd` into it and `cmake ..`, so all people have to do is learn `make build` instead of `cmake . && make`.

mvandeberg:
I'm not quite sure I am ready to merge this. I want to wait until after the next release so that we can give everybody plenty of heads up about a new build process. I also want to add a few other make targets for common build configurations. (web being one of them)

theoreticalbts:
The in-tree build instructions [here](https://github.com/steemit/steem/blob/develop/doc/build-ubuntu.md) and [here](https://github.com/steemit/steem/blob/develop/doc/build-osx.md) do in-tree builds. If recommended practice is out-of-tree, those docs should be changed as part of this PR.

sneak:
@mvandeberg Could you please give me some examples of build configs for web, witness, et c?  I've just been using the ones in the doc - I'm happy to update the Makefile to support any common type of build.

mvandeberg:
Default build should be:

```
LOW_MEMORY_MODE=OFF
ENABLE_CONTENT_PATCHING=OFF
CLEAR_VOTES=ON
```

This should handle witnesses and exchanges. (exchanges will load plugins to customize their node)

Web build should be:
```
LOW_MEMORY_MODE=ON
ENABLE_CONTENT_PATCHING=ON
CLEAR_VOTES=OFF
```

Test build should be
```
LOW_MEMORY_NODE=OFF
ENABLE_CONTENT_PATCHING=ON
CLEAR_VOTES=ON
```

If a user wants to build with a different configuration they can always invoke cmake manually.

valzav:
@mvandeberg, so you are saying we can run it with LOW_MEMORY_MODE=ON on steemit.com?

mvandeberg:
Sorry, that was a mistake. Should be `LOW_MEMORY_MODE=OFF`

sneak:
is `CMAKE_BUILD_TYPE=Debug` required for tests, or no?

mvandeberg:
`CMAKE_BUILD_TYPE=Debug` is not required for tests, but we tend to use it because it usually results in a slightly faster build time by skipping optimizations. And it also emits a symbol table just in case we need gdb.

sneak:
@steemit/steem-develop-committers

Requested changes have been made, commit history squashed, tests passing.  Approval requested.

Please let me know if this requires additional work/changes for merge.

sneak:
I'm retiring this for now.  Leave the branch, I'm gonna scavenge the Makefile and osx docker build/test setup for another PR.

PR #428, opened 2016-09-16
20160913 follow

mvandeberg:
Improvements to follow plugin, blogs, feeds, and database_api get_state. This is a non-consensus change but is needed for the next Steemit.com deployment.


roadscape:
Does this new `get_discussions_by_blog ` replace `get_discussions_by_author_before_date`?

mvandeberg:
I'm not quite sure what `get_discussions_by_author_before_date` is supposed to do, so yeah. Let's say it replaces it.

roadscape:
`get_discussions_by_author_before_date` is the current method of paginating the Blog tab ;)

mvandeberg:
Ok, so yes. It is the new way!

Issue #413, opened 2016-09-15
Error message "can only update owner authority once a minute"


abitmore:
https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/steem_evaluator.cpp#L140

According to config.cpp, it should be once a hour.

mvandeberg:
Thanks. I believe it was once a minute for a small time in development and we bumped it to owner an hour.

abitmore:
Another message here: https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/steem_evaluator.cpp#L240

"operations contains options that are only updatedable on an account with no rshares"
should be "... on a comment ..."

abitmore:
More message: https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/steem_evaluator.cpp#L423, https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/steem_evaluator.cpp#L428

"The permlink of a comment cannot change"
should be "The parent permlink of ..."

abitmore:
https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/steem_evaluator.cpp#L1470

"account does not sufficient balance" is strange.

abitmore:
https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/steem_evaluator.cpp#L1610

"cannot recover an account that does not have you as there recovery partner"
should be "... as their ..."

abitmore:
https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/steem_evaluator.cpp#L1656

"there are no active recovery requests for this account"
perhaps should be "there is no active recovery request for this account"?

mvandeberg:
This was fixed some time ago.

PR #435, opened 2016-09-17
Error message adjusted to value of STEEMIT_OWNER_UPDATE_LIMIT

noisy:
`STEEMIT_OWNER_UPDATE_LIMIT` is currently set to `fc::minutes(60)`:
https://github.com/steemit/steem/blob/develop/libraries/chain/include/steemit/chain/config.hpp#L62

Error message should reflect that.


abitmore:
Thanks. Some other message errors can be found here: https://github.com/steemit/steem/issues/413

mvandeberg:
This is no longer an issue of `develop`.

PR #448, opened 2016-09-21
Add shutdown_witness command to cli_wallet

arhag:
The documentation of `update_witness` was not accurate. Passing the empty string for the `block_signing_key` argument throws the following error:

```
base58str.size() > prefix_len:
    {}
    th_a  types.cpp:26 public_key_type
```

While one could use the special signing key `STM1111111111111111111111111111111114T1Anm` in the `update_witness` command to shutdown the witness, it is far more convenient to have a dedicated cli_wallet command to do it. This pull request adds that command and it updates the documentation of `update_witness` to reflect the reality of the existing code that passing the empty string for the `block_signing_key` argument will not work.


mvandeberg:
I think we should update the existing cli command to match the documentation rather than make a new command.

mvandeberg:
Nearly 1 year old.

Issue #445, opened 2016-09-20
Content fetching APIs - minor fixes


roadscape:
1) In `get_state "@user/posts"`, comment keys are returned in account.posts without '@author/' prefix, which is not consistent with how blogs and feeds are returned (as of late).

2) Loading of "recent replies" seems to be broken (after the 1st page). Here's the method: [get_replies_by_last_update(start_parent_author, start_permlink, limit)](https://github.com/steemit/steem/blob/76592f04af4a46bce9e8ad2d84020f7ed2b258c0/libraries/app/database_api.cpp#L1013)

Locally, I got the method to work (seemingly) as expected by making the following change:

```
-   bool filter_by_parent_author = true;
+   bool filter_by_parent_author = false;
```

theoreticalbts:
@mvandeberg if this is merged, can this ticket be closed?

PR #449, opened 2016-09-21
Posts in "@user/posts" are now stored as author/permlink.

mvandeberg:
Fix and document get_replies_by_last_update
#445

### DIFF
--- a/libraries/app/database_api.cpp
+++ b/libraries/app/database_api.cpp
@@ -1012,6 +1012,7 @@ vector<discussion> database_api::get_content_replies( string author, string perm
  */
 vector<discussion> database_api::get_replies_by_last_update( string start_parent_author, string start_permlink, uint32_t limit )const
 {
+   FC_ASSERT( limit <= 100 );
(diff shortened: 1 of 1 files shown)


theoreticalbts:
- Needs `FC_ASSERT( limit <= k );` for some fixed `k` (maybe 1000?)
- Should explicitly handle the case where both strings are empty (i.e. add `else` clause after `else if`)

mvandeberg:
`k` should be `100` as used by `discussion_query`

``` c++
struct discussion_query {
   void validate()const{
      FC_ASSERT( filter_tags.find(tag) == filter_tags.end() );
      FC_ASSERT( limit <= 100 );
   }

   string           tag;
   uint32_t         limit = 0;
   set<string>      filter_tags;
   optional<string> start_author;
   optional<string> start_permlink;
   optional<string> parent_author;
   optional<string> parent_permlink;
};
```

The else case is handled implicitly by 
``` c++
auto itr = last_update_idx.begin();
```

I can add an else case

```
else
   itr = last_update_idx.begin();
```

But that seems redundant...

PR #460, opened 2016-09-23
Implement JSON database schema

theoreticalbts:


theoreticalbts:
Bump fc and try again

jcalfee:
Very interesting!  What is this ??

theoreticalbts:
@jcalfee The database can output a JSON blob which fully describes the type of everything.  To anticipate your next question, we should be able to move the functionality of the js_operation_serializer to an entirely JavaScript based implementation which takes this JSON blob as input.

In fact you should be able to write generic code which runs on the client and parses the JSON blob, without ever generating code at all. I'm not sure what performance would be like, or how much refactoring of supporting code would be needed. How much the generic code could mimic the current generated code's interface depends a lot on the specifics of the generated code's interface and how dynamic the JavaScript language is.

Issue #458, opened 2016-09-22
Protocol Library


mvandeberg:
Separate chain/protocol into its own protocol library.

The protocol library should contain code pertaining to the blockchain protocol (blocks, transactions, operations, etc).

The chain library should contain code pertaining to the in memory representation of the chain state (database objects, evaluators, etc).

faddat:
Hot.  When that refactor is done I will use this:

https://cgogen.com/

For ultimate golang binding success.  For the time being, I am going to pull away from github.com/go-steemit/rpc and stick with net/http for steemjson.  How can I help with this, and what is the best dev env setup for hacking on steem directly?

sneak:
We're testing on Ubuntu 16.04 primarily right now.

faddat:
What do you use in terms of an IDE, etc?  Basically I want to mirror your configuration in my env so that I can pickup on the memes more easily.

mvandeberg:
@faddat I think you have misinterpreted the purpose of this ticket. On master there is a folder `libraries/chain/protocol` that contains code pertaining to the blockchain protocol. `libraries/chain` pertains to the in memory representation of the chain state. We are moving `libraries/chain/protocol` to become `libraries/protocol` and ensuring it is its own compilation unit. This is a good litmus test of our encapsulation of different aspects of steemd. Essentially, `libraries/protocol` would need to be replicated almost exactly in order for another binary to be able to understand the steem blockchain, but you could substitute out `libraries/chain` for any other backend representation of the chain state (SQL, for example) and everything would continue to work. When the refactor is done, nothing will have actually changed, except for the internal structure of the source code.

None of us use IDEs. We each use a text editor of our choice and bash. gdb when it is needed.

faddat:
I certainly did misinterpert.  

I really appreciate your taking time to explain the layout.  I've been reading through the source code, and hope to understand it well enough to make some contributions in the future.  My big hope for Steem's economic future is that we are able to expand the developer ecosystem, and grpc/swagger (probably grpc, not swagger ,as from what I've gleaned so far, grpc requires fewer changes to be made to the code base than swagger would) API support would be pretty enormous.  Coding against STEEM in golang has been quite challenging because I just don't know every single possible return.  I've been trying to find a place in the source code where I can find this but haven't yet authoratitively done so.  I did find this:

```go
func (op *VoteOperation) MarshalTransaction(encoder *transaction.Encoder) error {
	enc := transaction.NewRollingEncoder(encoder)
[...]

Issue #444, opened 2016-09-20
gRPC or Swagger API Provider plugin for generated api clients


faddat:
Basically, I'm writing to ask if it's possible.  

If it is possible, and anytone will give me pseudocode and/or advice that can get me started, then I'm happy to do the work of implementing this.  We as a community / economic entity need this **badly**. Both gRPC and swagger are designed to take the time burden off of development teams and to make well-documented interfaces something that's both standard and easy.  

 Off the top of my head, 100% of devs I know working on steem projects think the API is a weak point, and have complained to me about it at length, just as I have complained to them about it as well.  I'm very happy to help change that, provided that anyone can:

1) Let me know if I am on the right track
2) Let me know if this can be implemented through a plugin
3) Tell me about the dev env where steem what tools and processes are in use, so that I might mirror them on my own system and enjoy the benefits of sharing the procedural idioms.   

To clarify:  I don't need pseudocode for the grpc / swagger part, I've checked out their docs and despite my relative lack of experience in c++, I'm pretty sure that I can handle this.  I'd just need some pointers on how to efficiently interface with steemd via a plugin, and if I'm to undertake this then I definitely want to be able to configure my dev setup the same way as core team members so that we have shared context.  

Another point that is probably worthy of mention here is that gRPC uses protocol buffers, meaning that we could reasonably expect to enjoy a speed-boost in communication between the blockchain and its clients by implementing gRPC.

abitmore:
I have little to add to this discussion, it seems https://github.com/steemit/steem/issues/347 is related?

faddat:
It might as well be #347 , sorry about that, I think I'd forgotten it existed.   I'll close that, as I feel this one better describes the situation.

theoreticalbts:
I'm not familiar with gRPC or Swagger.

There are two different things you might want to do:

- (a) If you think the layout of methods is poorly designed, illogical, or don't expose needed functionality, and you want to add new methods and make them available over the existing HTTP and Websocket RPC's, you can already do so in a way that's well-supported and used by many of our own plugins.

- (b) If you want to make the existing methods available over an alternate wire protocol, that's more tricky.

For the rest of this comment, I'll be talking about option (b).

The simplest way to do this would be a proxy which translates requests and responses -- a script in Go or Python or something that talks gRPC or Swagger to downstream clients, and Websocket to upstream steemd.  It lives out-of-process and would require no patches to steemd unless bugs are found -- as far as we're concerned, you'd just be another client.

[...]

PR #461, opened 2016-09-23
Refactor to create steemit protocol library #458

mvandeberg:
Refactors chain/protocol into its own protocol library. Also moves some files from chain into protocol that are used for the protocol. #458


faddat:
Do I understand correctly that this could be a step towards #444 ?  

I think that it's also likely to help me implement:

https://cgogen.com/

For cgogen, I know that this is often best run in the testing stage, so I may send a PR that gets that into the tests, as long as I get the A-OK.

theoreticalbts:
Two minor nitpicks which can easily be fixed in this PR:

- Double slash in `libraries/app/include/steemit/app/api.hpp`
- compound.hpp should be in chain, not protocol

One slightly larger issue that I will file a ticket for, because it is non-trivial and will take a little time to do, and we don't want to block merging this PR:

- Split config.hpp into constants needed to validate operations (in protocol) and constants not needed to validate operations (in chain)

Issue #465, opened 2016-09-26
Fix Validation of Authorities


mvandeberg:
Posting authority is not validated in account_update and account auths do not check that accounts exist.

xeroc:
If you add a `hardfork 15`-soft-fork .. why is the tag called 0.14.3?

mvandeberg:
0.14.3 is an intermediate upgrade. Working on a hardfork change soon.

mvandeberg:
Fixed in 0.15.0

PR #466, opened 2016-09-26
Validate posting auth and check for account auth existence as a soft …

mvandeberg:
…fork #465


theoreticalbts:
It gets my approval, but we should make sure we can replay before release.

PR #475, opened 2016-09-30
dos fix per dan - stop returning results after 20ms runtime

sneak:


theoreticalbts:
I'm checking whether excessive iterations in this loop are in fact causing the problem.

theoreticalbts:
Too many iterations in this loop are in fact the problem, but the above patch is sub-optimal because it makes the behavior non-deterministic -- it depends on the specifics of timing.

theoreticalbts:
Force-pushed new patch which doesn't depend on timing and logs useful information about failures.

theoreticalbts:
Updated to do 10x as many

PR #478, opened 2016-09-30
Update README.md

faddat:
This re-organizes the readme.md a little bit, and should be considered a work in progress towards a readme.md that explains STEEM somewhat better than the original while retaining the original intent of conveying that the code itself serves as the documentation of STEEM.


sneak:
superseded by #487

PR #488, opened 2016-10-04
does not run tests as part of main docker build

sneak:
Now that the `Dockerfile` is building both a web node and a p2p node type, the builds on Docker Hub are timing out. This removes the full test build/run in the Dockerfile (which was only there as a sort of belt-and-suspenders approach, as we do the exact same tests before merging to `develop` or `master` anyway).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,29 +30,6 @@ ADD . /usr/local/src/steem
 RUN \
     cd /usr/local/src/steem && \
     git submodule update --init --recursive && \
-    rsync -a \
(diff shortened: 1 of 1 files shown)


sneak:
@steemit/steem-develop-committers ping

theoreticalbts:
Rebase these into a single commit and I'll merge it.

sneak:
Github lets you squash them on PR merge, afaik.  Rebasing as requested.

sneak:
ping @theoreticalbts

PR #496, opened 2016-10-10
Code Changes that are compatible with VS2013

alphab3ta:
@abit thanks for your help. 
Here are 2 additional changes needed inside the ***fc*** folder in order to get rid `error C2593: 'operator <' is ambiguous`

 add line to **libraries\fc\include\fc\network\ip.hpp**
` friend bool operator<( const address& a, const address& b ); `

and add line to **libraries\fc\src\network\ip.cpp**
`  bool operator<( const address& a, const address& b ) {
    return uint32_t(a) < uint32_t(b);
  }`

### DIFF
--- a/libraries/chain/CMakeLists.txt
+++ b/libraries/chain/CMakeLists.txt
@@ -42,7 +42,7 @@ add_library( steemit_chain
              block_database.cpp
 
              ${HEADERS}
-             ${hardfork_hpp_file}
+ #            ${hardfork_hpp_file}
              ${PROTOCOL_HEADERS}
              "${CMAKE_CURRENT_BINARY_DIR}/include/steemit/chain/hardfork.hpp"
            )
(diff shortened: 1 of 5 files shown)


alphab3ta:
* With all the code change above. I encountered two errors when I try to compile the latest commit with vs2013. Can someone see what is causing the error? Thanks.

>F:\Git\steem\libraries\fc\include\fc/reflect/variant.hpp(53): error C2668: 'fc::from_variant' : ambiguous call to overloaded function
          F:\Git\steem\libraries\chain\include\steemit/chain/protocol/operations.hpp(95): could be 'void fc::from_variant(const fc::variant &,steemit::chain::operation &)'

>F:\Git\steem\libraries\fc\include\fc/io/raw.hpp(292): error C2668: 'fc::raw::unpack' : ambiguous call to overloaded function
          F:\Git\steem\libraries\fc\include\fc/uint128.hpp(131): could be 'void fc::raw::unpack<fc::datastream<const char *>>(Stream &,fc::uint128 &)'
          with
          [
              Stream=fc::datastream<const char *>
          ]

PR #487, opened 2016-10-04
refactored documentation

sneak:
- consolidated unix build instructions
- factored seednodes out into parseable text file
- cleaned up frontpage repo `/README.md` for clarity
- added docker quickstart examples
- wrapped markdown lines to reasonable width

PR #502, opened 2016-10-12
Update README.md

faddat:
Removed outdated list of seed nodes and added link to status.steemnodes.com


sneak:
We will always ship seed nodes with the codebase.  This refactor is already underway, please see #487.

Issue #510, opened 2016-10-14
Review compile-time constants


theoreticalbts:
- Make sure all testnet-specific code tests constant `IS_TEST_NET`
- Make sure all testnet-specific code uses `#if[n]def` or `#if defined()`
- Make sure Dockerfile explicitly specifies all "user-facing" configuration parameters
- Document "user-facing" configuration parameters

theoreticalbts:
The following compile-time constants seem to be related to the p2p code.  I don't think any of them should be regarded as "user-facing" so I won't attempt to document them:

- `DEFAULT_LOGGER`
- `ENABLE_P2P_DEBUGGING_API`
- `P2P_IN_DEDICATED_THREAD`
- `USE_PEERS_TO_DELETE_MUTEX`
- `VERBOSE_NODE_DELEGATE_LOGGING`

Does our p2p guru @emfrias have any opinion on the matter?

theoreticalbts:
The remaining options were already documented as part of #487.  That means less work for me!

I'm removing all traces of `ENABLE_CONTENT_PATCHING` from the CMakeLists, Dockerfile and docs.  It is not actually used, content patching is always enabled [1].

[1] Except when `LOW_MEMORY_NODE` is set.  Since `LOW_MEMORY_NODE` throws away comment bodies, content patching doesn't even make sense in the context of `LOW_MEMORY_NODE`.

theoreticalbts:
Closing because it was merged.

PR #512, opened 2016-10-14
Update compile-time constants

theoreticalbts:
Discussion in #510


sneak:
To verify, `ENABLE_CONTENT_PATCHING` isn't a thing anymore?

(Please also update the `Dockerfile.test` too any time the `Dockerfile` is updated. It sucks that we have to violate DRY but otherwise tests take too long.)

Issue #519, opened 2016-10-19
Allow debug_node_plugin to mine blocks without editing database


theoreticalbts:
This is some additional functionality for debug_node_plugin for use with some testnet scripts I'm developing.

theoreticalbts:
Relevant code is mostly in ea07bbe45e8ecf004a261e406e5132045534d3ae

theoreticalbts:
Closing, this PR was merged a while ago.

PR #520, opened 2016-10-19
Allow debug node to generate blocks without editing database

theoreticalbts:
PR for #519


mvandeberg:
I created an issue to take of the problem mentioned above. #526

Issue #558, opened 2016-11-07
Transparent dmca claims


roadscape:
Operators of public nodes/APIs need a transparent way to voluntarily handle DMCA claims.

All claims should be published publicly as posts on the blockchain. The format might be:

 - 'Decline Payout' enabled (`max_accepted_payout = 0.000 SBD`)
 - category: `dmca-claim`
 - `json_metadata` contains a `posts` key with an array of affected `author/permlink`s

The plugin would honor the claim only after a specific reviewer has upvoted it. So the config.ini should accept a list of 'reviewers' whose votes to watch. When offending posts are requested from the API, their content should be replaced with a link to the relevant claim.

roadscape:
![](https://github.trello.services/images/mini-trello-icon.png) [DMCA Handling](https://trello.com/c/IDtMwGcF/103-dmca-handling)

faddat:
....but is this the memory hole?

mvandeberg:
What do you mean by memory hole?

faddat:
Ever read 1984?

Jacob Gadikian
E-mail: faddat@gmail.com
SKYPE: faddat
Phone/SMS: +84 167 789 6421

abitmore:
IIRC there will be a link to the claim showing on the "removed" post. "This content is removed due to this claim."

mvandeberg:
This will not be implemented on the block chain.

PR #561, opened 2016-11-08
Implement DCMA takedown plugin #558

bytemaster:
This pull request will look in the json_metadata of comments for

```
{ "author_permlinks" : [
  ["author1","permlink1"],
  ["authro2","permlink2"]  
 ]
}
```

There is a command line argument --follow-dmca-authority which allows one or more account names to be authorized to take down content.  

If an authorized author makes (or reblogs) a post in the "dcma-takedown" category then the json_meta field will be processed.

The content of all posts specified in the "author_permlinks" field will be changed to a relative link to the takedown request.

### DIFF
--- a/libraries/plugins/follow/follow_evaluators.cpp
+++ b/libraries/plugins/follow/follow_evaluators.cpp
@@ -72,7 +72,10 @@ void reblog_evaluator::do_apply( const reblog_operation& o )
    {
       auto& db = _plugin->database();
       const auto& c = db.get_comment( o.author, o.permlink );
+
       FC_ASSERT( c.parent_author.size() == 0, "Only top level posts can be reblogged" );
+
+      _plugin->takedown( c );
 
       const auto& reblog_account = db.get_account( o.account );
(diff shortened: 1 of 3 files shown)


faddat:
Can you share with us weather steemit has recieved such takedown notices?  

And furtherwhat effect this has on the chain itself?  Walking backwards through the chain, will the original commits remain or is this "the memory hole"?

mvandeberg:
This is a non-consesus plugin that would allow a node to choose not to serve certain content. The data still exists on the blockchain, but will prevent it from being returned from any API call.

faddat:
Huh.

I understand.  :(

Still, better than one of our witnesses suffering legal issues.

mvandeberg:
This will only effect nodes that choose to run the plugin and have the correct configuration.

abitmore:
@faddat see https://github.com/steemit/steemit.com/issues/525

theoreticalbts:
Does not pass review, please fix typo DCMA -> DMCA wherever it occurs.

mvandeberg:
We are not adding more plugins to steemd at this time.

PR #564, opened 2016-11-09
Improve some comments

Alex-GR:
One of the most weird comments I've seen is the "Changing your vote requires actually changing you vote."... this usually happens, when I've already voted through the feed but I also have a tab open to read the article. If I've already clicked from the feed, and then re-click from inside the tab, I get that one. But it's a weird message and that's why I think it should change.

Going over the messages I saw a few that could be improved in syntax/grammar or first capital letter (makes things look more professional compared to all-lower capital letters).


mvandeberg:
Merging this to a branch instead of develop. There are some good changes but I want to do one more pass over the proposed error messages.

PR #577, opened 2016-11-17
Fix get_followers and get_following API calls in the follow plugin

mvandeberg:
Fixes a bug in the follow API introduced when the follow objects were updated to use account IDs instead of account names.


roadscape:
There's still something up with the call:

image

It's returning records now but the values are ints

theoreticalbts:
This doesn't pass review, the logic in this PR is incorrect.  If we consider Fred is followed by Alice, Bob, Cara and Dan with ID's like this:

```
Fred : 9
Alice : 1
Bob : 3
Cara : 2
Dan : 4
```

The problem is you mix traversals of followers by ID with the `following_follower` index, and followers by name with the `by_name` index.  So you could e.g. start at Alice, then walk to Cara by ID, then walk to Dan by name, then reach the end of the follower list by ID -- but you skipped Bob!

PR #582, opened 2016-11-18
Multiprocess

mvandeberg:
Adds proper read/write semantics for operating multiple steemd instances with the same memory mapped file.

A read node can be created with the `--read-only` command line argument.


theoreticalbts:
Changes to tags_plugin.cpp need to be reverted.  Plugin indexes are initialized in plugin_initialize(), not the constructor.

mvandeberg:
Closing this in favor of merging #587

Issue #591, opened 2016-11-22
steemd --version


Riverhead:
As near as I can tell there isn't a way to know which version of steemd you have without connecting a wallet to a running node. A simple --version flag would be great.

BhuzOr:
There is not a way even with a connected wallet really. 
The "about" cli command shows the version of the cli_wallet and not the one of steemd. 

So a --version flag for steemd really seems needed imo

mvandeberg:
merged

PR #600, opened 2016-11-22
‘steemd --version'  displays version info and exits. ‘about’ in the c…

mvandeberg:
…li-wallet now queries server version info as well. #591


theoreticalbts:
- Use underscores instead of dashes for server_blockchain_version, server_steem_revision, server_fc_revision
- For --version output say "steem_blockchain_version   :", "steem_git_revision    :", "fc_git_revision    :" (i.e. we make this output more script-friendly)

Issue #562, opened 2016-11-08
Unable to cancel a power down when amount of VESTS is less than 10x account_creation_fee


abitmore:
As described in this post: https://steemit.com/steemit-bug/@acidyo/a-bug-when-powering-down-cancelling-power-down :
> He was over 100 SP when he powered down but can't cancel the power down now even though he invested more and is at 230 SP now.

(current account creation fee is 30.000 STEEM)

mvandeberg:
Merged

PR #604, opened 2016-11-23
Allows accounts to cancel a power down when they control less than 10…

mvandeberg:
…x the account creation fee of VESTS #562


theoreticalbts:
Should this be >0 [here](https://github.com/steemit/steem/pull/604/files#diff-04f6af9cdbf180a31ccdbf6335b8f240R767)?

PR #621, opened 2016-11-28
Increment bandwidth and transactions in blockchain statistics plugin.

NateBrune:
I worked with @arhag to find and correct some bugs in the blockchain statistics model over the weekend. The noticeable differences are 

1) The bucket that has an id of zero will now have it's statistics updated.
2) If `_maximum_history_per_bucket_size` was equal to 0, an error would have occured at line 325.


mvandeberg:
We aren't using this plugin and the bandwidth model it is using is outdated.

Issue #612, opened 2016-11-26
Network crosstalk


Riverhead:
With the release of 0.16.0rc3 a node mostly just reports crosstalk from other networks, currently golos. It is  impossible to debug node issues since nearly all messages are errors from excluded crosstalk blocks.

This is a request to have a node block, exclude, or otherwise suppress messages from other networks.

BhuzOr:
Would be nice to also fix the root cause #535

mvandeberg:
We have a plan to fix the issue, but it will take some time to upgrade everyone and deploy. The problem right now is that neither network are reporting their blockchain. We have to add that back into the protocol and then at a later date start requiring nodes be on the same network.

mvandeberg:
merged

PR #625, opened 2016-11-28
Add chain_id reporting to network hello message #612

mvandeberg:
#612


theoreticalbts:
Let's make sure you can still connect to the network when running this patch.

PR #642, opened 2016-11-30
Print free memory

### LINKED ISSUE #635: How to monitor free space of a ChainBase memory mapped file?
It's good to monitor how much space can be allocated in the file, so we can be prepared to upgrade when needed.

theoreticalbts:
Fix #635


abitmore:
Possible issues of the implementation:
1. when starting the node, if free memory is less than 2GB, `_last_free_gb_printed` will keep `0` and no message will be printed ever.
2. assume not in case of above situation, it will print only one message when free memory dropped below 1GB, which will likely get ignored. I think it's better to print more after that.
3. default console log level is `warn`, using `ilog` here doesn't help much

Issue #644, opened 2016-12-01
Unnecessary double application of follow operation


theoreticalbts:
[This entire visitor method](https://github.com/steemit/steem/blob/v0.16.0/libraries/plugins/follow/follow_plugin.cpp#L149-L178) is redundant, it duplicates the generic logic implemented [here](https://github.com/steemit/steem/blob/v0.16.0/libraries/chain/steem_evaluator.cpp#L1334-L1354) and [here](https://github.com/steemit/steem/blob/v0.16.0/libraries/chain/include/steemit/chain/generic_custom_operation_interpreter.hpp#L70-L84).

mvandeberg:
The json format of custom ops changed. Originally follow has a single operation. The json contained just that operation. Now it contains the static variant of operations defined by the plugin.

PR #645, opened 2016-12-01
Remove duplicate custom dispatch logic in follow_plugin

theoreticalbts:
See #644


mvandeberg:
This change breaks backwards compatibility of the follow plugin.

Issue #501, opened 2016-10-12
Default impacted behavior ignores accounts whose posting authorities were required


arhag:
See the code here: https://github.com/steemit/steem/blob/v0.14.2/libraries/app/impacted.cpp#L45-L46

It seems it needs an additional line like:

```
op.get_required_posting_authorities( _impacted );
```

Currently without it, `custom_json` operations that only had a non-empty `required_posting_auths` do not show up in the account history, which includes follows and reblogs.

mvandeberg:
This was merged.

PR #649, opened 2016-12-02
Add get_ops_in_block method to database_api

arhag:
This pull request adds a new method to database_api which, if full account history is enabled for all accounts, allows the caller to get an ordered sequence of both real and virtual operations within a particular block (as well as location metadata attached).  This method has been useful for sites like steemdb.com.

The pull request also fixes [this bug](https://github.com/steemit/steem/issues/501) which prevents custom operations that only required posting authority from showing up in the account history.


aaroncox:
For anyone seeking to test this and in need of a block number: `479660` contains a `fill_vesting_withdraw` and a `transfer`, making it ideal for testing the `virtual_only` flag.

mvandeberg:
In the future please branch issue branch on develop, not master.

Issue #648, opened 2016-12-02
lookup_accounts call returns incorrect results


roadscape:
```
> lookup_accounts("", 10)
a-0 a-1 h20 hd0 lc0 le0 ob1 pr0 rp0 vc1

> lookup_accounts("vc1", 10)
ah1 at1 cm1 hr1 jr1 qk1 sp1 ss1 vc1 vl1

> lookup_accounts("vl1", 10)
at1 cm1 dw1 hr1 jr1 ov1 sp1 ss1 v22 vl1

> lookup_accounts("vl1", 10)
at1 cm1 dw1 hr1 jr1 ov1 sp1 ss1 v22 vl1
```

mvandeberg:
Merged

PR #652, opened 2016-12-02
Fix account name index #648

mvandeberg:
PR for #648

Account names were indexed on the comparator `std::less< std::pair< uint64_t, uint64_t > >` which yielded a different sort order than `std::less< string >`. This replaces the default comparator with a custom comparator that has the correct behavior and overloads it such that we can index account names by a string as well.


theoreticalbts:
AFAICT we never iterate by this index so we should be OK to do this without any hardfork.

Issue #666, opened 2016-12-06
static_variant does not contain a value of type steemit::protocol::pow2


Gandalf-the-Grey:
```
2106888ms th_a       database.cpp:929              notify_post_apply_op ] Caught exception in plugin: 10 assert_exception: Assert Exception
static_variant does not contain a value of type steemit::protocol::pow2
    {"t":"steemit::protocol::pow2"}
    th_a  static_variant.hpp:305 get
2109363ms th_a       database.cpp:929              notify_post_apply_op ] Caught exception in plugin: 10 assert_exception: Assert Exception
static_variant does not contain a value of type steemit::protocol::pow2
    {"t":"steemit::protocol::pow2"}
    th_a  static_variant.hpp:305 get
```

pfunks:
I'm seeing this message simultaneously on a couple nodes that aren't running plugins. Node running only the witness plugin does not show the exception message.

mvandeberg:
Merged

PR #669, opened 2016-12-06
Fix error message caused by account_by_plugin when processing equihas…

mvandeberg:
…h pow #666


theoreticalbts:
I think it's better to do it generically, I wrote it in #672.  This PR should be closed if #672 is merged.

PR #672, opened 2016-12-07
Generically unbox pow2 static_variant

theoreticalbts:
Generic solution to #666.  In contrast to #669 this solution uses the same logic for both cases and will continue to work with any future work data structure, so long as it has a `work_input` field.

### DIFF
--- a/libraries/plugins/account_by_key/account_by_key_plugin.cpp
+++ b/libraries/plugins/account_by_key/account_by_key_plugin.cpp
@@ -75,6 +75,17 @@ struct pre_operation_visitor
    }
 };
 
+struct pow2_work_get_account_visitor
+{
+   typedef const account_name_type* result_type;
+
+   template< typename WorkType >
+   result_type operator()( const WorkType& work )const
+   {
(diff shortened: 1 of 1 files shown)


liondani:
I made the changes that are described here but I still get the same red messages after I replied the blockchain....

1203082ms th_a       database.cpp:929              notify_post_apply_op ] Caught exception in plugin: 10 assert_exception: Assert Exception
static_variant does not contain a value of type steemit::protocol::pow2
    {"t":"steemit::protocol::pow2"}
    th_a  static_variant.hpp:305 get
1206048ms th_a       database.cpp:929              notify_post_apply_op ] Caught exception in plugin: 10 assert_exception: Assert Exception
static_variant does not contain a value of type steemit::protocol::pow2
    {"t":"steemit::protocol::pow2"}
    th_a  static_variant.hpp:305 get

mvandeberg:
I am not experiencing this problem. Did you try applying the fix yourself or are you running `master`? (note: `master` is now the defunct `develop` and `stable` is what `master` was)

Issue #676, opened 2016-12-07
`author_reward` virtual operation has a wrong permlink type


Netherdrake:
The bug was introduced with following commit in `steem_virtual_operations.hpp`:
https://github.com/steemit/steem/commit/184c75c6#diff-89f23fbe8a808e8e2cd84ea68168446dR16

```
   struct author_reward_operation : public virtual_operation {
      author_reward_operation(){}
      author_reward_operation( const account_name_type& a, const account_name_type& p, const asset& s, const asset& st, const asset& v )
         :author(a), permlink(p), sbd_payout(s), steem_payout(st), vesting_payout(v){}

      account_name_type author;
      account_name_type permlink; // <- wrong type
      asset             sbd_payout;
      asset             steem_payout;
      asset             vesting_payout;
   };
```

Should be:
```
   struct author_reward_operation : public virtual_operation {
      author_reward_operation(){}
      author_reward_operation( const account_name_type& a, const account_name_type& p, const asset& s, const asset& st, const asset& v )
         :author(a), permlink(p), sbd_payout(s), steem_payout(st), vesting_payout(v){}

      account_name_type author;
      string permlink;  // <- right type
      asset             sbd_payout;
      asset             steem_payout;
      asset             vesting_payout;
   };
```

The effect of this bug is that the permlinks on author rewards are formatted as usernames, and thus rendered useless.

![issue](https://cloud.githubusercontent.com/assets/3516903/20982245/0136ae12-bcb8-11e6-876e-f3c3aaadd41d.png)

sneak:
Thanks for this - both a bug report and a fix!

mvandeberg:
Merged

PR #677, opened 2016-12-07
fixed issue #676 - author_reward permlink is now a string

Netherdrake:
Fixes https://github.com/steemit/steem/issues/676


Netherdrake:
@mvandeberg fixed

PR #679, opened 2016-12-08
675 reduce chain mem

mvandeberg:
PR for #675

Removes block data redundancy between chain state and block log. Blocks that are irreversible are removed from the chain state cache and are read from the block log if needed.

Issue #675, opened 2016-12-07
Don't keep all blocks in memory


theoreticalbts:
Currently a fair amount of the memory usage of steemd is due to keeping all the blocks in memory.  It's fairly low-hanging fruit to take this data out of memory and read from the block log instead.

mvandeberg:
I have been working on this a little.

BhuzOr:
#650

arhag:
Why are you storing blocks into ChainBase at all in the first place? Even if you remove that data later, that is still causing additional pages of memory to become dirty unnecessarily, which may lead to additional IO. There is no need for the `packed_block` field in the `block_stats_object`. Blocks newer than the last irreversible block can be retrieved from `_fork_db` in the `fetch_block_by_number` function.

mvandeberg:
Merged

PR #689, opened 2016-12-09
Explicitly name sentinel value

### LINKED ISSUE #606: debug_node wants but does not get sentinel value for reading blocks
Warning with `~0` [here](https://github.com/steemit/steem/blob/61167d23b60670190f9af46a446feabe5a067274/libraries/plugins/debug_node/debug_node_api.cpp#L194) led me to look at this code in `debug_node`.  It seems to be expecting to get `~0` as a sentinel value for "I couldn't read a block" but the corresponding code in `block_log.cpp` [here](https://github.com/steemit/steem/blob/61167d23b60670190f9af46a446feabe5a067274/libraries/chain/block_log.cpp#L192-L201) cannot produce such sentinel value.  This issue merits further investigation.

theoreticalbts:
Fixes #606

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -204,7 +204,7 @@ namespace steemit { namespace chain {
    {
       optional< signed_block > b;
       uint64_t pos = get_block_pos( block_num );
-      if( ~pos )
+      if( pos != npos )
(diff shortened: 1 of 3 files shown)


mvandeberg:
Was merged into #675, closing in favor of PR #679

Issue #673, opened 2016-12-07
STEEMIT BUG DEC 6th (TAGS NOT WORKING IF USEING 5 TAGS)


jacobcards:
Hey guys I came across a bug today and its happening with many users not just myself. If you post a post with specifically "5 tags". Your post will only show up under the first tags topic. This bug does not apply if you use 4 or less tags. Here is the post on it. Thanks.
https://steemit.com/steemit/@jacobcards/possible-hardfork-bug-tags-not-working

arhag:
The bug is in the tags plugin.

If the post/comment is `safe_for_work`, i.e. [this branch of the if statement](https://github.com/steemit/steem/blob/v0.16.0/libraries/plugins/tags/tags_plugin.cpp#L204-L205) is reached, then the size of `lower_tags` will increase by 1, assuming there were no empty string tags in the `tags` field of the `json_metadata`. (By the way, that is another bug that can allow NSFW content with negative `net_rshares` show up in the default feed. You really should filter out any empty string tags [here](https://github.com/steemit/steem/blob/v0.16.0/libraries/plugins/tags/tags_plugin.cpp#L192).) So if there were 5 unique non-empty tags specified for the post/comment, `lower_tags` would actually have a size of 6 by [this line](https://github.com/steemit/steem/blob/v0.16.0/libraries/plugins/tags/tags_plugin.cpp#L207), which would mean that [this condition](https://github.com/steemit/steem/blob/v0.16.0/libraries/plugins/tags/tags_plugin.cpp#L209) would incorrectly be true, thus causing, via [this line](https://github.com/steemit/steem/blob/v0.16.0/libraries/plugins/tags/tags_plugin.cpp#L212), the post/comment to show up in the default feed and under the tag for its category but no where else.

sneak:
confirmed by @NateBrune

theoreticalbts:
@mvandeberg this patch removes NSFW flagging from the backend, has someone from the frontend team confirmed that the frontend handles this now?

RenaudGagne:
So if I understand properly, nsfw tag won't show up on the public page anymore?

theoreticalbts:
To keep NSFW off frontpage, `||` [here](https://github.com/steemit/steem/commit/62b7a0daa42e26e22acef263b4f700111162a5d4#diff-29ef2d1f0052c3113dbcb2162f9b39b8L200) should be `&&`

mvandeberg:
[...]

PR #705, opened 2016-12-14
673 tags filtering

mvandeberg:
PR for #673

Correctly categorizes posts into 5 tags regardless if which tag is the "category" (parent_permlink).

Removes NSFW filtering from the global ranking to defer filtering to UI.

### DIFF
--- a/libraries/plugins/tags/tags_plugin.cpp
+++ b/libraries/plugins/tags/tags_plugin.cpp
@@ -189,7 +189,8 @@ struct operation_visitor {
       }
 
       set< string > lower_tags;
-      meta.tags.insert( fc::to_lower( to_string( c.category ) ) );
+      if( c.category != "" )
+         meta.tags.insert( fc::to_lower( to_string( c.category ) ) );
 
       uint8_t tag_limit = 5;
       uint8_t count = 0;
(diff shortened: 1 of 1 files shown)


arhag:
@mvandeberg: I want to make sure you realize that the 0a4e9f7 commit changes the behavior of the main feeds. Before a post that wasn't tagged "spam", "test", or "nsfw" could potentially show up in the various feeds (e.g. hot) even if it had negative `net_rshares`. Now, such a post will suddenly disappear from the feeds if even a minnow downvotes it and causes `net_rshares` to even become slightly negative. I believe it will still be visible in the particular tag feeds and the personal feed of followers, but making the post far less visible makes it unlikely for other benevolent users to find the post and upvote it (assuming it is deserved) to make its `net_rshares` positive again. Perhaps filtering based on `net_rshares` being below a threshold should be done at the frontend level like with NSFW posts.

mvandeberg:
Good point. The change from or to and was to prevent nsfw content from showing up when it was upvoted. But since we are doing that all on the front end, there is little (if any need for that behavior).

mvandeberg:
The point is if spam gets positive votes it needs to show up so it can be down voted. If I wanted to siphon funds I would make spam posts, up vote them, and wait because no one would down vote them.

PR #708, opened 2016-12-14
Jenkinsci

jredbeard:
This updates the tag stable to be 'latest' on dockerhub, removes exitted containers from the CI server, and updates the build context to report from 'jenkins-ci-steemit' instead of 'default'.

### DIFF
--- a/ciscripts/buildpending.sh
+++ b/ciscripts/buildpending.sh
@@ -2,5 +2,6 @@
 curl -XPOST -H "Authorization: token $GITHUB_SECRET" https://api.github.com/repos/steemit/steem/statuses/$(git rev-parse HEAD) -d "{
   \"state\": \"pending\",
   \"target_url\": \"${BUILD_URL}\",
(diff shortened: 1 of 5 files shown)


sneak:
updated github settings! @jredbeard

ciscripts/buildpending.sh
```diff
@@ -2,5 +2,6 @@
 curl -XPOST -H "Authorization: token $GITHUB_SECRET" https://api.github.com/repos/steemit/steem/statuses/$(git rev-parse HEAD) -d "{
```
sneak: thanks for figuring out all this arcane bullshit @jredbeard - super glad i don't have to. :D

jredbeard: @mvandeberg - It needs to be approved and rebased with all branches or you'll see that 'UI error' - jenkins was reporting it's status as 'default' and I updated the context it reports in JSON to github as part of this pull request.

PR #712, opened 2016-12-15
675 reduce chain mem

mvandeberg:
There was a bug that could cause corruption in the block log and prevent linking future blocks.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -224,9 +224,9 @@ void database::close(bool rewind)
 }
 
 bool database::is_known_block( const block_id_type& id )const
-{
+{ try {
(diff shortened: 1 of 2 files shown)


theoreticalbts:
In `update_last_irreversible_block()` I locate the following issues:

- We should have handling in case `next` is `nullptr`.  This should never happen, but in case it does, logging and throwing is better than crashing.
- We should maybe have `ilog()` when the `else` activates which dumps the block ID it ends up selecting.  It should be rare enough to be spammy, and it will help if there's some problem with this code.
- Instead of copying the `signed_block` into a local variable, use a pointer.  Block is probably the single largest object we have, and it's a performance win to avoid copying it (especially since the copy would occur for every block).

PR #713, opened 2016-12-15
563 bandwidth algorithm

mvandeberg:
This is a soft fork update of the old bandwidth algorithm that properly accounts for multiple transactions in a single block.


theoreticalbts:
`enum` members go into the enclosing namespace.  So if we're going to have these be in `steemit::chain`, let's use longer, more descriptive names than `post`, `forum`, `market`.

Alternatively, we could put the `enum` in some class such as the `account_bandwidth_object` class, or use the C++11 `enum class` construct (see `Scoped Enumerations` [here](http://en.cppreference.com/w/cpp/language/enum) ).  I think this is the better solution.

Issue #715, opened 2016-12-16
Add pkg-config to Dockerfiles, docs


theoreticalbts:
We depend on `pkg-config`, which is usually present on normal systems, but for some reason it is not installed when running in Docker on AWS dev environment.

jredbeard:
Approved #720

PR #720, opened 2016-12-19
Add pkg-config to Dockerfile, docs

theoreticalbts:
See #715


mvandeberg:
Deferring to @jredbeard as he has more experience with our Dockerfile on AWS.

PR #741, opened 2016-12-30
740 submodule refactor

bytemaster:
This will simply move where the protocol and chain libraries live so that they can be upgraded, versioned, and released independently of the steemd infrastructure that wraps them.  

@sneak - we will need to update the build / commit rules on these two submodules to match the process we use for the steem master repository.  These two submodules do not build on their own.


sneak:
Review or no review, it needs to pass tests to merge. :)

As for the branch protection on the other repos, I can set that up, but one of the main reasons for it is to gate on CI passing, which if these don't build/test on their own, I can't see much use in it.

Why pull this code out into separate repos anyway, if they don't have separate tests?

abitmore:
By the way, after this is done, we need to watch more repositories for at least PR's (I guess we won't set up independent issue tracking system for them).

sneak:
The submodule needs to use the public github http endpoint, not your private/authenticated git+ssh one:

```
 ---> Running in f9266df9e1e7
Submodule 'libraries/chain' (git@github.com:steemit/steem_chain.git) registered for path 'libraries/chain'
Submodule 'libraries/chainbase' (https://github.com/steemit/chainbase.git) registered for path 'libraries/chainbase'
Submodule 'libraries/fc' (https://github.com/steemit/fc.git) registered for path 'libraries/fc'
Submodule 'libraries/protocol' (git@github.com:steemit/steem_protocol.git) registered for path 'libraries/protocol'
[91mCloning into 'libraries/chain'...
```

sneak:
@jredbeard could you please add dan to the CI server's list of allowed users?  `dan@$DOMAIN`.

bytemaster:
```
Submodule 'libraries/protocol' (git@github.com:steemit/steem_protocol.git) registered for path 'libraries/protocol'
[91mCloning into 'libraries/chain'...
[0m[91mHost key verification failed.
[0m[91mfatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

It is a public repo, anyone should be able to checkout via ssh if the server has the proper keys for their github account.

sneak:
Not everyone has a github account or uses SSH to clone.  Our CI server, for example, clones via https - which is why this is failing.

.gitmodules
```diff
@@ -4,3 +4,9 @@
 [submodule "libraries/chainbase"]
 	path = libraries/chainbase
 	url = https://github.com/steemit/chainbase.git
+[submodule "libraries/protocol"]
+	path = libraries/protocol
+	url = git@github.com:steemit/steem_protocol.git
```
sneak: just change this part to use `https://` URLs like the other submodules.

jredbeard:
@sneak he's already in the list of users (dan@steemit.com), but it also looks like you pointed out the reason this build failed already. @bytemaster in the future to see console output from the CI server you just click 'details' or the red X next to a build failure in github, it will link you directly to that job on the CI server and then click 'console output' on the left. We'll be making the CI server build jobs a little more intuitive in the future.

jredbeard:
Or below where it says 'review required' and 'all checks have failed', right below that is 'jenkins-ci-steemit', the 'Details' link there brings you to the job as well.

You can also see the last success/failure and link to them by branch here: https://github.com/steemit/steem/branches/all

mvandeberg:
After discussing the change we are going to hold off on moving these directories to submodules. The entire directory structure needs to be redone when moving to the multi-chain fabric. There is no sense making this change now.

Issue #751, opened 2017-01-03
Incorrect total_payouts for `tags` in get_state


roadscape:
![image](https://cloud.githubusercontent.com/assets/5168676/21616219/48dcd3b0-d1a6-11e6-8d23-505a8a5a1f51.png)

```
>>> get_state "xxxxxxxxxxxx"
...
  "tags": {
...
    "steemit": {
      "name": "steemit",
      "total_children_rshares2": "200039810295173886873591931589",
      "total_payouts": "295800102.592 SBD",
      "net_votes": 920150,
      "top_posts": 22429,
      "comments": 83806
    },
```

mvandeberg:
Upon investigation of where this data is calculated in the tags plugin, correcting the issue is not a simple fix given the structure of the tags plugin. Is this data used anywhere besides this view and how important is it that we maintain this view?

mvandeberg:
Disregard the previous comment. I think I have a way to do it.

mvandeberg:
Closing as this was merged.

PR #752, opened 2017-01-05
Incorrect total_payouts for `tags` in get_state #751

mvandeberg:
Fix for #751

`total_payout` in `tag_stats_object` was incorrectly being summed repeatedly on every update to a comment. Now it only adds the payout from the `comment_reward_operation` vop.


theoreticalbts:
Looks good.  `filter_tags()` is done twice, but getting rid of that is an optimization we can do later.

mvandeberg:
This entire plugin will be replaced as an external service. And payouts do not occur as often as other updates, so the inefficiency is a low percent of the time spent in the plugin.

PR #755, opened 2017-01-05
Bump version to 0.16.1

mvandeberg:


theoreticalbts:
Let's bump fc as well

PR #757, opened 2017-01-06
.gitignore file updated to exclude IntelliJ IDEA metadata.

Nemo1369:


mvandeberg:
58a9c7f is contained in PR #758

PR #758, opened 2017-01-06
Tests linking fix

Nemo1369:
Fixed linking error with boost::test dynamic libraries usage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,95 @@
+## Directory-based project format:
+.idea/
+# if you remove the above rule, at least ignore the follwing:
+
+# User-specific stuff:
+.idea/workspace.xml
+.idea/task_types.xml
+.idea/dictionaries
+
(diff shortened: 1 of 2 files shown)


theoreticalbts:
As far as the `.gitignore` changes go, I don't think we should add any IDE-generated files to it (except possibly `*.so` and `*.dylib`).  Only build script / compiler intermediates.

As explained [here](https://help.github.com/articles/ignoring-files/) it seems like a good alternative is using `core.excludesfile` to set up a global `.gitignore` for your user account, and put ignore rules for your favorite IDE in there.  You can also put these in a per-repo ignore using `.git/info/excludes`.

This pull request is very difficult to review because a bunch of irrelevant stuff is lowercased, making for a very cluttered diff.

Nemo1369:
Well, I think, then I would need to apply changes to unlinted CMakeLists.txt and open a new pull request when it's done.

Issue #766, opened 2017-01-09
Remove Bandwidth Rate Limiting from Consensus


mvandeberg:
Bandwidth only needs to be used to filter transactions by witnesses and can safely be implemented as a soft fork outside of consensus. There will still be a supported method by which to calculate bandwidth.

This issue usurps #733

theoreticalbts:
Contains partial revert of #176

mvandeberg:
The branch contains code for #732 as well. I knocked them both out at the same time and needed the hardfork constants. If you would rather, we can merge #732, rebase #766 on master and then merge #766.

mvandeberg:
Merged

Issue #732, opened 2016-12-26
Remove Posting Limit


bytemaster:
Currently there is a 4-post per day soft limit on posts.  This limit should be removed as part of an overall simplification plan.

mvandeberg:
This issue usurps #733

bitcoinparadise:
Would this not allow lower quality content? It could turn into a numbers game and people could try to pump out posts as much as possible to see what percentage of their posts they created. 

This can make it confusing and even harder to find REAL content when I find it hard as it as already.

3beanzz:
Is it a good thing or a bad thing if somebody powers up a lot more than [this user](https://steemit.com/@tamim) did in order to post more than 4 times and upvote himself?  How would we address the issue of spam if this were a user who powered up a lot more SP?

mvandeberg:
This has been merged.

PR #775, opened 2017-01-10
766 remove bandwidth rate limiing

mvandeberg:
This contains updates for both #766 and #732


mvandeberg:
Postpone merging this.

mvandeberg:
Rebased and force pushed. This can be merged after review.

TimCliff:
Is there a change to handle the rate limiting outside of consensus?

mvandeberg:
It will be handled the same way it currently is except that it will be in an `is_producing()` block. Meaning we only check it on block production and on fully validating nodes. In a future release we will migrating all of that into the witness plugin itself and out of consensus code entirely.

Issue #776, opened 2017-01-11
Implement chainbase lock checking


theoreticalbts:
- Import code from https://github.com/steemit/chainbase/issues/8
- Enable at compile-time in `CMakeLists.txt`
- Enable at run-time in `main.cpp` or `application.cpp`

theoreticalbts:
Merged and seems to be working, so I'm closing this.

PR #777, opened 2017-01-11
Implement chainbase lock checking #776

theoreticalbts:


mvandeberg:
Can you make a PR to for the chainbase branch so I can more easily review that change?

PR #780, opened 2017-01-12
Steem v0.16.2

mvandeberg:
Fixes a bug in some APIs for read only nodes that could cause the node to crash. This does not effect write nodes and has no impact on block production or the stability of the network.


mvandeberg:
Postpone this. We can add the lock checking to this so that our production environment can check for errors.

mvandeberg:
Cherry picked the lock fixes.

Issue #782, opened 2017-01-12
Refactor reward computations out of database


theoreticalbts:
Let's decouple the comment reward computation helper methods from the database, communicating via a struct exactly what information is being passed.  This is a refactoring to simplify implementation of: equalizing rs2 values within a block #771, #773, #768, and makes it simpler to prove that potential future parallelizations are correct.

theoreticalbts:
Added to HF17 because this blocks #771.

theoreticalbts:
Merged, closing.

Issue #771, opened 2017-01-10
Equalize Per Block rshare Value


mvandeberg:
In each block where a payout occurs, each comment should receive the same STEEM/rshares such that the comments should be able to be paid out in any order and arrive at the same state.

theoreticalbts:
Basically the fraction `total_reward_fund_steem / total_reward_shares2` needs to be pre-computed.  Since `total_reward_shares2` is much greater than `total_reward_fund_steem` it suffices to divide by a pre-computed fraction `total_rshares2 / rf`.  This fraction needs to be rounded *up* so the reward of each author will effectively be rounded down.

theoreticalbts:
A unit test to ensure these are commutative will entail using one of several clamps [1] to reduce one payout to the point where a significant amount would be awarded to the next cashout, then ensure the payouts remain the same under the new HF rules.

[1] Clamps include `reward_weight`, `max_steem`, `STEEMIT_MIN_PAYOUT_SBD`.

theoreticalbts:
Closing since this is merged.

mvandeberg:
This is not merged. #790 was merged.

theoreticalbts:
> This is not merged.

Saw a merged PR in the ticket and no PR's referencing this feature, so I assumed it had been merged.  Turns out someone didn't create a PR because someone wanted to hold off merging until someone had a unit test for this.  And I guess "someone" is me.  Sorry about that.

It's ready to be merged now though, Rebased and added a unit test.

mvandeberg:
Merged

PR #790, opened 2017-01-13
Refactor rewards computations to be stateless

theoreticalbts:
See #782


theoreticalbts:
This is a dependency of #771.  Please review.

theoreticalbts:
Added to HF17 because this blocks #771.

PR #793, opened 2017-01-15
Added seed node

clayop:
If wrong branch was chosen, please let me know.


mvandeberg:
This change is include in #795

Issue #803, opened 2017-01-17
Refactor adjust_rshares2 to be non-recursive


theoreticalbts:
`adjust_rshares2` does two separate things:

- (A) Update the `children_rshares2` method of the current comment and all ancestors
- (B) Update the `total_reward_shares2` of `dynamic_global_properties`

Address this with the following strategy:

- Refactor (A) into its own function called `update_children_rshares2()`
- Refactor `update_children_rshares2()` to use a `while` loop instead of recursion
- Have `adjust_rshares2()` call `update_children_rshares2()`

Related:  #801 #774

theoreticalbts:
Added to HF17 because this blocks #774.

mvandeberg:
Merged

Issue #774, opened 2017-01-10
Comment Reward Pool


mvandeberg:
Comments (as opposed to posts) should have their own reward pool. The implementation should be abstract to allow for further divisions if we choose to do so in the future.

Team-Steem:
I agree with pfunk that if this was to be implemented then comment would benefit from having their own voting power. 

Also I feel like we should begin with 2-4% of the reward pool to comments. On reddit the comments are super nice and they have a big 0 reward.

mvandeberg:
We are interested in feedback on whether comments and post should each have their own voting power.

On the one hand it allows users to upvote each without having to worry about how their votes effect the other category. However, voting power is already enough of an enigma to users that creating another type of voting power could create confusion.

bitcoinparadise:
I'm in the middle of this Idea. One hand it's another reward pool to increase rewards and increase engagement in the comment section.

On the other hand, Most likely we will see complaints about how both reward pools are being "drained" from the comment reward pool. Most likely have bots to vote on comments and not read the content itself.

Team-Steem:
I'm beginning to doubt there should be a comment reward pool. I'm currently trying to read as much Steem post as I can and I can't even keep up already. 

The comment reward pool would split our attention.

Also wouldn't there be a risk of whales abusing this too easily? 

It was a mess when there wasn't a 30 minutes penalty on posts. Would there be a penalty on comments? If not what would be the consequence. 

At this point, I use comments like this. I upvote the comments on my posts to make them in order I want to see them. I see mostly the same done on other posts and maybe it's the way it should stay.

abitmore:
This feature is unneeded imo. If you want to give more rewards to comments, just have some different UI for example "trending comments", "hot comments", "new comments", "comment feeds" then people will see them more and upvote them more. As @arhag said:
> It is a little bit ironic though, that the motivation is to not have the interface drive the blockchain, but that's exactly what separating comment reward pools does.

However perhaps this shouldn't be discussed here. See more discussion: https://steemit.com/steem/@steemitblog/steem-0-17-change-proposal-introduction

Mustard5:
As a rare post creator and frequent commentator, I would like to see a comment rewards pool.  The question I have about the detail is does that mean that we will have separate voting power levels for comments and articles?

[...]

PR #804, opened 2017-01-17
Refactor adjust_rshares2 to be non-recursive

theoreticalbts:
See #803.  Added to HF17 because this blocks #774.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -1723,30 +1723,40 @@ void database::clear_null_account_balance()
       adjust_supply( -total_sbd );
 }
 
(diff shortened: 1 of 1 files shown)


libraries/chain/database.cpp
```diff
@@ -1723,30 +1723,40 @@ void database::clear_null_account_balance()
...
+      } );
+
+      if( current_comment->depth == 0 )
+         break;
+
+      current_comment = &db.get_comment( current_comment->parent_author, current_comment->parent_permlink );
+   }
+}
+
 /**
```
mvandeberg: This comment is not accurate. It wasn't before this change, but it certainly is not accurate now.

Updating rshares2 for a single comment is needed for payout calculations.
Updating children_rshares2 is non-consensus and could be skipped for validation only nodes.

theoreticalbts:
Fixed the comment, should be ready to merge

PR #815, opened 2017-01-18
Equalize rewards within a block

theoreticalbts:
See #771.


tests/tests/operation_time_tests.cpp
```diff
@@ -23,6 +23,131 @@ using namespace steemit::protocol;
...
+      authors.emplace_back( "dave" , dave_private_key );
+      voters.emplace_back( "ulysses", ulysses_private_key, "alice");
+      voters.emplace_back( "vivian" , vivian_private_key , "bob"  );
+      voters.emplace_back( "wendy"  , wendy_private_key  , "dave" );
+
+      // A,B,D : posters
+      // U,V,W : voters
+
+      // set a ridiculously high STEEM price ($1 / satoshi) to disable dust threshold
+      set_price_feed( price( asset::from_string( "0.001 TESTS" ), asset::from_string( "1.000 TBD" ) ) );
```
mvandeberg: How does the algorithm handle equalizing payouts when there are dust payouts?

tests/tests/operation_time_tests.cpp
```diff
@@ -23,6 +23,131 @@ using namespace steemit::protocol;
...
+         ilog( "${n} : ${steem} ${sbd}", ("n", voter.name)("steem", a.balance)("sbd", a.sbd_balance) );
+      }
+
+      const account_object& alice_account = db.get_account("alice");
+      const account_object& bob_account   = db.get_account("bob");
+      const account_object& dave_account  = db.get_account("dave");
+
+      BOOST_CHECK( alice_account.sbd_balance == ASSET( "10720.000 TBD" ) );
+      BOOST_CHECK( bob_account.sbd_balance == ASSET( "0.000 TBD" ) );
+      BOOST_CHECK( dave_account.sbd_balance == alice_account.sbd_balance );
```
mvandeberg: What if the comments have different rshares? The test should prove that they are paid out equal Steem per rshares.

theoreticalbts:
OK, I need to explain a little about the architecture of these changes.  And this will combine explaining this PR, and #782 which is already merged.

In 0.16 and earlier versions, the computation of how much reward a post gets is a member of `database` and thus, in principle it can reference arbitrary information out of database.  However in practice, it only actually uses a few specific pieces of information from the database.  #782 makes the communication between the code that computes rewards and the database more "at arms length," by putting the needed information from the database into `struct comment_reward_context` [here](https://github.com/steemit/steem/blob/2ca460f03fc9555259dc4050d2891655608958c6/libraries/chain/include/steemit/chain/util/reward.hpp#L21-L29) and then passing the `struct` to a free-floating function [here](https://github.com/steemit/steem/blob/2ca460f03fc9555259dc4050d2891655608958c6/libraries/chain/util/reward.cpp#L7-L34) that computes the reward.

Some of the information from the database is specific to the particular comment (e.g. how many rshares are on the comment), call this *local information*.  Other information from the database exists outside of the comments, basically the current price feed and `dynamic_global_property_object` fields, call this *global information*.  These sets of information are filled into the `struct` from the database by two different methods [here](https://github.com/steemit/steem/blob/2ca460f03fc9555259dc4050d2891655608958c6/libraries/chain/database.cpp#L1955-L1968).

The #782 code re-updates the global information in the `struct` before processing each comment, because the global information can change as a result of a comment being processed, and the old code used the changed version.  #782 does not change any behavior, assuming there were no bugs in the PR the code after #782 should do exactly what it did before #782, it's just been re-structured to use a more modular architecture.

Now we get to this PR and the behavior-changing #771.  The architectural change in #782 makes implementing this trivial -- you just fill the global information into the `struct` once, before you start looping through the payouts, instead of filling it in on each comment.

PR #817, opened 2017-01-19
API Updates

bytemaster:
Allow improved filtering of feeds by tag and author.
Track extra stats about authors, followers, and tags
Allow arbitrary truncation of post content when fetching indexes

### DIFF
--- a/libraries/app/database_api.cpp
+++ b/libraries/app/database_api.cpp
@@ -1244,7 +1244,7 @@ map< uint32_t, applied_operation > database_api::get_account_history( string acc
 vector<pair<string,uint32_t> > database_api::get_tags_used_by_author( const string& author )const {
    return my->_db.with_read_lock( [&]()
    {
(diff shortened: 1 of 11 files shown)


theoreticalbts:
In implementation of `select_authors` parameter [here](https://github.com/steemit/steem/pull/817/commits/926b9f9e8e5311fdb136f7ef7ed49608ff24ccfa) `blog_itr.end()` may be forced to iterate over all of `b_idx`.  How big is this index?  Is this a potential DoS?

roadscape:
`get_state` is not performing any truncation, yet it would be the most useful place to do so.

Experiments show that most posts render previews just as before if truncation happens at 512 chars. 
256 is not enough. 1024 would be even safer.

roadscape:
Truncation in `get_state` should only be applied to lists of top-level posts, never when loading a full discussion thread (post+comments).

PR #821, opened 2017-01-23
Beanstalk

jredbeard:


theoreticalbts:
This is very much specific to Steemit's deployment process, so I think this repo isn't the best place for this code.  IMHO this should be modularized by either (a) making a `Dockerfile` that inherits from `steemit/steem`, or (b) making a new repo that includes this repo as a submodule.

contrib/fullnode.config.ini
```diff
@@ -0,0 +1,111 @@
...
+# dbg-init-key = 
+
+# API user specification, may be specified multiple times
+# api-user = 
+
+# Set an API to be publicly available, may be specified multiple times
+public-api = database_api login_api network_broadcast_api follow_api market_history_api raw_block_api block_info_api
+
+# Plugin(s) to enable, may be specified multiple times
+enable-plugin = witness account_history tags follow market_history
```
sneak: @jredbeard should the witness plugin be enabled on a normal full node?

contrib/startpaassteemd.sh
```diff
@@ -0,0 +1,117 @@
...
+    done
+    # start haproxy now that the config file is complete with all endpoints
+    # all of the read-only processes will connect to the write node onport 8091
+    # haproxy will balance all incoming traffic on port 8090
+    /usr/sbin/haproxy -D -f /etc/haproxy/haproxy.steem.cfg 2>&1
+    # start runsv script that kills containers if they die
+    mkdir -p /etc/service/steemd
+    cp /usr/local/bin/paas-sv-run.sh /etc/service/steemd/run
+    chmod +x /etc/service/steemd/run
+    runsv /etc/service/steemd
```
sneak: i think this line might be redundant - runit will pick up everything in /etc/service/* that has an executable `run` inside of it.

sneak:
@theoreticalbts as discussed, this is relevant to anyone who deploys steemd via docker, not just steemit.  there's nothing specific to our deployment in here.

PR #836, opened 2017-01-31
Stable

tzipcloud:


mvandeberg:
The commits that exist in stable but not in master will be cherry picked in. We will not do a downstream merge.

PR #846, opened 2017-02-03
770 disable pow

theoreticalbts:
Disable PoW.  Still needs testing.


libraries/chain/include/steemit/chain/witness_objects.hpp
```diff
@@ -159,6 +159,11 @@ namespace steemit { namespace chain {
          uint32_t                                                          witness_pay_normalization_factor = 25;
          chain_properties                                                  median_props;
          version                                                           majority_version;
+
+         uint8_t max_voted_witnesses            = STEEMIT_MAX_VOTED_WITNESSES_HF0;
```
mvandeberg: These properties are set, but not updated in the hardfork logic

mvandeberg: Nevermind, database.cpp wasn't showing on my diff initially.

PR #847, opened 2017-02-04
Ci multi build exec

jredbeard:
This branch adds support for the Pipeline Multibranch style project in Jenkins. The plugin will scan the repo for branches that include a Jenkinsfile in the root and they will be included in the folder view within Jenkins.

This adds quite a few desirable features for both now and when we switch Jenkins to build off of ec2 spot instances.

1. You can now pause and resume builds (they will pick up where they left off).
2. You can kick off a build for an individual branch in Jenkins without a push to github.
3. The groovy script via the Jenkinsfile can be scripted to do parallel operations (like multiple tests at the same time, etc.)
4. Potential for further abstraction of stages of the build process (build, test, deploy) and visual feedback of each stage in Jenkins which will be beneficial in achieving continuous delivery. For now it's one stage that executes essentially the same scripts we've been using but it can be dramatically expanded upon.
5. Workspaces are separated for each branch which will allow multiple build executors.

I've left the old 'Steem' project intact so CI builds will complete as they have been without changes to every branch (to not impact the flow of current CI). Once this is merged in I'd like to remove the old build project in jenkins and rename the new 'Steem multibranch project' to just 'steem'. At that point we can turn on multiple build executors. If you'd like to test on a branch before merging, you can add the Jenkinsfile to the root of any branch along with the scripts from the ciscripts directory. The plugin will automatically pick it up as part of the multibranch pipeline project. Keep in mind that until the old project is removed in Jenkins the webhook will kick off a build from both project types once you add a Jenkinsfile (but will fail in the old project type of course).

@mvandeberg and @theoreticalbts let me know when you're ready to switch and I'll remove the old project from Jenkins and turn up the build executors. @sneak tagging you so you see this change too.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 #!groovy
 node {
   try {
-    properties([pipelineTriggers([[$class: 'GitHubPushTrigger'], pollSCM('H/2 * * * *')])])
+    properties([pipelineTriggers([[$class: 'GitHubPushTrigger']])])
     checkout scm
     sh 'ciscripts/triggerbuild.sh'
     sh 'ciscripts/buildsuccess.sh'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,9 +1,11 @@
 #!groovy
-node {
+pipeline {
+  agent any
+
   stages {
     stage('Build') {
       steps {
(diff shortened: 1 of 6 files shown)


mvandeberg:
Let's wait until after 0.17.0 is release to remove to old scripts since we many issue branches that are not yet merged.

Issue #773, opened 2017-01-10
Comment Reward Beneficiaries


mvandeberg:
As a comment option, authors should be able to specify a number of beneficiaries to receive a percentage of the rewards and the percent of the potential liquid rewards to receive as Steem Power. Unallocated rewards by default go to the author as they currently do.

21xhipster:
Should work for comments also. Especially if #774 will be implemented. Comments to the same post can be published from different apps. Developers should be rewarded for all efforts.

mvandeberg:
Yes. All content will be eligible for this.

mvandeberg:
The `comment_options_operation` extension field can contain a vector of pairs of account names and weights to specify the comment beneficiaries. 

```
struct comment_payount_beneficiaries
{
   vector< std::pair< account_name_type, uint16_t > > beneficiaries;

   void validate();
};
```

Beneficiaries can be set when there are no votes on a comment. Setting options of any kind should be done in the same transaction that the comment is created in.

mvandeberg:
Testing is waiting on #659 to be merged.

mvandeberg:
Merged

PR #866, opened 2017-02-13
Comment Reward Beneficiaries

mvandeberg:
PR for #773


libraries/chain/database.cpp
```diff
@@ -1595,7 +1596,20 @@ void database::cashout_comment_helper( util::comment_reward_context& ctx, const
             share_type curation_tokens = ( ( reward_tokens * get_curation_rewards_percent() ) / STEEMIT_100_PERCENT ).to_uint64();
             share_type author_tokens = reward_tokens.to_uint64() - curation_tokens;
 
-            author_tokens += pay_curators( comment, curation_tokens );
+            idump( (author_tokens) );
```
bytemaster: extra spam that should be removed

libraries/chain/database.cpp
```diff
@@ -1595,7 +1596,20 @@ void database::cashout_comment_helper( util::comment_reward_context& ctx, const
...
+            curation_tokens = pay_curators( comment, curation_tokens );
+            author_tokens += curation_tokens;
+            idump( (author_tokens) );
```
bytemaster: console spam

libraries/chain/include/steemit/chain/comment_object.hpp
```diff
@@ -168,6 +169,8 @@ namespace steemit { namespace chain {
          bool              allow_replies = true;      /// allows a post to disable replies.
          bool              allow_votes   = true;      /// allows a post to receive votes;
          bool              allow_curation_rewards = true;
+
+         bip::vector< std::pair< account_name_type, uint16_t >, allocator< std::pair< account_name_type, uint16_t > > > beneficiaries;
```
bytemaster: use shared_vector<beneficiary_type>  and define beneficiary type as { account_name, percent_type }

libraries/chain/steem_evaluator.cpp
```diff
@@ -316,6 +316,35 @@ void delete_comment_evaluator::do_apply( const delete_comment_operation& o )
...
+struct comment_options_extension_visitor
+{
+   comment_options_extension_visitor( const comment_object& c, database& db ) : _c( c ), _db( db ) {}
+
+   typedef void result_type;
+
+   const comment_object& _c;
+   database& _db;
+
+   void operator()( const comment_payout_beneficiaries& cpb ) const
```
bytemaster: require that beneficiares are all unique and sorted by weight then account name, ideally this check would be done as part of static validation.

libraries/protocol/steem_operations.cpp
```diff
@@ -76,13 +76,42 @@ namespace steemit { namespace protocol {
...
+      }
+   };
+
+   void comment_payout_beneficiaries::validate()const
+   {
+      uint32_t sum = 0;
+
+      FC_ASSERT( beneficiaries.size() < 128, "Cannot specify more than 127 beneficiaries." ); // Require size serializtion fits in one byte.
+
+      for( auto& route : beneficiaries )
```
bytemaster: ensure they are sorted by weight / beneficiary and that each beneficiary is only included once

libraries/protocol/steem_operations.cpp
```diff
@@ -76,13 +76,52 @@ namespace steemit { namespace protocol {
...
+   void comment_payout_beneficiaries::validate()const
+   {
+      uint32_t sum = 0;
+
+      FC_ASSERT( beneficiaries.size(), "Must specify at least one beneficiary" );
+      FC_ASSERT( beneficiaries.size() < 128, "Cannot specify more than 127 beneficiaries." ); // Require size serializtion fits in one byte.
+
+      string_less str_cmp;
+
+      for( size_t i = 0; i < beneficiaries.size() - 1; i++ )
```
bytemaster: This loop looks dangerous because it depends upon line 95, consider changing it to be 

``` 
for( size_t i = 1; i < beneficiars.size(); ++i )
```

And then update internals to use i-1 and i rather than i and i+1

PR #875, opened 2017-02-17
Remove ntp option from config file

mvandeberg:
This was one change from `stable` that never made it upstream into `master`


mvandeberg:
After closer review, this is not needed.

PR #884, opened 2017-02-23
adding beanstalk specifics to master for ease of merging

jredbeard:
We do intend to possibly split this out into it's own repo, but for now, since we're actually running this code in production, we need to get it in the repo so that new versions are easy to deploy (per @sneak) - either way, it doesn't effect anything else and all lives in the contrib directory and the dockerfile. Anyone can still run a node with this code with the same options they've been using as specified in the readme file, nothing changes there. We may also make the S3 buckets public read-only that contain blockchainstate (shared memory) files, if we do that then anyone could use this code to start a synced node about as quickly as they could download a 10gb file from S3. One thing to note that I would like to possibly make a standard if we can - the addition of a version listed at the top of the dockerfile (ex. ENV VERSION 0.0.0) - this will make it easy to keep generated blockchainstate files separated for different versions (which means the version would probably need to change at least as frequently as the shared memory files need to be updated). Otherwise let's come up with something else to keep track of those. Set reviewer to sneak but tagging @mvandeberg and @theoreticalbts because I want you guys to see this and share your opinion on keeping an updated version in the dockerfile as well.


jredbeard:
@sneak let's try to get this merged in before 0.17rc1 is released?

contrib/startpaassteemd.sh
```diff
@@ -34,17 +34,28 @@ chown steemd:steemd $HOME/config.ini
...
+  mount -t ramfs -o size=34816m ramfs /mnt/ramdisk
+  s3cmd get s3://$S3_DOWNLOAD_BUCKET/blockchain-$VERSION-latest.tar.bz2 - | pbzip2 -m2000dc | tar x --wildcards 'blockchain/block*' -C /mnt/ramdisk 'blockchain/shared*'
+  ln -s blockchain/block_log /mnt/ramdisk/blockchain/block_log
+  ln -s blockchain/block_log.index /mnt/ramdisk/blockchain/block_log.index
+  ARGS+=" --shared-file-dir=/mnt/ramdisk/blockchain"
+  chown -R steemd:steemd /mnt/ramdisk/blockchain
 else
-  echo error: no URL specified to get blockchain state from - exiting
+  s3cmd get s3://$S3_DOWNLOAD_BUCKET/blockchain-$VERSION-latest.tar.bz2 - | pbzip2 -m2000dc | tar x
+  touch /tmp/issyncnode
```
sneak: i would have used /var/run myself

PR #886, opened 2017-02-27
774 comment reward fund

mvandeberg:
This implements the new reward curve on both funds along with curation rewards.

### DIFF
--- a/libraries/app/database_api.cpp
+++ b/libraries/app/database_api.cpp
@@ -1112,7 +1112,7 @@ void database_api::set_pending_payout( discussion& d )const
    u256 total_r2 = to256( props.total_reward_shares2 );
 
    if( props.total_reward_shares2 > 0 ){
-      auto vshares = steemit::chain::util::calculate_vshares( d.net_rshares.value > 0 ? d.net_rshares.value : 0  );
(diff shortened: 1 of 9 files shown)


theoreticalbts:
Here are the most obvious issues:

- Magic numbers (`40` and `277777777777ll`) should be in a `#define` in `config.hpp`, not inline in the code.  My notation has an advantage here in that the number `40` already has a name.
- This code should have comments with overflow analysis showing that arithmetic does not overflow the data type.  In particular code [here](https://github.com/steemit/steem/blob/087b83820fe31aeaaeed515518b74e16213ec8c0/libraries/chain/util/reward.cpp#L97-L99) looks like it may need `u256` for numerator.
- `(r+a)^2 - a^2` makes it clear that this is expression is a shift of `r^2`, but having the expression be `r^2 + 2*r*a = r*(r+2a)` would let us use one less multiplication [here](https://github.com/steemit/steem/blob/087b83820fe31aeaaeed515518b74e16213ec8c0/libraries/chain/util/reward.cpp#L88) and [here](https://github.com/steemit/steem/blob/087b83820fe31aeaaeed515518b74e16213ec8c0/libraries/chain/util/reward.cpp#L99) (multiplication by 2 can be optimized to addition by the compiler)
- Instead of using `std::numeric_limits<uint64_t>()` as scaling constant, let's use shifting by 64 bits (perhaps by initializing the words of `uint128_t` manually)

PR #927, opened 2017-03-10
Explicit UTF8 json_metadata checks implemented

### LINKED ISSUE #926: UTF8 json_metadata check absence leads to Condenser vulnerability
A try to write a non-UTF8 symbols sequence to json_metadata field in comment_operation leads to successful write in blockchain, failure of websocket output on comment retrieval try.

Nemo1369:
Closes #926

### DIFF
--- a/libraries/protocol/steem_operations.cpp
+++ b/libraries/protocol/steem_operations.cpp
@@ -93,6 +93,7 @@ namespace steemit { namespace protocol {
 
       if( json_metadata.size() > 0 )
       {
+         FC_ASSERT( fc::is_utf8(json_metadata), "JSON Metadata not formatted in UTF8" );
          FC_ASSERT( fc::json::is_valid(json_metadata), "JSON Metadata not valid JSON" );
       }
    }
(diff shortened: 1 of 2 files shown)


mvandeberg:
This PR will not correctly fix the problem. We cannot add another requirement to validate or else it would invalidate bad operations that are already in the blockchain. The added pruning on returning the comment is not needed. The reason we need to do it in the body is because sometimes we substring the body and need to make sure we split on a UTF-8 character boundary. This is not needed for title or category because those fields are never split.

libraries/protocol/steem_operations.cpp
```diff
@@ -93,6 +93,7 @@ namespace steemit { namespace protocol {
 
       if( json_metadata.size() > 0 )
       {
+         FC_ASSERT( fc::is_utf8(json_metadata), "JSON Metadata not formatted in UTF8" );
```
mvandeberg: Has to go into an evaluator and only take effect after a hardfork.

mvandeberg:
#930 usurps this PR

Issue #925, opened 2017-03-09
0.17.0rc2: wrong check when SP delegation is decreasing


abitmore:
The check [here](https://github.com/steemit/steem/blob/v0.17.0rc2/libraries/chain/steem_evaluator.cpp#L2295-L2299) (`delegate_vesting_shares_evaluator::do_apply(...)` of *steem_evaluator.cpp*) is wrong:
```
   else /* delegation->vesting_shares > op.vesting_shares */
   {
      FC_ASSERT( delegation->vesting_shares - op.vesting_shares >= min_delegation || op.vesting_shares.amount == 0, "Delegation must be removed or leave minimum delegation amount of ${v}", ("v", min_delegation) );

      auto delta = delegation->vesting_shares - op.vesting_shares;
```

It should be:

```
op.vesting_shares.amount == 0
||
( op.vesting_shares.amount >= min_delegation && delta >= min_update )
```

Note that `min_update` is needed here to fight abusing.

mvandeberg:
Merged

PR #928, opened 2017-03-10
0.17.0rc2: wrong check when SP delegation is decreasing

mvandeberg:
PR for #925

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -2284,10 +2284,10 @@ void delegate_vesting_shares_evaluator::do_apply( const delegate_vesting_shares_
    // Else if the delegation is increasing
    else if( op.vesting_shares >= delegation->vesting_shares )
    {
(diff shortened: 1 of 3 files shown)


libraries/chain/steem_evaluator.cpp
```diff
@@ -2307,10 +2307,11 @@ void delegate_vesting_shares_evaluator::do_apply( const delegate_vesting_shares_
    // Else the delegation is decreasing
    else /* delegation->vesting_shares > op.vesting_shares */
    {
-      FC_ASSERT( delegation->vesting_shares - op.vesting_shares >= min_delegation || op.vesting_shares.amount == 0, "Delegation must be removed or leave minimum delegation amount of ${v}", ("v", min_delegation) );
-
       auto delta = delegation->vesting_shares - op.vesting_shares;
 
+      FC_ASSERT( delta >= min_update, "Steem Power increase is not enough of a different. min_update: ${min}", ("min", min_update) );
```
abitmore: There are edge cases that `op.vesting_shares == 0` is true but `delta` be less than `min_update`, for example when account registration fee is increased. So in #925 I suggested `op.vesting_shares == 0` OR other checks.

PR #929, opened 2017-03-10
Update impacted

mvandeberg:
This is a review over the impacted visitor. There were a number of ops that were missing impacted accounts and several visitors that only listed the authority, which is covered in a template default case.


theoreticalbts:
This change will re-number historical operations within each account's history and insert operations in the past that weren't there before.  Some clients may consider this a breaking change (especially those running automated scripts that scan account histories).

I think this is a change that needs to be discussed further.  It is a non-consensus change, so I propose postponing this until at least 0.17.1.

PR #936, opened 2017-03-13
Update License

bytemaster:


neoxian:
This needs to be done.  It's very important

cryptohazard:
Following @neoxian, I think this is very important.

PR #945, opened 2017-03-16
Update LICENSE.md to reflect views of Steemit, Inc. expressed by @sneak

TimCliff:
There has been a lot of discussion on Steemit.com over the terms in the Steem blockchain license. This pull request clarifies the language of the license to express the spirit of the license that was expressed by @sneak in this post.

https://steemit.com/steem/@demotruk/petition-to-demand-steem-be-made-open-source#@sneak/re-smooth-re-sneak-re-timcliff-re-sneak-re-neoxian-re-demotruk-petition-to-demand-steem-be-made-open-source-20170316t010456554z

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -13,7 +13,7 @@ modification, are permitted provided that the following conditions are met:
 4. The STEEMIT_INIT_PUBLIC_KEY_STR is not changed from STM8GC13uCZbP44HzMLV6zPZGwVQ8Nt4Kji8PapsPiNq1BK153XTX,
 and the software is not modified in any way that would bypass the need for the corresponding private to start
 a new blockchain.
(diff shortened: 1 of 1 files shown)


sneak:
The problem is "the witnesses" is defined by consensus, and the code is what defines consensus, so you've introduced a circular dependency which cannot be unambiguously resolved—nor could you or anyone reasonably expect most non-technical people implementing the law to fully grok the analysis.

Please make these discussions on steemit.com, ideally with the help of lawyers.  This is absolutely in the spirit of the license, but this specific version isn't clear enough.  Hammer it out on the site and make opening a PR as the last step after the dust is settled, please. :)

Closing this not as a denial of the concept, just as an indication that this specific wording isn't up to snuff yet.  We're all on the same page here, I think.

PR #962, opened 2017-03-22
Steem v0.17.1

mvandeberg:


roadscape:
Will #965 be included here?

mvandeberg:
Yes, when @theoreticalbts approves that PR

theoreticalbts:
If you guys want #965 in here, I'll hold off approving this until we can add that patch to the PR.  Maybe #968 should be in here as well.

Issue #967, opened 2017-03-22
Improve per block and fork logging


mvandeberg:

mvandeberg:
Merged

PR #966, opened 2017-03-22
show witness and latency on 'Got block' messages

roadscape:
This can greatly help with debugging witness/fork issues.

adapted from https://github.com/abitmore/steem/tree/v0.16.0-a-bit-more-logs

sample output:

```
777275ms th_a       application.cpp:499           handle_block         ] Got 8 transactions on block 10405693 by jesta -- latency: 275 ms
780250ms th_a       application.cpp:499           handle_block         ] Got 7 transactions on block 10405694 by roelandp -- latency: 250 ms
783214ms th_a       application.cpp:499           handle_block         ] Got 4 transactions on block 10405695 by good-karma -- latency: 214 ms
786271ms th_a       application.cpp:499           handle_block         ] Got 9 transactions on block 10405696 by pfunk -- latency: 271 ms
789392ms th_a       application.cpp:499           handle_block         ] Got 9 transactions on block 10405697 by blocktrades -- latency: 392 ms
792227ms th_a       application.cpp:499           handle_block         ] Got 5 transactions on block 10405698 by witness.svk -- latency: 227 ms
795292ms th_a       application.cpp:499           handle_block         ] Got 5 transactions on block 10405699 by charlieshrem -- latency: 292 ms
798148ms th_a       application.cpp:499           handle_block         ] Got 10 transactions on block 10405700 by roadscape -- latency: 148 ms
801571ms th_a       application.cpp:499           handle_block         ] Got 5 transactions on block 10405701 by abit -- latency: 571 ms
804455ms th_a       application.cpp:499           handle_block         ] Got 8 transactions on block 10405702 by pharesim -- latency: 455 ms
807247ms th_a       application.cpp:499           handle_block         ] Got 13 transactions on block 10405703 by arhag -- latency: 247 ms
810391ms th_a       application.cpp:499           handle_block         ] Got 8 transactions on block 10405704 by klye -- latency: 391 ms
813225ms th_a       application.cpp:499           handle_block         ] Got 7 transactions on block 10405705 by gtg -- latency: 225 ms
816168ms th_a       application.cpp:499           handle_block         ] Got 7 transactions on block 10405706 by xeldal -- latency: 168 ms
819288ms th_a       application.cpp:499           handle_block         ] Got 7 transactions on block 10405707 by yme-08 -- latency: 288 ms
822231ms th_a       application.cpp:499           handle_block         ] Got 8 transactions on block 10405708 by bhuz -- latency: 231 ms
[...]

PR #968, opened 2017-03-22
Improve per block and fork logging

mvandeberg:
PR for #967 
Usurps #966


theoreticalbts:
Please change `auto` to `const signed_block&` -- blocks are big, copying them should be avoided.

Issue #991, opened 2017-03-29
Price constructor should be explicit


mvandeberg:
A root cause of the bug in #987 was caused by am implicit conversions from `asset` to `price`. We should define the `price` constructor as explicit to fail complication on other.

mvandeberg:
After changing the constructor to explicit, this bug shows up in the cli wallet as well.

mvandeberg:
We want to also remove the default args from the price constructor and prevent prices from being created with a quote of 0. However, those changes are more invasive and have unintended side-effects. We will do more investigating and testing before implementing those changes.

theoreticalbts:
Needs updated PR and possibly hardfork check.

vogel76:
I propose to add static method:
static price nullPrice();
and use it in inital assignment in all places where directly default constructor is used.

PR #995, opened 2017-03-29
Price constructor should be explicit

mvandeberg:
PR for #991


theoreticalbts:
This will need a hardfork check, and should be held off until the next hardfork.

Issue #832, opened 2017-01-26
Witness Block Reward Visibility


bacchist:
Witness rewards for block signing are hard to account for. Making these rewards visible will help witnesses and prospective witnesses by providing them with more complete and accurate information to guide their decisions to invest in the platform.

Issue #521, opened 2016-10-19
ability to blacklist ops for account_history plugin


roadscape:
Use case: for most accounts on steemit.com, vote operations make up the majority of entries in account histories, even though we do not have any reason to access these ops.

This is an issue because the `get_state @account/transfers` API only takes the last 1000 items and filters out everything but ops relating to transfers/orders/rewards. 

This API call would be much more useful if the last 1000 ops didn't contain votes. Users would have a much longer history of transfers/rewards (some report a history as short as 5 hours)

Reference: https://github.com/steemit/steemit.com/issues/375

roadscape:
There are 2 main issues which this workaround would fix:
 - limited author/curation rewards history https://github.com/steemit/steemit.com/issues/375
 - limited conversion request history https://github.com/steemit/steemit.com/issues/896

An alternative/easier solution may be to simply iterate over the last 10,000 items when loading `get_state "@user/transfers"` instead of just the last 1,000. 

Looking at some active accounts, it seems 1,000 can be barely enough to cover 24hrs, which means it's simply not enough data to be useful for these reports.

mvandeberg:
For now I am going to file this as won't fix. We are planning on moving non-consensus data to an external db, account history included, which will allow us to more easily query account history. I don't think we should spend developer time on updating a plugin that will become deprecated in a few months. 

I am leaving the ticket open until we have a solution, but will no be working on one right now.

roadscape:
What about:

> An alternative/easier solution may be to simply iterate over the last 10,000 items when loading get_state "@user/transfers" instead of just the last 1,000.

mvandeberg:
#301 addresses this

Issue #301, opened 2016-08-18
Filtering of account history


theoreticalbts:
We will need several account histories -- in particular we need to filter out "spammy" transactions like fill orders, voting, curation rewards, and (when #300 is implemented) block production.  These should still be accessible in a separate history, but not included in the main history.  (The website already does this, but internal organization in steemd only has a single account history type -- it should be partitioned instead.)

It might also be useful to summarize these operations somehow -- e.g. you may not care too much about each individual block produced or each individual vote rewarded, but it might be useful to get a period summary each day / week.  The summary would make it possible to audit the user's balance in the block where the summary is included (along the lines of #300).

mvandeberg:
I would like us to have different categories of operations and then be able to white list or black list the categories.

PR #1015, opened 2017-04-03
#832 Add virtual operation for producer reward.

NateBrune:


roadscape:
This will get very spammy for witness accounts, would be great to have op filtering config -- https://github.com/steemit/steem/issues/521 (@mvandeberg had better idea how to implement this, not sure if there is a separate issue)

libraries/protocol/include/steemit/protocol/steem_virtual_operations.hpp
```diff
@@ -166,6 +166,16 @@ namespace steemit { namespace protocol {
       asset             reward;
    };
 
+   struct producer_reward_operation : public virtual_operation
+   {
+      producer_reward_operation(){}
+      producer_reward_operation( const string& o, const asset& v ) : owner( o ), vesting_shares( v ) {}
+
+      account_name_type owner;
```
mvandeberg: Change the variable name to `witness` or `producer`. In the context of the witness object `owner` makes sense because the account name is the owner of the witness object. In the virtual op, it makes less sense.

mvandeberg:
#301 allows for op white and black listing so this is safe to merge now.

Issue #1030, opened 2017-04-12
boost 1.58 issues & doc needs to be updated


roadscape:
Building latest with Boost 1.58 did not work on Ubuntu. @mvandeberg suggested the minimum may need to be 1.59 instead.

Gandalf-the-Grey:
https://github.com/steemit/steem/pull/1034

pfunks:
Boost 1.58 is in the Ubuntu 16.04 APT repo. Since boost takes forever to build, that's been pretty convenient.

mvandeberg:
The error on 14.04 is gcc, not with Boost. The requirements are Boost 1.58 and gcc 4.9. 16.04 is compatible on both fronts out of box.

PR #1031, opened 2017-04-12
update min boost version in doc/building.md

roadscape:
#1030

### DIFF
--- a/doc/building.md
+++ b/doc/building.md
@@ -108,28 +108,21 @@ Here are the required packages:
         libreadline-dev \
         perl
 
-Steem requires Boost 1.57 or later. The Boost provided in the Ubuntu 14.04
+Steem requires Boost 1.59 or later. The Boost provided in the Ubuntu 14.04
 package manager (Boost 1.55) is too old. So building Steem on Ubuntu 14.04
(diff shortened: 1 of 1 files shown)


Gandalf-the-Grey:
That's not a small change because Ubuntu 16.04 LTS has packaged boost 1.58
On Debian Stable which has boost 1.55 packaged, people need to build some boost ( 1.58 <= x <= 1.60) anyway
But for Ubuntu 16.04 LTS (AFAICT most common distro/version among witnesses) it is:
to build boost or not to build

mvandeberg:
Our docker images are Ubuntu 16.04 LTS afaik. Maybe I am mistaken and they are using Boost 1.58 (I thought they were using 1.60). Which then begs the question, why did roadscape's build fail with 1.58?

roadscape:
My build on Ubuntu 14.04 failed with 1.58

Gandalf-the-Grey:
so probably because of his gcc version, not boost as such

mvandeberg:
This appears to be caused by a gcc version, not a boost version.

roadscape:
The doc still needs to be updated if boost 1.57 is no longer supported

Issue #853, opened 2017-02-06
unexpected/invalid trx hashes reported for vops in get_account_history


roadscape:
```
>>> get_account_history blood -1 10
#     BLOCK #    TRX ID          OPERATION            DETAILS
-------------------------------------------------------------------------------
69    8092893    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"blood","withdrawn":"2356.674631 VESTS","deposited":"0.000 STEEM"}
 70    8293553    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"pfunk","withdrawn":"2356.674631 VESTS","deposited":"2356.674631 VESTS"}
 71    8293553    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"blood","withdrawn":"2356.674631 VESTS","deposited":"0.000 STEEM"}
 72    8494853    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"pfunk","withdrawn":"2356.674631 VESTS","deposited":"2356.674631 VESTS"}
 73    8494853    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"blood","withdrawn":"2356.674631 VESTS","deposited":"0.000 STEEM"}
 74    8696244    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"pfunk","withdrawn":"2356.674631 VESTS","deposited":"2356.674631 VESTS"}
 75    8696244    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"blood","withdrawn":"2356.674631 VESTS","deposited":"0.000 STEEM"}
 76    8897557    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"pfunk","withdrawn":"2356.674631 VESTS","deposited":"2356.674631 VESTS"}
 77    8897557    0000000000000000000000000000000000000000 fill_vesting_withdraw {"from_account":"blood","to_account":"blood","withdrawn":"2356.674631 VESTS","deposited":"0.000 STEEM"}
[...]

PR #1036, opened 2017-04-14
#853-fix-invalid-vop-trx

NateBrune:


theoreticalbts:
What causes the incorrect virtual op addressed by this patch?  We should find an example and turn it into a unit test, which fails without the patch and passes with the patch.  So we know for sure that this patch addresses the issue.

I'm pretty sure this patch isn't correct.  Imagine a transaction T with ops A, B, C, D, where C is a v-op caused by processing B.  I am thinking the txid of op C should be T.id(), this patch would cause it to be zeros.  Further, I think the txid for op D would also be zeroed out by this patch!

Based on a quick `grep` of `_current_trx_id`, it is being set at the beginning of `_apply_transaction` and set to `transaction_id_type()` at the end of `_apply_transaction`.  It's clear the intended semantics are that any v-op processed during a transaction should inherit the transaction's txid, and any v-op processed outside a transaction should have a zeroed txid.

How could incorrect txid's appear in some v-ops with this code?  The code zeroing `_current_trx_id` might not run if control's diverted by an exception.  So `_current_trx_id` needs to be zeroed unconditionally -- even if exceptions occur.  Many languages have a `finally` keyword for "unwinding" actions that need to execute whenever control flow goes outside a block by any path, what we want is this:

```
void database::_apply_transaction(const signed_transaction& trx)
{
   try
   {
      _current_trx_id = trx.id();
      ...   // exception could be thrown by this code
   }
   finally
   {
      // this block runs when the try block is exited by any path
      _current_trx_id = transaction_id_type();
   }
}
```

However, C++ does not have `finally`.  You have to create an object type which implements the "unwinding" action in its destructor, then create an instance of this object on the local stack.  The destructor runs whenever control flows out of the function by any path.  Maybe something like this:

```
struct trx_id_resetter
{
   transaction_id_type& txid;
   trx_id_resetter( transaction_id_type& t ) : txid(t) {}
   ~trx_id_resetter() { txid = transaction_id_type(); }
};

void database::_apply_transaction(const signed_transaction& trx)
{
   trx_id_resetter r( _current_trx_id );
   ...   // this code could throw an exception
   // trx_id_resetter destructor will be called if we exit on any path
}
```

I'm pretty familiar with this problem as I wrote code to deal with a similar issue for `skip_flags`.

mvandeberg:
I agree with @theoreticalbts. If the vop is generated during a transaction (filling an order, for example), then the vop should have the tx id of the associated transaction. Other than that, it should be 0. Sorry for changing the spec on you.

Issue #1035, opened 2017-04-13
Add locks in application.cpp


mvandeberg:
The net code accesses chain state through application.cpp and is not properly locked.

PR #1038, opened 2017-04-14
Add locks in application.cpp

mvandeberg:
PR for #1035


theoreticalbts:
More locking looks necessary.

mvandeberg:
I still don't think it is necessary because these methods are only called on a write node, but for now it is better safe than sorry. (I draw this conclusion because required locks here implies BitShares needs locking as well)

Issue #894, opened 2017-03-02
Move Soft Forks into Witness Plugin


mvandeberg:
All soft forking logic and data structures needed to handle soft forks should belong in the witness plugin. They are not part of consensus and this plugin is the most logical destination for that logic.

PR #1041, opened 2017-04-14
Move Soft Forks into Witness Plugin

mvandeberg:
PR for #894

### DIFF
--- a/libraries/app/CMakeLists.txt
+++ b/libraries/app/CMakeLists.txt
@@ -9,7 +9,7 @@ add_library( steemit_app
              ${HEADERS}
            )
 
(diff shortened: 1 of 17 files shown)


theoreticalbts:
About the `is_reindexing` flag introduced by this patch:  Its design is philosophically questionable -- it affects the current processing of the blockchain based on information from the future.  A separate discussion is needed about `is_reindexing`.

However, the `is_reindexing` flag is not actually used by this patch.  So let's remove it.

Also, merge conflicts.

theoreticalbts:
Sorry, I wasn't specific enough, should remove the related `_reindex_head_num` field as well.  `git commit --amend` will suffice

mvandeberg:
This was my bad. We had talked, decided not to implement this particular subfeature, and I didn't clean up my code.

Issue #1023, opened 2017-04-07
Unnecessary assignment of _db in evaluators


mvandeberg:
In [evaluator.hpp:36-39](https://github.com/steemit/steem/blob/master/libraries/chain/include/steemit/chain/evaluator.hpp#L36-#L39) the evaluator has `_db` defined as protected and `db()` returns `_db`.

All evaluators have a line `database& _db = db();` which needlessly redefines `_db` in the local scope. This line can be removed.

If keeping `_db` safe is important then `evaluator::_db` should be made private and `evaluator::db()` inlined.

theoreticalbts:
I don't like having the `db()` accessor, we should just have it access the member variable.

PR #1058, opened 2017-04-19
Unnecessary assignment of _db in evaluators

mvandeberg:
PR for #1023

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -63,7 +63,6 @@ struct strcmp_equal
 
 void witness_update_evaluator::do_apply( const witness_update_operation& o )
 {
-   database& _db = db();
    _db.get_account( o.owner ); // verify owner exists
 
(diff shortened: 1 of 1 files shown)


theoreticalbts:
I think at some point very early in Graphene history, `_db` was `private` and only accessible by `db()` getter.  Of course then you have `db()` all over the place (and that may still the case in some evaluators), so it's more convenient to pull it out with one call and use it everywhere else.

I personally don't like getters / setters or `private` / `protected`.  Basically their purpose is to use the compiler to enforce a design principle (avoiding tight coupling / breaking modularity).  The thing about design principles though is that they're qualified -- sometimes it makes sense to access to something that's private, in which case the compiler-enforced rules simply get in your way.

I like the Python approach -- everything is public, but things that implementation details start with an underscore.  If you use something with an underscore, you can, but you should understand that you're creating a tight coupling to something that the other side considers a candidate for future breaking changes.

Issue #1018, opened 2017-04-05
Hardcode claim amount and comment/post list sizes for hf17


mvandeberg:
HF17 loaded the recent_claims field by calculating the value from hf16 to hf17. The value is now a part of consensus and can be hard coded so it does not need to recalculated on reindex. Likewise, to minimize the impact of the hardfork, list sizes for payout time transformation can be specified exactly to reduce allocation impact.

PR #1059, opened 2017-04-19
Hardcode claim amount and comment/post list sizes for hf17

mvandeberg:
PR for #1018


theoreticalbts:
This approval is conditional on being able to replay with this code.

Issue #187, opened 2016-07-20
steemd: get_block does not return transaction_ids


roadscape:
`get_block` in cli_wallet returns transaction_ids array, but calling it on the node itself seems to not -- ideally it would return this data!

xallavon:
+1 here.
IMHO, `transaction_id` should come inside transaction object, not as distinct list. And it should be provided by the node api as well as wallet.

theoreticalbts:
In theory, the txid is redundant information, since it can be computed from the tx content by serializing and hashing the transaction.

In practice, transaction serialization is very much non-trivial, and a moving target, if you're not writing C++ code with FC and Steem headers.

sneak:
It would be nice if get_block also returned the block hash, too.  It returns the previous, but not the one being got.

sneak:
@theoreticalbts could you take a peek at this and see if it's a ton of effort to add?  I imagine it's a very quick change.

abitmore:
The CLI command `get_block` does return trx ids. But the database API doesn't.

sneak:
The database API is what I'm using.  I hope to avoid using cli_wallet until the end of my days.

roadscape:
> It would be nice if get_block also returned the block hash, too. It returns the previous, but not the one being got.

👍

mvandeberg:
We can add the derived data with an api object wrapper class.

Need to be sure that the tx IDs and block hashes are not derived again in the CLI wallet once they are being returned in the node API. The API objects should be made the match what the current output of the CLI wallet is.

PR #1062, opened 2017-04-20
steemd: get_block does not return transaction_ids

mvandeberg:
PR for #187


theoreticalbts:
`signed_block_api_obj` constructor parameter should be passed by reference, not value.

Other than that, looks good.

Issue #931, opened 2017-03-10
impacted.cpp Not Complete


mvandeberg:
Not all impacted accounts are actually returned. Ideally, this would not have been the case in past releases. Some services rely on the numbering we give to operations in history and adding operations retroactively mutates the numbering.

Opening this up for discussion. Should impacted.cpp be correct, or should existing operations be immutable to maintain backwards compatibility with existing services?

theoreticalbts:
Related: #929

sneak:
I think changing the order/returning more data should be fine. Clients shouldn't be depending on the order anyway.

xeroc:
I can remember that when there was a bug in BitShares that made invalid proposals appear in the operations history, we had turned some exchanges mad when we change the operations id.
I would request that you contact exchanges and businesses and ensure that they at least know about the change if you decided to do so.

mvandeberg:
#1019

mvandeberg:
We have added virtual ops retroactively in the past. That changes IDs as well. Not sure if we contacted exchanges, but I never heard complaints. We should contact them nonetheless.

PR #1065, opened 2017-04-20
impacted.cpp Not Complete

mvandeberg:
PR for #931


theoreticalbts:
We discussed this in #931 so it looks reasonable.

Issue #1032, opened 2017-04-12
Remove Category Object


mvandeberg:
The category object is non consensus and is an outdated precursor to the tags plugin. The object should be removed from consensus and anything important that it was doing should be migrated to the tags plugin.

PR #1088, opened 2017-05-04
#1032 remove category object

NateBrune:


libraries/app/database_api.cpp
```diff
@@ -1170,7 +1170,28 @@ void database_api::set_pending_payout( discussion& d )const
...
+   {
+      try
+      {
+         tags  = fc::json::from_string( root.json_metadata )["tags"].as< std::vector< std::string > >();
+         if(tags.at(0) != ""){
+           d.url = "/" + tags.at(0) + "/@" + root.author + "/" + root.permlink;
+         }
+         else {
+           d.url = "/" + tags.at(1) + "/@" + root.author + "/" + root.permlink;
+         }
```
mvandeberg: what if there are no tags in the meta data?

libraries/chain/include/steemit/chain/comment_object.hpp
```diff
@@ -37,79 +37,20 @@ namespace steemit { namespace chain {
...
          comment_object( Constructor&& c, allocator< Allocator > a )
-            :category( a ), parent_permlink( a ), permlink( a ), title( a ), body( a ), json_metadata( a ), beneficiaries( a )
+            :parent_permlink( a ), permlink( a ), title( a ), body( a ), json_metadata( a ), beneficiaries( a )
          {
             c( *this );
          }
 
          id_type           id;
 
-         shared_string     category;
```
mvandeberg: I think for backwards compatibility of comments that existed before we were use metadata tags, this field should still exist.

Issue #1102, opened 2017-05-11
Restore Tag Length to 32 characters


mvandeberg:
In 0.18.3 the tag length was shortened to 16 characters and needs to be expanded to 32 characters to support the length that Condenser limits to.

PR #1104, opened 2017-05-11
Restore Tag Length to 32 characters

mvandeberg:
PR for #1102


theoreticalbts:
`endian_reverse()` should be in `namespace fc` not `namespace std`.  Putting it in `std` is actually [undefined behavior](http://en.cppreference.com/w/cpp/language/extending_std).

However, according to [the docs](http://www.boost.org/doc/libs/1_63_0/libs/endian/doc/conversion.html), `endian_reverse<T>()` must be "a non-member function in the same namespace as T that can be found by argument dependent lookup (ADL)."

This means we cannot retrofit `std::pair` to be EndianReversible / EndianReversibleInplace.  We must make our own endian-reversible pair type, `fc::erpair` type in namespace `fc`.

I added some commits to do this.

Issue #1103, opened 2017-05-11
Check Chainbase Indices on Debug Only


mvandeberg:
Checking if an index exists on every access is slowing down steemd response times for API calls. We should only force check for index existence in debug and not in release.

theoreticalbts:
This is not a change we should make.  I really don't want to remove any defensive programming checks from steemd without a good reason.  Substantial performance improvement *may* be a good reason.

My intuition is that the performance gain from this patch will be less than 0.1%, and may be so small that even measuring it is not practical.  I would find it very surprising if this patch adds a substantial performance improvement.

I will reconsider letting this patch into `master` if we can get some numbers to show that my intuition is wrong, and measure a performance improvement of, say, 5% or more.

- If we have benchmarking to show this change gives a substantial improvement, then the question becomes a tradeoff of less safety checks versus more performance.  Maybe this is a worthwhile tradeoff, maybe not.
- If we can't show this change gives a substantial improvement, then it's a tradeoff of less safety checks for no gain.  Not a worthwhile tradeoff.

PR #1105, opened 2017-05-11
Check Chainbase Indices on Debug Only

mvandeberg:
PR for #1103


theoreticalbts:
Not approved.  See discussion in #1103.

Issue #971, opened 2017-03-23
0.17.0: cancelling delegated SP sometimes will fail.


abitmore:
[Code](https://github.com/steemit/steem/blob/v0.17.0/libraries/chain/steem_evaluator.cpp#L2312-L2313):
```
      FC_ASSERT( delta >= min_update, "Steem Power increase is not enough of a different. min_update: ${min}", ("min", min_update) );
      FC_ASSERT( op.vesting_shares >= min_delegation || op.vesting_shares.amount == 0, "Delegation must be removed or leave minimum delegation amount of ${v}", ("v", min_delegation) );
```

There are edge cases that `op.vesting_shares == 0` is true but `delta` be less than `min_update`, for example when account registration fee is increased. So in #925 I suggested `op.vesting_shares == 0` OR other checks.

Also mentioned [here](https://github.com/steemit/steem/pull/928#pullrequestreview-27201353).

theoreticalbts:
Good catch.  This is more of an inconvenience than a serious issue.

Postponing a fix until the next major release, because the fix will change consensus.  Fix should be:

```
if (has_hf_18)
{
   FC_ASSERT(  (op.vesting_shares.amount == 0) || ((delta >= min_update) && (op.vesting_shares >= min_delegation) );
}
else { /* old checks */ }
```

We can't softfork the new check in with `is_producing()`, because it relaxes the old check (we can only softfork in checks that tighten, not relax, the acceptable conditions).

PR #1110, opened 2017-05-16
0.17.0: cancelling delegated SP sometimes will fail.

mvandeberg:
PR for #971


theoreticalbts:
Actually, can we add a TODO comment here to remove the old version once the HF date passes?

Issue #997, opened 2017-03-29
Using account_create_with_delegation with 0 delegation creates unnecessary objects


arhag:
There needs to be a check for `o.delegation > 0` around [these lines](https://github.com/steemit/steem/blob/v0.18.0/libraries/chain/steem_evaluator.cpp#L268-L274).

As the code exists right now, it is inefficiently allowing unnecessary `vesting_delegation_object`s to be created. And furthermore because the `vesting_shares` of that unnecessary object would be 0, [this branch](https://github.com/steemit/steem/blob/v0.18.0/libraries/chain/steem_evaluator.cpp#L2290) of the code is activated rather than the else branch, making it difficult for the user to directly remove that object (they can still remove it indirectly by first adding more steem power to the delegation and then removing it).

If you leave this fix for a future hard fork, you will need to make sure to iterate through the `vesting_delegation_index` when the hardfork is applied and erase any objects with a `vesting_shares` field equal to 0.

It is tempting to think that this can be done as a change in v0.18.1 (with a reindex) without causing hardforks. But I believe a hardfork could theoretically be possible if someone adds a small amount (larger than `min_update` but smaller than `min_delegation`) to such a 0 `vesting_shares` object. In fact, the timing differences of `min_delegation_time` alone could cause a fork issue.

mvandeberg:
I tend to agree with you on the fix for this. Better to do it as a hardfork and be safe then try it as a softfork and fork the blockchain.

PR #1111, opened 2017-05-16
Using account_create_with_delegation with 0 delegation creates unnecessary objects

mvandeberg:
PR for #997

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -3881,6 +3881,23 @@ void database::apply_hardfork( uint32_t hardfork )
                   });
                }
             }
+
+            vector< const vesting_delegation_object* > to_remove;
+            const auto& delegation_idx = get_index< vesting_delegation_index, by_id >();
+            auto delegation_itr = delegation_idx.begin();
+
+            while( delegation_itr != delegation_idx.end() )
+            {
+               if( delegation_itr->vesting_shares.amount == 0 )
+                  to_remove.push_back( &(*delegation_itr) );
+
+               ++delegation_itr;
+            }
+
+            for( const vesting_delegation_object* delegation_ptr: to_remove )
+            {
+               remove( *delegation_ptr );
+            }
          }
          break;
       default:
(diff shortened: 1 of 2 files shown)


libraries/chain/database.cpp
```diff
@@ -3809,6 +3809,19 @@ void database::apply_hardfork( uint32_t hardfork )
       case STEEMIT_HARDFORK_0_18:
          break;
       case STEEMIT_HARDFORK_0_19:
+         {
+            const auto& delegation_idx = get_index< vesting_delegation_index, by_id >();
+            auto itr = delegation_idx.begin();
+
+            while( itr != delegation_idx.end() )
```
NateBrune: Might be worth optimizing this using Boost's `erase(iterator first,iterator last)` if we order by delegated vesting shares, or just `erase(const key_type& x)` if we keep vesting shares a member in the index.

mvandeberg: We do not track that index. This is the safe way to implement the fix. The 0 delegation objects should have no effect on consensus. After the hardfork we can change the 0.17 behavior to match 0.19 and verify that not creating 0 delegation objects has no effect. In which case we can delete this logic entirely.

theoreticalbts:
You have to be really careful about concurrently modifying collections.  Let's use a `lower_bound()` on the ID to get a completely fresh iterator after that `remove()`.

Let's add a TODO comment noting how this can be simplified after the HF passes.

Whether zero delegations are represented by a zero delegation object, or by the lack of a delegation object, is a non-consensus implementation detail.  So I'd personally avoid implementing this as a hardfork.  But this implementation is probably a little safer, in case there is some consensus difference I overlooked, it won't split the network.

mvandeberg:
We use the same pattern [here](https://github.com/steemit/steem/blob/master/libraries/chain/database.cpp#L1571-L1588).

theoreticalbts:
The evaluator check is:

```
if( hf19 && amount > 0 || !hf19 )
```

You can simplify any expression of the form `(a & b) | !a`, it distributes to `(a|!a) & (b|!a)`, the first term is tautological so it simplifies further to `b|!a`.  We cannot always apply this simplification with short circuit logical operators (e.g. when evaluation of `b` is undefined or will cause side effects when `a` is false), but we can do so here (`o.delegation_amount` is always defined).  So a clearer test is

```
if( (amount > 0) || !hf19 )
```

Also, `for( auto itr : remove )` should be `for( const delegation_object* object_ptr : remove )` for two reasons:

- Code is clearer if you avoid `auto`
- These are pointers to objects, not iterators being dereferenced

Issue #987, opened 2017-03-29
account_create_operation after HF17


abitmore:
The [code](https://github.com/steemit/steem/blob/v0.18.0/libraries/chain/steem_evaluator.cpp#L120-L126) seems not right.

```
   if( _db.has_hardfork( STEEMIT_HARDFORK_0_17__818 ) )
   {
      const witness_schedule_object& wso = _db.get_witness_schedule_object();
      FC_ASSERT( o.fee >= wso.median_props.account_creation_fee * asset( STEEMIT_CREATE_ACCOUNT_WITH_STEEM_MODIFIER, STEEM_SYMBOL ), "Insufficient Fee: ${f} required, ${p} provided.",
                 ("f", wso.median_props.account_creation_fee * asset( STEEMIT_CREATE_ACCOUNT_WITH_STEEM_MODIFIER, STEEM_SYMBOL ) )
                 ("p", o.fee) );
   }
```

Both `wso.median_props.account_creation_fee` and `asset( STEEMIT_CREATE_ACCOUNT_WITH_STEEM_MODIFIER, STEEM_SYMBOL )` are instances of `struct asset`, what will happen with that `*` operator? The latter will be converted to a `price` instance implicitly?

mvandeberg:
We are going to disable to op through a soft fork. `account_create_with_delegation` works identically if the delegation is set to 0.

mvandeberg:
The permanent fix will be in a later hardfork.

arhag:
@mvandeberg:

If `account_create_with_delegation_operation` can do everything `account_create_operation` can do, then do you plan to just disable `account_create_operation` in hard fork 19?

Also, [this code](https://github.com/steemit/steem/blob/v0.18.0/libraries/chain/steem_evaluator.cpp#L268-L274) unnecessarily creates an object if o.delegation is 0.

mvandeberg:
I want to deprecate `account_create_operation`, but want to give enough time for other services to update.

Yeah, it looks like it would but I want to be sure that removing it doesn't cause forking.

xeroc:
> I want to deprecate account_create_operation, but want to give enough time for other services to update.

Could you guys please consider setting up a page where you summarize things like that? Not everyone has the time to read all the issues on github. Already plenty of features broken on my side with 0.18 :(

PR #1112, opened 2017-05-16
account_create_operation after HF17

mvandeberg:
PR for #987


theoreticalbts:
If we want to deprecate `account_create_operation`, why are we re-enabling it?  Given that it's softfork-disabled on the live network, everyone who cares has surely updated their stuff to not use `account_create_operation` by now.

mvandeberg:
We did not intend to deprecate `account_create_operation`. We just unintentionally broke it. Because it is only a softfork and allows free account creation, we either need to deprecate it completely, or fix it.

Issue #977, opened 2017-03-24
Delete Comment Condition Should be Consensus


mvandeberg:
The restriction on only deleting posts without positive net_rshares should be a full consensus requirement.

https://github.com/steemit/steem/blob/master/libraries/chain/steem_evaluator.cpp#L373-#L375

PR #1113, opened 2017-05-16
Delete Comment Condition Should be Consensus

mvandeberg:
PR for #977


theoreticalbts:
Should really add a `TODO` to remove `is_producing()` from the conditional once the HF time has passed.

Issue #1052, opened 2017-04-18
Squareroot Curation Rewards Curve


mvandeberg:
The curation rewards curve should be changed to a square root function.

We need to decide how to handle curation rewards of posts prior to the hardfork. Because the actual hardfork time is flexible, I propose having two weight fields in the `comment_vote_object` and every vote on a comment for the week prior to the hardfork tracks both weights. Then, when the hardfork occurs, we can determine when the first comment paid out on the new curve was created and make a clean break at that particular time to optimize the implementation later.

This issue is a WIP while we are discussing the HF19 proposal [here](https://steemit.com/steemit/@steemitblog/proposing-steem-equality-0-19-0-as-the-next-fork).

Issue #1051, opened 2017-04-18
Linear Rewards Curve


mvandeberg:
Change the rewards curve to be `n` from `n^2`. We will need a "dummy" rewards fund object to track linear claims prior to the hardfork. This dummy reward fund track back to HF16.

This issue is a WIP while we are discussing the HF19 proposal [here](https://steemit.com/steemit/@steemitblog/proposing-steem-equality-0-19-0-as-the-next-fork).

PR #1120, opened 2017-05-18
Linear Rewards Curve

mvandeberg:
PR for #1051 and #1052


theoreticalbts:
In `process_comment_cashout`, looks like both branches of the `if` do the same calculation except for the denominator.  Let's just say something like:

```
fc::seconds decay_rate;
if( rfo.name == TEMP_LINEAR || has_hardfork(HF19) )
    decay_rate = DECAY_HF19;
else
    decay_rate = DECAY_HF17;
```

theoreticalbts:
The magic number `1` in `funds[ 1 ]` should be a `#define` constant to tell us which fund that is.

Issue #1126, opened 2017-05-22
Forbid no-op zero delegation


theoreticalbts:
We should forbid setting a zero delegation when the delegation is already zero.

Related: #971

theoreticalbts:
On further review, this isn't actually a problem.  In the old code we enforce `delta >= min_update`.  In the new code, zero delegation objects are deleted, so setting delegation to zero goes through the `delegation == nullptr` path and is denied.

theoreticalbts:
Some of the messages have readability issues.  Also, let's add the check anyway -- it shouldn't hurt.

PR #1130, opened 2017-05-22
Forbid nop delegation #1126

theoreticalbts:
I suggest reviewing this as two commits.  The first commit is intended to be a refactoring and cleanup with no functional changes (aside from minor cosmetic changes to error messages).

The second commit is for the issue addressed in #1126.


mvandeberg:
The last else should only be needed temporarily (and should never execute currently).

I think it would be useful to add the condition on delegation creation that the delegation is greater than 0 on the off-chance that the witnesses set the account creation fee to 0.

PR #1136, opened 2017-05-25
Comment typo

testzcrypto:


mvandeberg:
The comment is correct, https://github.com/steemit/steem/blob/master/libraries/app/database_api.cpp#L1248

Issue #963, opened 2017-03-22
Fork monitoring plugin


theoreticalbts:
This plugin would save each incoming block in `block_id.bin` / `block_id.json` file, then delete the file when the block becomes irreversible.  If the plugin's running while the network's forking, the blocks that end up on minority forks will never be deleted.  Allowing us to preserve them to analyze why the network's forking.

sneak:
That's a neat idea.

mvandeberg:
For a first step, let's idump the blocks so that we can catch them in our logs, rather than saving them and trying to get them off our nodes.

PR #1140, opened 2017-05-28
#963 Fork monitoring plugin

NateBrune:
This plugin will analyze blocks as they come in to detect forks and log them.

### DIFF
--- a/libraries/plugins/fork_monitor/CMakeLists.txt
+++ b/libraries/plugins/fork_monitor/CMakeLists.txt
@@ -1,0 +1,17 @@
+file(GLOB HEADERS "include/steem/plugins/fork_monitor/*.hpp")
+
+add_library( fork_monitor_plugin
+             fork_monitor_plugin.cpp
+             ${HEADERS} )
+             
+target_link_libraries( fork_monitor_plugin chain_plugin )
+target_include_directories( fork_monitor_plugin
+                            PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" )
+
+install( TARGETS
+   fork_monitor_plugin
+
+   RUNTIME DESTINATION bin
+   LIBRARY DESTINATION lib
+   ARCHIVE DESTINATION lib
+)
--- a/libraries/plugins/fork_monitor/fork_monitor_plugin.cpp
+++ b/libraries/plugins/fork_monitor/fork_monitor_plugin.cpp
@@ -1,0 +1,80 @@
+#include <steem/plugins/fork_monitor/fork_monitor_plugin.hpp>
+
+#include <steem/protocol/types.hpp>
+
+#include <steem/chain/database.hpp>
+
+#include <fc/network/http/websocket.hpp>
+
+#include <fc/smart_ref_impl.hpp>
+
+namespace steem { namespace plugins { namespace fork_monitor {
+
+namespace detail {
+   using namespace steem::protocol;
+
+   class fork_monitor_plugin_impl
+   {
+      public:
+         fork_monitor_plugin_impl() {};
+         ~fork_monitor_plugin_impl() {};
+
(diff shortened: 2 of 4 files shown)


libraries/plugins/fork_info/fork_info_plugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
mvandeberg: Remove this notice

libraries/plugins/fork_info/include/steemit/fork_info/fork_info_plugin.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
mvandeberg: Remove this notice

libraries/plugins/fork_info/fork_info_plugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+         fork_info_plugin_impl( fork_info_plugin& plugin )
+            :_self( plugin ) {}
+         virtual ~fork_info_plugin_impl() {}
+
+         void on_block( const signed_block& b );
+
+         fork_info_plugin&       _self;
+         std::deque<signed_block> _last_blocks;
+   };
+   void fork_info_plugin_impl::on_block( const signed_block& b )
```
mvandeberg: This gets called every block that is pushed. So when we switch forks, all the blocks on the other fork will be pushed. They will have already been pushed at some point. It would be possible to have the vector have the blocks in it in the following order.

{1a, 1b, 1a, 2a, 1b, 2b, 1a, 2a, 3a, 1b, 2b, 3b, ... }

The 21 block limit could be exceeded before the block is potentially made reversible and repeat blocks may be logged.

Instead, I would use a map from block num to a vector of blocks with that number. Then, when a the last irreversible block number is increased, check blocks with num less than or equal to LIB until there is no vector of blocks. For each vector check if each block is in the current chain. If it is not, then log it.

mvandeberg:
Replaced with #1759

PR #1142, opened 2017-05-30
Update README.md

mvandeberg:
Update README.md for using CLI wallet in Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ Future invocations of your wallet can be simply be made with:
 
     docker exec -ti steemd-wallet /usr/local/steemd-default/bin/cli-wallet --server-rpc-endpoint=wss://steemd.steemit.com
 
+**WARNING:** If you want to keep your wallet file, you must copy it out of the container before you delete the container.
+
 # Environment variables
 
 There are quite a few environment variables that can be set to run steemd in different ways:
--- a/README.md
+++ b/README.md
@@ -55,6 +55,23 @@ that uses ca. 14GB of memory and growing:
 
     docker logs -f steemd-full
 
+## CLI Wallet
+
+To set up a cli wallet:
+
+    docker run \
+        -d --name steemd-wallet steemit/steem
+        
+    docker exec -ti steemd-wallet /bin/bash
+    
+    sv stop /etc/service/steemd
+    
+    /usr/local/steemd-default/bin/cli_wallet --server-rpc-endpoint=wss://steemd.steemit.com
+    
+Future invocations of your wallet can be simply be made with:
+
+    docker exec -ti steemd-wallet /usr/local/steemd-default/bin/cli-wallet --server-rpc-endpoint=wss://steemd.steemit.com
+
 # Environment variables
 
 There are quite a few environment variables that can be set to run steemd in different ways:


theoreticalbts:
These instructions won't work, they will store the file inside the Docker container which means THE USER WALLET WILL BE WIPED WHENEVER THE CONTAINER IS RESET.  If you're using `cli_wallet` with Docker, you need to mount an outside volume that will contain the user wallet.  Then we should instruct, perhaps repeatedly and PERHAPS IN ALL CAPS, that the user needs to keep multiple backups of their wallet file.

Also, the Docker container for the Steem wallet should not run steemd by default.  We should have separate containers for steemd and for the wallet (perhaps they inherit a common base container but the steemd container has startup scripts?)

Advice from Docker experts about this would be welcome @jredbeard @sneak

jredbeard:
@theoreticalbts 

I'm sure those instructions do work, but yes, your wallet.json would be lost and not stored in an outside volume.

Personally, I don't want to ever keep a wallet.json around - for me it's one time use. Import keys, perform desired actions, remove. I'm assuming this is probably the case for most cli_wallet users. I have my doubts that anyone wants to keep a wallet.json around stored locally with their actual keys in it - I could be wrong, I suppose.

With all that being said, I think we should just mention that the stored wallet.json will be lost when the container exits - and modify the instructions to be a single command instead of launching a daemonized container and then stopping steemd. Example something like: `docker run -i -t steemit/steem /usr/local/steemd-full/bin/cli_wallet -s wss://steemd.steemit.com`

@sneak @mvandeberg comments welcome as well

mvandeberg:
The wallet file being removed when the container is removed is fine. Maybe a note should be added about this behavior, but considering removing a docker container is an explicit action on behalf of the user, I do not view the behavior as a problem.

theoreticalbts:
I'm closing this PR due to inactivity.  Additionally, nothing in these comments has addressed my main concern with approving this PR.  Namely, that it's a bad idea to recommend users configure their system in a way that will result in the silent and automatic deletion of their wallet.

4Ykw:
I agree with one thing. Having another docker for just the cli_wallet would ease the pain of some people.

sneak:
Anyone who is experienced enough to use the cli_wallet can be trusted to manage their container lifecycles and volumes.

I don't want to spend much time on the cli_wallet.  Any advanced tooling around transaction creation should be focused on the python or js libraries.

sneak:
The docker container's entrypoint can be overridden so that it *only* runs cli_wallet and no other processes.  There is no need for a separate image.

jredbeard:
I think all that's necessary is the warning about your wallet.json being removed when the container is removed, and this one liner:

`docker run -it steemit/steem /usr/local/steemd-full/bin/cli_wallet -s wss://steemd.steemit.com`

The only frustrating thing about that is the 60s disconnect - because of that, it alternatively could be two lines:

```
docker run -it steemit/steem /bin/bash
/usr/local/steemd-full/bin/cli_wallet -s wss://steemd.steemit.com
```

That way you can easily restart it with the keys you've already imported if you miss the 60s disconnect time.

mvandeberg:
This is made irrelevant by #1444

PR #1145, opened 2017-06-01
Remove false comment

nathanhourt:
Removed comment claimed that `active_witnesses.size()` is exactly `STEEMIT_MAX_WITNESSES`, but this is flatly false, and the code immediately following it even iterates from `active_witnesses.size()` up to `STEEMIT_MAX_WITNESSES` which would obviously be a nop if that comment were correct. Also, the assertion the comment ostensibly refers to (line 184) asserts that `active_witnesses.size()` is either equal to `STEEMIT_MAX_WITNESSES` or something else, so obviously it is not guaranteed that these will be equal.


mvandeberg:
Fixed in 55a0e21

Issue #1148, opened 2017-06-02
Square Root Vote Weight Errors


mvandeberg:
https://github.com/steemit/steem/blob/master/libraries/chain/steem_evaluator.cpp#L1421 Should be `+=` and not `=`.

https://github.com/steemit/steem/blob/master/libraries/chain/steem_evaluator.cpp#L1527 `sqrt_weight` should also be set to 0 when a vote is modified.

https://github.com/steemit/steem/blob/master/libraries/chain/steem_evaluator.cpp#L1518
`c.total_sqrt_vote_weight -= cv.sqrt_weight` should be added.

PR #1151, opened 2017-06-02
Square Root Vote Weight Errors

mvandeberg:
PR for #1148


mvandeberg:
Each bug described in #1148 fails individually and passes when all are fixed.

Issue #1161, opened 2017-06-06
`update_tags` inefficient


mvandeberg:
In the tags plugin is the monolithic method `update_tags`. It is called on anything that can change a comment. It updates rshares, hot, and trending scores, etc. It also deserializes json metadata to check if tags were updated this. It does this recursively up to the root comment. json metadata can only change on `comment_operation` and only on the leaf comment. The deserialization should only happen in this case, and only once. Not for every parent comment up to the root. The inefficiency has been profiled to account for 5-10% of reindex time on a full node.

PR #1166, opened 2017-06-06
`update_tags` inefficient

mvandeberg:
PR for #1161


theoreticalbts:
Everything here looks functionally correct.  I have a few more minor enhancements, but I will address those in another PR.

Issue #1175, opened 2017-06-08
Refactor main branch walking logic from update_last_irreversible_block into its own function


theoreticalbts:
Related:  #1174

PR #1176, opened 2017-06-08
Refactor branch walk #1175

theoreticalbts:
PR for #1175

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -500,6 +500,22 @@ bool database::push_block(const signed_block& new_block, uint32_t skip)
    return result;
 }
 
+void database::_maybe_warn_multiple_production( const database& db, uint32_t height )const
+{
+   auto blocks = _fork_db.fetch_block_by_number( height );
+   if( blocks.size() > 1 )
+   {
+      vector< std::pair< account_name_type, fc::time_point_sec > > witness_time_pairs;
+      for( const auto& b : blocks )
+      {
+         witness_time_pairs.push_back( std::make_pair( b->data.witness, b->data.timestamp ) );
+      }
+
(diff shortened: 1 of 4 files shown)


libraries/chain/database.cpp
```diff
@@ -500,6 +500,23 @@ bool database::push_block(const signed_block& new_block, uint32_t skip)
...
+   auto blocks = _fork_db.fetch_block_by_number( height );
+   if( blocks.size() > 1 )
+   {
+      vector< std::pair< account_name_type, fc::time_point_sec > > witness_time_pairs;
+      for( const auto& b : blocks )
+      {
+         witness_time_pairs.push_back( std::make_pair( b->data.witness, b->data.timestamp ) );
+      }
+
+      ilog( "Encountered a block num collision due to a fork. Walking the current fork to determine the correct block. block_num:${n}", ("n", height) );
```
mvandeberg: We should combine these messages and make the logging accurate. Something like, "Encountered block num collision at block ${n} produced by witnesses ${w}" This also removes the statement that we are walking the fork, which we are not doing at this point in the code.

theoreticalbts:
Still not building.  Will fix the build and address the above comment momentarily.

libraries/chain/database.cpp
```diff
@@ -500,6 +500,22 @@ bool database::push_block(const signed_block& new_block, uint32_t skip)
    return result;
 }
 
+void database::_maybe_warn_multiple_production( const database& db, uint32_t height )const
```
mvandeberg: If this method is a member of database, why pass a const reference to itself?

theoreticalbts: Good catch, originally it wasn't, but then I saw `_fork_db` was private.

Issue #1174, opened 2017-06-08
get_blockchain_synopsis throws when reference_point is in the future


theoreticalbts:
This ticket is for a bug in `get_blockchain_synopsis()` which causes it to throw when `reference_point` is in the future (basically it calls `is_included_block()` which calls `get_block_id_for_num()` which throws if the blocknum isn't in the local block DB).  The p2p code isn't expecting a throw there and somehow it ends up causing the peer to disconnect.

theoreticalbts:
The fix here is to add a `find_block_id_for_num()` function which indicates failure by returning `block_id_type()` instead of throwing, then use that at the problematic call site in `is_included_block()`.

To cut down on duplicate code, I'm going to also re-implements `get_block_id_for_num()` by calling `find_block_id_for_num()` and throwing if the result is `block_id_type()`.

theoreticalbts:
Relevant log messages from production:

```
Jun 08 11:32:03 steemit-production-steemd-001_i-0141d0a9b911f945d messages: Jun  8 15:32:03 ip-172-31-12-104 docker/be993797e8f2[4378]: 1923092ms p2p        message_oriented_connection.cpp:188 read_loop            ] message transmission failed 10 assert_exception: Assert Exception
Jun 08 11:32:03 steemit-production-steemd-001_i-0141d0a9b911f945d messages: Jun  8 15:32:03 ip-172-31-12-104 docker/be993797e8f2[4378]: results.size() == 1: 
Jun 08 11:32:03 steemit-production-steemd-001_i-0141d0a9b911f945d messages: Jun  8 15:32:03 ip-172-31-12-104 docker/be993797e8f2[4378]:     {}
Jun 08 11:32:03 steemit-production-steemd-001_i-0141d0a9b911f945d messages: Jun  8 15:32:03 ip-172-31-12-104 docker/be993797e8f2[4378]:     th_a  database.cpp:257 get_block_id_for_num
Jun 08 11:32:03 steemit-production-steemd-001_i-0141d0a9b911f945d messages: Jun  8 15:32:03 ip-172-31-12-104 docker/be993797e8f2[4378]: 
Jun 08 11:32:03 steemit-production-steemd-001_i-0141d0a9b911f945d messages: Jun  8 15:32:03 ip-172-31-12-104 docker/be993797e8f2[4378]:     {"block_num":12644205}
[...]

PR #1177, opened 2017-06-08
Fix get_blockchain_synopsis() implementation #1174

theoreticalbts:
PR for #1174.  Depends on #1175.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -500,6 +500,22 @@ bool database::push_block(const signed_block& new_block, uint32_t skip)
    return result;
 }
 
+void database::_maybe_warn_multiple_production( const database& db, uint32_t height )const
+{
+   auto blocks = _fork_db.fetch_block_by_number( height );
+   if( blocks.size() > 1 )
+   {
+      vector< std::pair< account_name_type, fc::time_point_sec > > witness_time_pairs;
+      for( const auto& b : blocks )
+      {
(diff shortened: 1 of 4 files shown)


libraries/chain/database.cpp
```diff
@@ -246,22 +246,46 @@ bool database::is_known_transaction( const transaction_id_type& id )const
...
+block_id_type database::find_block_id_for_num( uint32_t block_num )const
 {
    try
    {
+      if( block_num == 0 )
+         return block_id_type();
+
+      // Reversible blocks are *usually* in the TAPOS buffer.  Since this
+      // is the fastest check, we do it first.
+      block_summary_id_type bsid = block_num & 0xFFFF;
```
mvandeberg: Should we make 0xFFFF a config constant? I noticed when we set up the block summary objects we hardcode 0x10000. Might make sense to remove those magic numbers.

theoreticalbts: The assumption that TaPoS is exactly 16 bits is already baked into various places in the code, e.g. [here](https://github.com/steemit/steem/blob/e6ef4ebdb5cd160b71f70526deb8273401d6c271/libraries/protocol/transaction.cpp#L72).

Making the mask a configurable constant means you'd have to review / update the code in all places that would need to be changed to make it a flexible value.  Which is so far off-topic, I'd say it should be a new ticket entirely.

mvandeberg: Fair enough, the code is correct and we can change it later. We should probably do an audit of the codebase for magic numbers as it is. I'll create a ticket for it.

PR #1198, opened 2017-06-19
formatted for easy copying, added new seed

feruzm:


Gandalf-the-Grey:
that would break things, it changes format that are expected by scripts
you might want to use `sed -e 's/^/seed-node = /' seednodes.txt`
to get format expected in `config.ini` file

feruzm:
Oh right, didn't realized that! ;)

Issue #1197, opened 2017-06-17
Subscription API


jnordberg:
Hi, I'm trying to use the subscription api over the websocket rpc. 

`set_block_applied_callback` works and provides a stream of 
```
{ previous: '00c4973217aed660f336c465337e3718c64e0f36',
      timestamp: '2017-06-16T23:14:45',
      witness: 'krnel',
      transaction_merkle_root: '5eefe40a36b2ba64ef28c3b05b36e9c53dfbd898',
      extensions: [],
      witness_signature: '205546f70accf4460402edd6e29c2a358dd48c015bb894ea4ae985ff8c574593e958af49a304a23105b2d2417de9107dc2cc92e805ed622aafa8de752468d60d6f' }
```

But the other two calls `set_subscribe_callback` and `set_pending_transaction_callback` are accepted but no events are sent. Is this a bug or does the node have to be configured somehow before events are sent?

theoreticalbts:
The `set_subscribe_callback` and `set_pending_transaction_callback` API calls are leftovers from the BitShares UI infrastructure.  They don't actually do anything in Steem.  We'll get rid of them in the next release.

jnordberg:
Okay, thanks for the clarification.

Any plans to add more subscription calls specific to Steem? It would be great to be able to subscribe to events for a specific user. And a block callback that sends the entire block would be useful as well (right now many applications I've seen just spam `get_dynamic_global_properties` to see if a new block has arrived)

theoreticalbts:
The dev team isn't going to be adding subscription semantics to steemd, and existing subscription calls are deprecated.  The main reason is that load balancing is an important part of making steemit.com function, and load balancing is much easier to implement if you don't have long-lived or stateful connections.

Another technological issue with subscriptions in general is, when the blockchain switches forks, some notifications that have already been sent may have been on the minority forks.  So any notification protocol must include a way for the sender to tell clients that they need to revert subscription data they've already received, and clients would have to properly respond to that condition.

jnordberg:
You're not planning on deprecating web sockets altogether? As a third party developer the web socket rpc is the killer feature of steem.

> Another technological issue with subscriptions in general is, when the blockchain switches forks, some notifications that have already been sent may have been on the minority forks.

[...]

PR #1200, opened 2017-06-20
Remove unused API's #1197

theoreticalbts:
Get rid of some non-functional leftovers from Graphene / BitShares.


mvandeberg:
Is there a reason to keep the on applied block callback?

theoreticalbts:
AFAICT it's still functional.  This PR is about removing code that doesn't function as advertised.

Issue #1193, opened 2017-06-15
Cleanup 19.0 Temp Vars


mvandeberg:
There are a number of fields and objects that are temporary for a smooth hardfork. These should be removed and optimized after the fork.

PR #1203, opened 2017-06-21
Cleanup 19.0 Temp Vars

mvandeberg:
PR for #1193


mvandeberg:
Do not merge this PR yet. As seen in #1217

0 delegation objects that were created pre-19 have the 30 day expiration which makes subsequent delegations adhere to the 30 day expiration as well. If the 0 delegation object were not created then the delegation would be returned after 7 days instead of 30. This difference in behavior could lead to a fork. We need to keep pre-19 0 delegation objects as part of consensus.

theoreticalbts:
OK, so we had some code that said:

```
if( _db.has_hardfork( STEEMIT_HARDFORK_0_19__971 ) )
{ /* new_assert_logic */ }
else
#warning( "TODO: Check and remove conditional after HF19" )
{ /* old_assert_logic */ }
```

This PR removes the `#warning` line, but the reason the `#warning` line was there was to remind us that after HF19 this code can be simplified to:

```
/* new_assert_logic */
```

The only reason not to simplify it is if, sometime before HF19, some operation got in the blockchain that passed `old_assert_logic` but does not pass `new_assert_logic`.  It should be fairly easy to check this -- do the simplification, and if you can't sync because `new_assert_logic` trips on something, we need to keep `old_assert_logic` around.

mvandeberg:
There is only one place where the warning was removed without changing the code. And that was because the warning was wrongly added to a block that could not be optimized.

theoreticalbts:
The `if / else` block in the pseudo-code above refers to [this code](https://github.com/steemit/steem/blob/4f5761f7e9660076345a26acae1af84740a0a193/libraries/chain/steem_evaluator.cpp#L2263-L2283).

theoreticalbts:
We tried it and the simplified logic doesn't sync, so the PR is good as-is.

PR #1206, opened 2017-06-22
Update brew syntax due to change in handling versions

mvandeberg:
Internal PR for #1185


theoreticalbts:
@mvandeberg Did you test this on a Mac?

mvandeberg:
I recently imaged my machine and had to set up my environment from scratch. This is the correct command for old versions of boost.

Issue #1181, opened 2017-06-09
Soft fork: check memo


abitmore:
Check transfer memos against the sender's and receiver's keys, when it's detected as a private key or password, reject the transfer and warn the user.

abitmore:
More info: https://steemit.com/steemit/@noisy/we-just-hacked-11-accounts-on-steemit-1158-sbd-and-8250-steem-is-under-our-control-but-we-are-good-guys-so

pfunks:
How much overhead would an implementation of this have on transfer transactions?

mvandeberg:
I am not concerned about the overhead. If it is a soft fork it only needs to be done when producing the block. What worries me is that if a tx is sent containing a private key or password, and is rejected, there is no guarantee it was rejected by a friendly node. They could save the transaction or a middle man could have seen it and saved it. An error message might lull them into a false sense of security that they narrowly avoided disaster. Granted, there needs to be a better way of informing users.

From a security perspective, this check should be performed in the wallet itself. So the tx never leaves the local machine. With the wallet and steemit.com both checking and reporting this mistake, we are going to catch nearly all occurrences.

abitmore:
I agree that wallet is a good place to do this. However, there are more wallet implementations, not only the cli_wallet and steemit.com, but also piston, eSteem, BUSY and etc. By using a soft fork, the node that got the transaction in the first place, aka the API service provider, will reject the transaction and return a error message, so the harm will end there, in most cases the user will have enough time to change password, and usually users will be using API services that they trust. An edge case is when a key is sent to a 3rd party, E.G. when withdrawing from exchanges, the 3rd party will get the error message but sometimes won't send them to the user. But it's same if done in wallet.

PR #1208, opened 2017-06-22
Soft fork: check memo

mvandeberg:
PR for #1181


theoreticalbts:
I'm worried about performance -- we're doing math on every memo.  We should be able to speed this up with a pre-test that excludes memos that don't have the same length as keys generated by steemd, or have characters outside the b58 character set.

Since the patch is functionally correct, I'm approving it.  Feel free to add the pre-test functionality to this ticket, or file as a separate ticket.

PR #1211, opened 2017-06-23
#301 Implement account history filter

NateBrune:


theoreticalbts:
This patch seems to be functionally correct.  A potential performance optimization filed as new ticket, #1226.

I was thinking a more modular architecture would improve both this functionality, and the existing code in the account history plugin for tracking specific accounts.  Since it affects multiple parts of the code, I'll file that as a new ticket.

mvandeberg:
Yes, while this patch isn't the most efficient, the savings on object creation should be a net win. The points brought up in the ticket are good and future optimizations should happen.

Issue #1215, opened 2017-06-26
Expiring(?) transactions, feedback/logging requested


roadscape:
This is a new issue which is being reported now that both the backend and frontend are under considerably higher load.

**It seems transactions are in pending state for extended periods of time -- and in some cases more than 60s, causing them to expire.**

Example A:
1. I voted on a post, it's accepted by the node, and my client state updates with the post's new payout (via `get_content`).
2. I refresh the page, the vote is gone, and no record of a tx after several minutes.

Example B:
1. I vote on a post, wait a few seconds.
2. Vote on another post, get assertion error `you can only vote once per block`.
3. Wait 30s, vote again, same error.
4. Wait another minute, vote again, finally goes through.

---

IMV - condenser and other apps have largely been operating on the assumption that if a tx is good to broadcast, and the account has enough bandwidth, it’s pretty much guaranteed to make it into a block within 60s. (15s used to work fine too, except for client timing issues -- a separate issue)

This is an important issue. We need to collect more data on why this is happening and how often, so it would be great if steemd could at least report on expired transactions as it drops them.

4Ykw:
I have seen 6 and 10 seconds.. not yet more... But I have not been looking at the full logs.
Although it happened me once and I thought it was just my network side. But then I have set it as "I need to wait". Is this not caused by delayed acks from other nodes?

PR #1221, opened 2017-06-27
Expiring(?) transactions, feedback/logging requested

mvandeberg:
PR for #1215


theoreticalbts:
Looks good, but it can't go in until it passes CI.  Why's it failing?

mvandeberg:
I changes exception types for expiration and tapos. To actually use the exception hierarchy that we created. Need to change the unit test to catch the right exception.

theoreticalbts:
Huh.  I was under the impression that the tests generally used catch-all blocks to catch those FC_ASSERT's and would continue to catch any more fine-grained STEEMIT_ASSERT

Issue #1213, opened 2017-06-25
**get_transaction** of database_api returns empty transaction


oflyhigh:
I want to use the API ***get_transaction*** to fetch transaction by trx_id.
But when I tried it, I got an empty transaction returned.

### Steps to reproduce:

1) get a transaction id at steemd.com, in this case: 5077183fd03e873b7080a0330eb2bf0da4a8a514
2) Fill out the API request and run it：
`curl https://steemd.steemitdev.com --data '{"jsonrpc": "2.0", "method": "call", "params": ["database_api", "get_transaction", ['5077183fd03e873b7080a0330eb2bf0da4a8a514']], "id": 1}'`
3) check the return messages: 
```
{"id":1,"result":{"ref_block_num":0,"ref_block_prefix":0,"expiration":"1970-01-01T00:00:00","operations":[],"extensions":[],"signatures":[],"transaction_id":"0000000000000000000000000000000000000000","block_num":0,"transaction_num":0}}
```

### Results
It returns an empty transaction for me.

I think the correct result should be:
```
{'expiration': '2017-06-25T01:42:21',
 'extensions': [],
 'operations': [['vote',
                 {'author': 'oflyhigh',
                  'permlink': 're-sweetsssj-happy-birthday-to-me-my-first-birthday-as-a-steemian-20170625t012421385z',
                  'voter': 'sweetsssj',
                  'weight': 500}]],
 'ref_block_num': 9518,
 'ref_block_prefix': 4153056001,
 'signatures': ['20232b570f87e0938c81017b423bbbd1c91b6ecc5d31e1bd7031e0602f4a9d3ba334dce971e4c3f1ab4934b814e44791448aa5e9f415b1ce40c8db24d12fea9fff']}
```

----

I'm not sure whether it is a BUG or I am using it in a wrong way.
Hope someone can check it and help me.
Thank you.

mvandeberg:
On steemit.com, this is intended. The API still supports looking up transactions by ID, but we have disabled it on steemit.com because it was dominating reindex times and taking up several gigabytes of space.

mvandeberg:
Reopening to change the semantics to throw an exception rather than fail silently.

oflyhigh:
Thank you.

PR #1230, opened 2017-06-27
#1213 Add fc_assert if SKIP_BY_TX_ID is enabled

NateBrune:


libraries/app/database_api.cpp
```diff
@@ -2357,6 +2357,7 @@ state database_api::get_state( string path )const
 annotated_signed_transaction database_api::get_transaction( transaction_id_type id )const
 {
 #ifdef SKIP_BY_TX_ID
+   FC_ASSERT( false, "This node's operator has disabled operation indexing by transaction_id" );
    return annotated_signed_transaction();
```
mvandeberg: Remove unreachable return

Issue #1222, opened 2017-06-27
Virtual op generated even when interest=0


roadscape:
Currently SBD APR is set to 0, but the node still generates a vop whenever the interest is due.

PR #1231, opened 2017-06-27
#1222 omit 0% interest vops

NateBrune:


theoreticalbts:
This changes op history numbering, but according to https://github.com/steemit/steem/issues/931 users cannot rely on that remaining the same on release upgrades or configuration changes.

PR #1243, opened 2017-07-07
Update exchange doc to be complete command

mvandeberg:
Previous command was correct but did not include all options that exchanges would likely want. Adding detail to the command so that there is no ambiguity in the documentation.


mvandeberg:
Keeping branch around so we can merge this to `master` after we move branches around.

Issue #1266, opened 2017-07-20
Enable content editing issues


feruzm:
I have tested content editing with steem-js, but so far no luck...

```
var json = ['enable_content_editing', {'account':'youracct','relock_time':time_in_future}];
var wif = "5PRIVATEACTIVEKEY";

steem.broadcast.customJson(wif, ["youracct"], [], "witness", JSON.stringify(json), 
	function(err, result) {
		console.log(err, result);
	});
```

Transaction attempts:
https://steemd.com/tx/9b68876ced13408d8a810f1b908cb958eacc60cb
https://steemd.com/tx/51ca10a798dc1fcd92ada51fa7a8d95405934e64
https://steemd.com/tx/d773e2231590db4beef7ed69df7c655efbee2340

Tried different timestamp formats (for `time_in_future`) in case parsing had issues, didn't help.

Transactions goes through but when edited content is submitted, I got error msg: `The comment is archived`

inertia186:
Still doesn't work.

inertia186:
I got it to work once.  Usually, I get `The comment is archived` but I was just testing a few minutes ago, I got this unexpected response instead:

```json
{
   "id":"3454b885d3dc82dccb395c3ab3d5af4f3700a2bf",
   "block_num":16542043,
   "trx_num":-1,
   "expired":true
}
```

Note, it says that `trx_num` is -1 and `expired` is true.  Of course, that block doesn't have that transaction.  It's also especially unexpected because I didn't figure anyone to be running this version, yet.

PR #1277, opened 2017-07-24
#1266 Register witness plugin interpreters and write test case

NateBrune:


mvandeberg:
Add one more test case the relocks the comment by timeout and make sure it cannot be edited.

Issue #1285, opened 2017-08-01
Absorb 60*60*24 into numerator in vote_evaluator


theoreticalbts:
To support `STEEMIT_VOTE_REGENERATION_SECONDS` not equal to an integer number of days, we need to move `60*60*24` from the denominator [here](https://github.com/steemit/steem/blob/134a52c6ad7d4905587d6984941d7d82ea496059/libraries/chain/steem_evaluator.cpp#L1176) to the numerator.  This should be done in a way that guarantees old code and new code generate identical `used_power` values when `STEEMIT_VOTE_REGENERATION_SECONDS` is an integer number of days.

Issue #259, opened 2016-08-10
Can not remove upvotes from own post.


Alifton:
I've now tried several times to remove my own upvote from my own post and here's the error I now get. This issue just started today to the best of my knowledge.  This has happened several times today on several different posts of mine where I forgot to uncheck the auto self-upvote box.

>Transaction failed: abs_rshares > 30000000: voting weight is too small, please accumulate more voting power or steem power

I know this isn't a lack of voting power because I had over 98%+ of my voting power both times I tried to remove my own accidental upvotes.

abitmore:
I don't think above commit fixed the issue. `abs_rshares` is not `1` in most cases when `voting_weight` is `0`. Why not just check if `o.weight == 0`?

mvandeberg:
Just a couple lines previous...

``` c++
if( abs_rshares == 0 ) abs_rshares = 1;
```

It is there to prevent a divide by zero other places. Decided it was best to have the assert only deal with abs_rshares rather than add in another variable.

abitmore:
@mvandeberg you didn't get it.

When unvoting, `abs_weight == 0`, 
thus `used_power = (current_power * abs_weight / STEEMIT_100_PERCENT) / 200 + 1 = 1`, 
thus `abs_rshares    = voter.vesting_shares.amount.value * used_power / STEEMIT_100_PERCENT` should > `1` when `vesting_shares.amount` is more than `2 * STEEMIT_100_PERCENT` aka `20,000`, which is true in most cases.

abitmore:
#132 is related

arhag:
Here is my suggestion for how to fix this problem:

Replace [this line](https://github.com/steemit/steem/blob/6958b0c6cce693684a24a7594b0b056f9dbcd234/libraries/chain/steem_evaluator.cpp#L882) with: 
```
used_power = (used_power/200);
if( !db().has_hardfork( STEEMIT_HARDFORK_0_13__259 ) )
   used_power += 1;
```

Replace [this line](https://github.com/steemit/steem/blob/6958b0c6cce693684a24a7594b0b056f9dbcd234/libraries/chain/steem_evaluator.cpp#L886) with:
```
if( !db().has_hardfork( STEEMIT_HARDFORK_0_13__259 ) && abs_rshares == 0 ) abs_rshares = 1;
```

[...]

PR #1286, opened 2017-08-01
Fix denominator in vote_evaluator computation #1285

theoreticalbts:
Guide to review:

- The first commit does the new computation but throws away the result, except for printing a message when it does not match the old computation.
- The second commit throws away the old computation and only does the new computation.

If we replay the blockchain with the first commit only, and verify that the message is not printed, then the PR is valid and ready to merge.  I'm currently replaying and verifying this.

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -1168,24 +1168,34 @@ void vote_evaluator::do_apply( const vote_operation& o )
 
    int64_t  abs_weight    = abs(o.weight);
    int64_t  used_power    = (current_power * abs_weight) / STEEMIT_100_PERCENT;
+   // Less rounding error would occur if we did the division last, but we need to maintain backward
+   // compatibility with the previous implementation which was replaced in #1285
+   int64_t  used_power_2  = ((current_power * abs_weight) / STEEMIT_100_PERCENT) * (60*60*24);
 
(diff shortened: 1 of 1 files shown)


theoreticalbts:
Effectively we have an expression of the form `a/(b*(c/d))`, where `c/d` has no remainder.  Since the outer division may have a remainder, we transform this to `round(a / (b*(c/d)) )` where `round()` is the rounding used (which was changed in #259).  Since `c/d` is the same in integer vs. rational-number division, we can transform the entire expression inside `round()` using normal (rational-number) algebraic manipulations.  Thus `round(a / (b*(c/d)) ) = round(a*d / (b*c))`.

The above line of reasoning proves that this patch is valid when applied with these substitiutions:

```
a = (current_power * abs_weight) / STEEMIT_100_PERCENT
b = dgpo.vote_power_reserve_rate
c = STEEMIT_VOTE_REGENERATION_SECONDS
d = (60*60*24)
```

NateBrune:
I feel like a unit test might be in order to make sure this doesn't happen again.

mvandeberg:
The unit test that checks vote power is sufficient to test this, specifically `opertation_tests/vote_apply` still has the old expression and the unit test passes, showing the expressions are equivalent (at least on the given input). Looks like a self-contained patch. I assume the reindex of this branch was successful?

theoreticalbts:
I'm pretty sure it was successful, but I suppose I could re-run it.

Issue #1078, opened 2017-04-28
Extraneous Multiindex Structs


mvandeberg:
In comment_object.hpp

``` c++
struct by_cashout_time; /// cashout_time
struct by_permlink; /// author, perm
struct by_root;
struct by_parent;
struct by_active; /// parent_auth, active
struct by_pending_payout;
struct by_total_pending_payout;
struct by_last_update; /// parent_auth, last_update
struct by_created; /// parent_auth, last_update
struct by_payout; /// parent_auth, last_update
struct by_blog;
struct by_votes;
struct by_responses;
struct by_author_last_update;
```

`by_cashout_time`, `by_permlink`, `by_root`, and `by_parent` are currently used as consensus orderings.

`by_last_update` and `by_author_last_update` are non-consensus orderings.

The other structs should be deleted. There are probably more such extraneous structs that should be cleaned up.

mvandeberg:
There are also a number of orderings that are not required for consensus. They should be removed or at least disabled for consensus only nodes.

PR #1289, opened 2017-08-03
1265 parallel api

mvandeberg:


theoreticalbts:
- "Remove unnecessary indices #1078" should be separate PR.
- Spot refactors of database_api in 69403684cc3da5b4248f0e4fe47b0fe712973d96 should be separate PR.
- Commits creating the `db_api` plugin should be their own PR.

mvandeberg:
Realistically, refactoring those commits into their own PRs and then rebasing this branch back on master is going to be significant work to get us where we are right now, but with a prettier git history. While I value having a pretty git history, it is not worth the minimum 1 day of work (possibly more) to get things there.

theoreticalbts:
I just finished going over that entire diff.

CI is succeeding, but failing to communicate with Github.  Error message:  "Could not update commit status, please check if your scan credentials belong to a member of the organization or a collaborator of the repository and repo:status scope is selected".

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "libraries/fc"]
```
sneak: should we vendor these in instead of using submodules?  especially considering that we are now pulling in an external one?

mvandeberg:
-[ ] Remove cpp-netlib submodule
-[ ] Check appbase cpp14 requirement. If required, update documentation. Defer?
-[ ] Move comment content patch into branch

mvandeberg:
`witness_plugin_impl` is now in the detail namespace. It fixes large parts of the diff, but not all because the witness plugin in master, has a mix of methods in the plugin class and impl. This version of the plugin has everything in impl. It would probably be easiest to review with a manual side by side inspection of the relevant methods.

theoreticalbts:
Approved.

PR #1350, opened 2017-08-07
Issue 991 final changes

vogel76:
1) price default constructor is allowed again, because of fc::variant::as method needs. 2) price(const asset&, const asset&) constructor is declared as explicit, and passed arguments must be filled. 
3) Added asset operator * ( const asset& a, const asset& b ), needed after blocking implicit convertion to price (previously asset operator *  ( const asset& a, const price& b ); was chosen).


vogel76:
Shall I also do rebase to master in this branch if its merge can be performed automatically ?

mvandeberg:
Call `validate()` in the explicit constructor. We use that logic in our operation validators, so it will be consistent.

Three operations contain prices, `feed_publish_operation`, `limit_order_create_operation`, and `limit_order_create2_operation`. Each of the operation validators call `validate()` on their respective prices, so if we call `validate()` in the constructor. Each use of price in database.cpp is related to the price feed (protected by `feed_public_operation` validator), limit orders (protected by validators) or vesting share price (enforced to be valid always). I am confident that we will not introduce a fork with this PR.

libraries/protocol/include/steemit/protocol/asset.hpp
```diff
@@ -78,13 +78,50 @@ namespace steemit { namespace protocol {
...
+        creation of invalid price object. \see validate() method.
+      */
+      explicit price(const asset& base, const asset& quote) : base(base),quote(quote)
+      {
+          /** Even price having base.amount == 0 is invalid too, probably this place
+          is located at too low level to make such checks.
+          Zero base does not affect this object sanity (in opposite to quote, which can lead to
+          division by zero during usage of this class methods).
+          */
+          FC_ASSERT(quote.amount != 0, "quote must be valid asset");
```
mvandeberg: Call validate()

vogel76: Calling a validate will result in throwing an exception in case the price object being constructed has invalid parts. This will lead to the same behavior as assert inside explicit constructor, and throwing an exception immediately during price construction instead of transaction commit as in original scenario (present in operation_tests/limit_order_create2_apply). Does such change is expected and acceptable ?

mvandeberg: I am assuming the tests that end up breaking are ones that create an invalid price object in an operation and check that validate fails. I am comfortable with the change so long as the only unit tests we need to edit are in relation to the three ops documented above.

mvandeberg: We can also change the test to assign `price()` and then modify the fields directly. Actually, I would prefer that.

theoreticalbts:
Can we rebase this into a single commit on `master`?

sneak:
github lets you do that in the PR - the button says "squash and merge".

PR #1353, opened 2017-08-08
Issue 834 - changing switch to visitor.

mariusztrela:


theoreticalbts:
- `auto& e` should be `const auto& e`
- Instantiate visitor once outside loop (avoid unnecessary instantiation)
- `_witness` visitor field should be `const string&` (avoid unnecessary string copy)

Issue #1288, opened 2017-08-03
Tags Plugin delete_comment visitor inefficient


mvandeberg:
<img width="1394" alt="screen shot 2017-08-03 at 16 15 23" src="https://user-images.githubusercontent.com/15834128/28941748-0430cdb2-7867-11e7-96f1-485fa845a154.png">

A process sample shows during a reindex that 97.56% of CPU thread time was spent in an iterator.

Looking into the specific code I found this.

``` c++
auto itr = idx.lower_bound( boost::make_tuple( auth.id ) );
while( itr != idx.end() && itr->author == auth.id )
```

The index in question is ordered by (author_id, comment_id, tag), yet is iterating starting at the author. The reason this algorithmic decision was made is because this visitor happens after the comment is deleted. It then checks every comment the author has created until it finds one that does not exist. The correct solution is to register another method on the pre-apply operation signal and go directly to the comment. The undo state on the entire database is made around each transaction and so if the op fails for whatever reason, the tags objects will be restored.

PR #1356, opened 2017-08-08
Tags Plugin delete_comment visitor inefficient #1288

mvandeberg:


theoreticalbts:
I'm not sure of the concurrent modification guarantees provided by the chain database.  So the deletion loop may not be correct.  A safer deletion loop would use one of the following two approaches:

- (a) Use `lower_bound()` to get a new iterator on each loop iteration
- (b) Use a two-pass approach where a read-only first pass saves ID's of objects to be deleted in a local variable, then an iterator-free second pass deletes the ID's

I think (a) will be slightly shorter in number of lines of code, and slightly better in performance, but I'd accept a patch using either approach.

mvandeberg:
We should check for this pattern everywhere in the codebase and replace it. It exists several other places and appears to be fine, but an alternative implementation that we know works is good with me.

I think I am in favor of the two pass approach similar to https://github.com/steemit/steem/blob/master/libraries/chain/database.cpp#L3850-L3866

It does not make sense to do `lower_bound()` every iteration because that is an O(lg n) call. So we have O(m lg n) execution where m is the specific number of items to remove. The two pass is O(lg n + 2m). One lg n lookup for `lower_bound()` and the two O(m) passes with the caveat that we now have a size m vector of const pointers.

PR #1362, opened 2017-08-09
Issue 834 - changing switch to visitor

mariusztrela:


mvandeberg:
The templated visitor method needs to be `const`.

Also, check spacing. We use 3 spaces for an indent. The code you refactored may have been wrong. If so, apologies.

libraries/chain/database.cpp
```diff
@@ -2688,55 +2688,61 @@ void database::_apply_block( const signed_block& next_block )
...
+        if( hfv.hf_version != signing_witness.hardfork_version_vote || hfv.hf_time != signing_witness.hardfork_time_vote )
+           _db.modify( signing_witness, [&]( witness_object& wo )
+           {
+              wo.hardfork_version_vote = hfv.hf_version;
+              wo.hardfork_time_vote = hfv.hf_time;
+           });
+    }
+
+    template<typename T>
+    void operator()( const T& unknown_obj )
```
mvandeberg: This method should be `const`

mariusztrela:
'const' in templated method - added
indents - fixed

PR #1363, opened 2017-08-09
Issue 1251 - Checking URL inside validate method.

mariusztrela:


libraries/protocol/include/steemit/protocol/steem_operations.hpp
```diff
@@ -422,6 +422,8 @@ namespace steemit { namespace protocol {
       public_key_type   block_signing_key;
       chain_properties  props;
       asset             fee; ///< the fee paid to register a new witness, should be 10x current block production pay
+      bool              has_hardfork;
```
mvandeberg: We cannot change operation serialization without causing a hardfork.

libraries/protocol/steem_operations.cpp
```diff
@@ -188,6 +188,17 @@ namespace steemit { namespace protocol {
    void witness_update_operation::validate() const
    {
       validate_account_name( owner );
+
+      if ( has_hardfork )
```
mvandeberg: The point of the previous change was to determine if there were any invalid operations prior to hardfork 0_1. There were not, so we can safely put the post hardfork logic in the validator without a hardfork check.

libraries/wallet/wallet.cpp
```diff
@@ -1622,6 +1622,7 @@ annotated_signed_transaction wallet_api::update_witness( string witness_account_
 {
    FC_ASSERT( !is_locked() );
 
+   steemit::chain::database& db = appbase::app().get_plugin< steemit::plugins::chain::chain_plugin >().db();
```
mvandeberg: The wallet does not have access to the database through appbase.

libraries/protocol/steem_operations.cpp
mariusztrela: I have to make sure:


1) I only remove below condition from 'witness_update_evaluator::do_apply' method

	if ( _db.has_hardfork( STEEMIT_HARDFORK_0_1 ) )
	{
		FC_ASSERT( o.url.size() <= STEEMIT_MAX_WITNESS_URL_LENGTH, "URL is too long" );
	}

2) I have to add below code to 'witness_update_operation::validate' method.

FC_ASSERT( o.url.size() <= STEEMIT_MAX_WITNESS_URL_LENGTH, "URL is too long" );

You said 'we can safely put the post hardfork logic in the validator without a hardfork check.', so
So, I will add only 'FC_ASSERT', without condition.

Is it correct?

mvandeberg: Looks right.

libraries/chain/steem_evaluator.cpp
```diff
@@ -65,11 +65,7 @@ void witness_update_evaluator::do_apply( const witness_update_operation& o )
 {
    _db.get_account( o.owner ); // verify owner exists
 
-   if ( _db.has_hardfork( STEEMIT_HARDFORK_0_1 ) )
-   {
-      FC_ASSERT( o.url.size() <= STEEMIT_MAX_WITNESS_URL_LENGTH, "URL is too long" );
-   }
-   else if( o.url.size() > STEEMIT_MAX_WITNESS_URL_LENGTH )
+   if( o.url.size() > STEEMIT_MAX_WITNESS_URL_LENGTH )
```
mvandeberg: Get rid of this if statement and block altogether. I will never be triggered.

theoreticalbts:
Let me say a few words about what is happening here.

- (1) We had an old rule in version 0.0, the old rule was that the URL length was not checked.
- (2) We decided to make a new rule, the URL length needs to be checked, in version 0.1.
- (3) Suppose someone made a transaction with a long URL, that is OK under the old rule -- but not OK under the new rule.  The network could have instantly split into two incompatible blockchains, one for people running version 0.0 including the long URL transaction, one for people running version 0.1 not including the long URL transaction.
- (4) Preventing this network split is the purpose of `has_hardfork()` check.  The version 0.1 code is written so that it does the same thing that version 0.0 code does, until the upgrade becomes active (the purpose of the `has_hardfork()` function is to check whether an upgrade has become active).
- (5) The upgrade becomes active once its activation date has passed and enough block producers have signaled support for the upgrade.
- (6) Once the upgrade is active on the network, newly created transactions / blocks are checked with the new rule.  Only historical transactions / blocks are checked with the old rule.  In other words, *the set of transactions / blocks checked with the old rule becomes fixed when the upgrade becomes active*.
- (7) Once this set of blocks is fixed, if no one made a long URL transaction that fails the old check, we can simplify the code by applying the new rule to both pre-upgrade and post-upgrade blocks.

Doing step (7) is what this ticket is about.

In other words:

- Before the activation date, the new version's code *must be functionally equivalent to the old rule on any possible user inputs (transactions)*.
- After the activation date, the new version's code *only needs to be functionally equivalent to the old rule on the historical user inputs (transactions) that actually occurred before the activation date*.

In this case the new code (check the length is within the maximum value) and the old code (don't check the length) are functionally equivalent because we're past the activation date, and we know the historical input didn't include any input over the maximum value.

I should also say a little bit about `validate()` vs `apply()` methods.  The `validate()` method is for checks that *do not need access to the database*, so `validate()` can only do things that do not reference the database.

The `apply()` method has full access to the database, so any check that needs access to the database must go in `apply()`.  The `has_hardfork()` method uses the database, so any check involving `has_hardfork()` must go in `apply()`.  However, the refactored version of the check no longer needs `has_hardfork()`, meaning the check can now be moved into `validate()`.

mariusztrela:
Thanx for explanations.

My little summary:

old rules -> ACTIVATION -> new rules

1) "old rules ":
	URL length is not checked.

2) ACTIVATION:
	Most witnessess( how many exactly? )decide: we make upgrade.
	New rule is introduced.
	Below condition returns 'true'.
		if ( _db.has_hardfork( STEEMIT_HARDFORK_0_1 ) )

3) " new rules "
	URL length is checked.


Finally we don't need check length of URL( in 'do_apply' method ), because:
- historical data( before activation ) don't need that checking
- actual data( after activation ) need that checking, but checking in 'validate' method is enough.
Why method 'validate'?
Because this method hasn't access to database with historical data.


Is it correct?

mvandeberg:
`validate()` is a state independent validation. It is the first level of filtering. We do this when the node is syncing or live because we have no idea what operations might show up, but we do not need to do this when reindexing from a local blockchain because it has already been cryptographically verified.

The check was made in `do_apply()` because the hardfork check is state dependent.

However, because all operations before the hardfork were valid under this check, we can make it state independent by moving it to `validate()`.

The first few hardforks were hardcoded to happen at certain times. Now hardforks are voted on by witnesses to happen at a certain time. When a 17/21 majority is reached the activation time is set by the witness votes.

Issue #1303, opened 2017-08-07
Appbase name semantic changes


mvandeberg:
Make appbase get_name virtual, name static, default const protected, remove name constructor, remove _name

- `get_name()` should be virtual
- `name()` should be a static method
- Default constructor should be protected
- Remove constructor that specifies a name
- Remove `_name` member.

All plugins should have a static method, `name()` that returns the plugin name and override `get_name()` so that it calls `name()`.

In appbase, anywhere that needs a plugin name should either call the virtual `get_name()` or static `name()` depending on whether it is dealing with an constructed plugin or not.

Plugins should all be instantiated with `make_shared` and indexed via static `name()`.

mvandeberg:
Instead of having each plugin override `get_name()`, I made `get_name()` final and have it call `name()` on the templated `Impl` class. That way, each plugin only needs to implement `name()` and cannot change the semantics of how the name is derived.

PR #1365, opened 2017-08-09
Appbase name semantic changes #1303

mvandeberg:
PR for #1303


theoreticalbts:
`name()` should return `const string&`

theoreticalbts:
Looks good

PR #1374, opened 2017-08-11
Issue 1339 - system time check.

mariusztrela:


libraries/plugins/p2p/p2p_plugin.cpp
```diff
@@ -271,73 +270,9 @@ bool p2p_plugin_impl::has_item( const graphene::net::item_id& id )
 }
 
 bool p2p_plugin_impl::handle_block( const graphene::net::block_message& blk_msg, bool sync_mode, std::vector<fc::uint160_t>& )
-{ try {
-   uint32_t head_block_num;
-
-   head_block_num = chain.db().head_block_num();
-
-   if (sync_mode)
```
mvandeberg: Keep this logging. The only thing that needed to move was the time check.

libraries/plugins/p2p/p2p_plugin.cpp
```diff
@@ -271,73 +270,9 @@ bool p2p_plugin_impl::has_item( const graphene::net::item_id& id )
...
-   uint64_t max_accept_time = now.sec_since_epoch();
-   max_accept_time += allow_future_time;
-   FC_ASSERT( blk_msg.block.timestamp.sec_since_epoch() <= max_accept_time );
-
-   try {
-      // TODO: in the case where this block is valid but on a fork that's too old for us to switch to,
-      // you can help the network code out by throwing a block_older_than_undo_history exception.
-      // when the net code sees that, it will stop trying to push blocks from that chain, but
-      // leave that peer connected so that they can get sync blocks from us
-      bool result = chain.db().push_block( blk_msg.block, ( block_producer | force_validate ) ? chain::database::skip_nothing : chain::database::skip_transaction_signatures );
```
mvandeberg: The p2p plugin should get a handle on the chain plugin in plugin_initialize. Add flag field to chain_plugin::accept_block if it does not have it. This line should call accept_block.

libraries/plugins/p2p/p2p_plugin.cpp
```diff
@@ -271,73 +270,9 @@ bool p2p_plugin_impl::has_item( const graphene::net::item_id& id )
...
-
-      if( !sync_mode )
```
mvandeberg: Keep all of this logging as well.

mariusztrela:
done

mariusztrela:
Problem with merging solved.

PR #1377, opened 2017-08-11
Configurable webserver-thread-pool-size in webserver_plugin/config files.

mkochanowicz:


libraries/plugins/webserver/webserver_plugin.cpp
```diff
@@ -34,6 +33,11 @@ using boost::asio::ip::tcp;
 using std::shared_ptr;
 using websocketpp::connection_hdl;
 
+#define WEBSERVER_DEFAULT_THREAD_POOL_SIZE 256
+#define WEBSERVER_THREAD_POOL_SIZE_OPTION "webserver-thread-pool-size"
```
mvandeberg: I don't think we have a need for the macro. Go ahead and inline the string in the boost options.

mvandeberg: Mainly because we do not want to be in the habit of change config option names. I like how the default value is set up.

libraries/plugins/webserver/webserver_plugin.cpp
```diff
@@ -134,19 +138,28 @@ class webserver_plugin_impl
...
+      ("webserver-http-endpoint", bpo::value< string >(), "The local IP and port to listen for incoming http connections.")
+      ("webserver-ws-endpoint", bpo::value< string >(), "The local IP and port to listen for incoming websocket connections.")
+      (WEBSERVER_THREAD_POOL_SIZE_OPTION, bpo::value<thread_pool_size_t>()->default_value(WEBSERVER_DEFAULT_THREAD_POOL_SIZE),
+      "Number of threads used to handle queries. Default: " BOOST_PP_STRINGIZE(WEBSERVER_DEFAULT_THREAD_POOL_SIZE) ".")
+      ;
 }
 
 void webserver_plugin::plugin_initialize( const variables_map& options )
 {
+   FC_ASSERT(options.count(WEBSERVER_THREAD_POOL_SIZE_OPTION) != 0, WEBSERVER_THREAD_POOL_SIZE_OPTION "is required!");
```
mvandeberg: If default value is set then it will always exist, even if not specified in the config file. This check is not needed.

libraries/plugins/webserver/webserver_plugin.cpp
```diff
@@ -134,19 +138,28 @@ class webserver_plugin_impl
...
+   auto thread_pool_size = options.at(WEBSERVER_THREAD_POOL_SIZE_OPTION).as<thread_pool_size_t>();
+   if (thread_pool_size <= 0)
```
mvandeberg: I would rather throw an exception and fail rather than silently ignore bad input.

mvandeberg:
@theoreticalbts Please review as well

libraries/plugins/webserver/webserver_plugin.cpp
mkochanowicz: FC_THROW will enough or additional log to elog?

libraries/plugins/webserver/webserver_plugin.cpp
mkochanowicz: Not needed until sb will remove default value. But OK.

mvandeberg: sb?

libraries/plugins/webserver/webserver_plugin.cpp
mvandeberg: FC_ASSERT( thread_pool_size > 0, "Thread pool size must be greater than 0" );

libraries/plugins/webserver/webserver_plugin.cpp
mkochanowicz: somebody who don't know plugin will crash without it

libraries/plugins/webserver/webserver_plugin.cpp
mkochanowicz: when i pass 0 such assertion results in "unknown exception"

mvandeberg: The catch block in main.cpp should have a catch for fc::exception that prints the error message.

mkochanowicz:
requested changes: done

programs/steemd/main.cpp
```diff
@@ -98,6 +98,10 @@ int main( int argc, char** argv )
    {
       std::cerr << boost::diagnostic_information(e) << "\n";
    }
+   catch ( const fc::exception& e )
```
mvandeberg: This is a merge conflict with `develop`

PR #1388, opened 2017-08-14
Issue 1313 - signal result storing.

mariusztrela:


libraries/chain/database.cpp
```diff
@@ -2583,6 +2583,13 @@ void database::show_free_memory( bool force )
    }
 }
 
+void database::disconnect_signal( boost::signals2::connection& signal )
```
mvandeberg: Could we move this into a util file.

Perhaps `chain/include/steemit/chain/util/signal.hpp`?

There is no reason this needs to belong in `database`. Make it inline and include the header in `database.hpp`.

@theoreticalbts Thoughts?

mariusztrela:
done

PR #1394, opened 2017-08-16
Issue 1335 - initialization with nullptr

mariusztrela:


mvandeberg:
I did not expand on this well in the ticket. Calling the api method should also be encapsulated in a nullptr check and fail safely if for whatever reason the call is still null.

mariusztrela:
done

PR #1396, opened 2017-08-16
Issue 1308, eliminate static_casts from appbase::plugin

vogel76:
Introduced new protected virt. methods to be reimplemented in plugin implementation: plugin_initialize, plugin_startup, plugin_shutdown, plugin_requires earlied being just provided by implementation via template argument.

### DIFF
--- a/libraries/plugins/market_history/include/steemit/plugins/market_history/market_history_plugin.hpp
+++ b/libraries/plugins/market_history/include/steemit/plugins/market_history/market_history_plugin.hpp
@@ -47,15 +47,17 @@ class market_history_plugin : public plugin< market_history_plugin >
 
       static const std::string& name() { static std::string name = STEEM_MARKET_HISTORY_PLUGIN_NAME; return name; }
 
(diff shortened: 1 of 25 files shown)


libraries/plugins/market_history/include/steemit/plugins/market_history/market_history_plugin.hpp
```diff
@@ -47,15 +47,17 @@ class market_history_plugin : public plugin< market_history_plugin >
...
          options_description& cli,
-         options_description& cfg );
-      void plugin_initialize( const variables_map& options );
-      void plugin_startup();
-      void plugin_shutdown();
+         options_description& cfg ) override;
 
-      flat_set< uint32_t > get_tracked_buckets() const;
-      uint32_t get_max_history_per_bucket() const;
+   protected:
```
mvandeberg: Is there a reason why these methods are protected? They aren't protected anywhere else.

mvandeberg:
@theoreticalbts Please review as well.

PR #1405, opened 2017-08-18
Issue 1327, remove 'mining' from debug_api.

vogel76:
Removed debug_mine method, definition of structure related to its argument.


mvandeberg:
I believe the key_storage and anything related to it can be removed as well.

PR #1408, opened 2017-08-21
Issue 1334 - JSON json_rpc_error initialization.

mariusztrela:


libraries/plugins/json_rpc/json_rpc_plugin.cpp
```diff
@@ -30,7 +33,7 @@ namespace detail
    {
       std::string                      jsonrpc = "2.0";
       fc::optional< fc::variant >      result;
-      fc::optional< json_rpc_error >   error;
+      json_rpc_error                   error;
```
mvandeberg: This field needs to be optional. By jsonrpc spec, the error field cannot exist unless there is an error.

mariusztrela:
Done

Issue #1278, opened 2017-07-26
Steem Testnet


mvandeberg:
Requirements for Steem Testnet

- Start with a large initial supply of 1,000,000,000.000 STEEM (verify this will not cause 64 bit overflow)
- On block 1, each hardfork with a date in the past should be automatically applied.
- If witnesses are not reporting a price feed, assume they are reporting 1 SBD/STEEM.\

All payouts will be on a shorter timescale and should already be configured that way for unit tests. Check other time based constants and shorted them if needed. This is to allow thorough and repeated testing of the various systems in real time. The chain id should not be set to `chain_id_type()`.

How the test net will work:

The init miner key is publicly known. Whoever starts the test net should transfer some of those funds to a privately controlled account (preferably linked to a faucet). Others can create accounts from the init miner account with whatever funds are left on the account.

STEEM on the testnet has the symbol TESTS.
SBD on the testnet has the symbol TBD.
VESTS are still VESTS.

theoreticalbts:
You should be able to handle the initial supply by setting `STEEMIT_INIT_SUPPLY` thanks to [this community-generated PR](https://github.com/steemit/steem/pull/1253).

I don't think it is good to have special testnet semantics for the price feed.  First of all, some unit tests (and potentially some tests performed on an actual testnet) should be testing what happens if there is no price feed -- adding special semantics will break those tests.

If you want price feeds on your testnet, you should publish them yourself via the normal mechanism.

theoreticalbts:
As well as code in steemd itself, there should be Python scripts for spinning up a testnet and handling things like automatically creating accounts for faucets / witnesses, causing the faucet (or some large VB you control) to vote for witness(es) you want, publishing feeds (if desired), generating dummy accounts/content/transactions/upvotes/trading or importing (perhaps some subset of) that from mainnet, etc.

Those scripts should probably live in a separate repo.  `debug_generate_blocks` API call may be useful.

mvandeberg:
The motivation behind having some default price feed mechanism is that there are several placers in the code that use the existence of the price feed as a milestone in the startup of the blockchain. Having an inconsistent price feed could lead to problems. Really though, those critical sections should be fixed so that losing the price feed does not harm the blockchain.

sneak:
> Those scripts should probably live in a separate repo.

I would like all testnet-related code to remain in the `steem` repo.

sneak:
[...]

PR #1409, opened 2017-08-21
Use STEEMIT_INIT_SUPPLY when reindexing

jnordberg:
This fixes a bug where a chain with a non-zero `STEEMIT_INIT_SUPPLY` would be unable to run a `--rebuild` if spending any of the `STEEMIT_INIT_SUPPLY`.

Refs #1225


jnordberg:
Refs #1278

Issue #1415, opened 2017-08-22
Implement MIN_BLOCK_SIZE constant


theoreticalbts:
The `block_log` corruption algorithm specified in #1414 works better if we can specify a lower bound on the size of a block.

This may be determined by instantiating a `signed_block` data structure and setting the `witness` member to the shortest possible account name.

Issue #1414, opened 2017-08-22
Better recovery of corrupt block_log


theoreticalbts:
This ticket is to create an algorithm to detect and recover in case `block_log` is corrupt, without any changes to the `block_log` format.  A wishlist of what we want to do when/if we redesign the `block_log` format exists in #844.

Let's make some definitions:

- For any file offset, define the *maximum previous block offset* to be the current offset minus the minimum block size.
- Define the *minimum previous block offset* to be the current offset minus the maximum block size.
- A *candidate block_pos* is an offset for which the 8-byte sequence at that offset, when interpreted as `uint64_t` (little endian), is between the minimum previous block offset and the maximum previous block offset.
- A *candidate chain* is a sequence of candidate block_pos, where each element of the chain is the next element's previous value.
- A *candidate N-chain* is a sequence of candidate block_pos, which either has length `N`, or proceeds to the beginning of the file.
- A candidate chain is said to be *dense* if its blocks deserialize successfully with no overlap or extra space between them.
- A chain is said to be *cryptographically linked* if each block's hash matches the next block's `previous` member.

The algorithm works as follows:

- Fix N to be a `#define`'d constant.
- Start at the end of the file.
- Scan backwards until you encounter a candidate N-chain.  
- Verify this N-chain is dense.
- Verify this N-chain is cryptographically linked.

theoreticalbts:
I'm not going to be able to get to this for a while, so I'm removing my assignment.

If anyone takes this over, feel free to use [this very incomplete code](https://github.com/steemit/steem/tree/1414-wip-block-log-recovery) as a starting point.  Or just start over entirely.

vogel76:
It could be worth to consider using also RocksDB as a storage for these data. It offers sanity checks, indexed access (through block number, blockID, whatever you want), **data compression**, like also online data replication what can be used to dump given file to external location without a need to stop server (what is impossible atm).

mvandeberg:
The design of the block log was chosen so that it would be easily digestible by other libraries that know the steem protocol binary serialization (all steem libraries need to know this for signing). While placing it in RocksDB certainly has disadvantages and reading data from it by another process is not impossible, it would be difficult to do so without having access to a RocksDB library for that language.

vogel76:
Problem you are describing above is caused by missing well defined interface between internal data (like internal format of block held in local storage) and its public representation. Such thing could be quite good done ie by [gRPC ](https://grpc.io/) library, but this is rather part of discussion to be done here:
[...]

PR #1420, opened 2017-08-23
Implement STEEMIT_MIN_BLOCK_SIZE constant #1415

theoreticalbts:
PR for #1415


libraries/chain/database.cpp
```diff
@@ -88,7 +88,11 @@ database_impl::database_impl( database& self )
    : _self(self), _evaluator_registry(self) {}
 
 database::database()
-   : _my( new database_impl(*this) ) {}
+   : _my( new database_impl(*this) )
+{
+   ilog( "Checking STEEMIT_MIN_BLOCK_SIZE matches expected value of ${m}", ("m", signed_block::calculate_min_size()) );
```
mvandeberg: Does this really belong in the database constructor? The application startup logic seems like a more logical place.

libraries/chain/database.cpp
```diff
@@ -2616,6 +2620,13 @@ void database::_apply_block( const signed_block& next_block )
       FC_ASSERT( block_size <= gprops.maximum_block_size, "Block Size is too Big", ("next_block_num",next_block_num)("block_size", block_size)("max",gprops.maximum_block_size) );
    }
 
+   if( block_size < STEEMIT_MIN_BLOCK_SIZE )
```
mvandeberg: Is this even necessary? A block size less than `STEEMIT_MIN_BLOCK_SIZE` cannot exist because the fields are not optional. If we want this check, I think it should at the very least exist in `push_block`, not `_apply_block` that way it only happens while syncing. Even better, it should be inside `#ifdef STEEMIT_TESTNET` (or whatever the macro is) so that we catch it on CI (there are plenty of empty blocks generated during our tests) without incurring the cost in production.

libraries/protocol/block.cpp
```diff
@@ -66,4 +66,12 @@ namespace steemit { namespace protocol {
       return checksum_type::hash( ids[0] );
    }
 
+   size_t signed_block::calculate_min_size()
```
mvandeberg: Because the method is static, the local variable, `b`, should also be static.

theoreticalbts:
> Is this even necessary? A block size less than STEEMIT_MIN_BLOCK_SIZE cannot exist because the fields are not optional.

That's how `STEEMIT_MIN_BLOCK_SIZE` is defined -- the smallest block that can exist.  So I can then write code which *assumes* valid blocks are always at least `STEEMIT_MIN_BLOCK_SIZE`.  Specifically, in the new `block_log` checking code I'm writing (ticket #1414), if the difference between two successive index file offsets is less than `STEEMIT_MIN_BLOCK_SIZE`, the index file must be corrupt.

> If we want this check, I think it should at the very least exist in push_block, not _apply_block that way it only happens while syncing.

OK, there are two different checks we need to do:  (1) A historical check that tells us if a too-small block never got into the chain in the past, (2) A consensus check that causes witnesses to actively reject too-small blocks.

I think (2) is actually sort-of dangerous to blockchain stability without any code in block generation to detect and enlarge too-small blocks, so I only implemented check (1).

> Even better, it should be inside #ifdef STEEMIT_TESTNET (or whatever the macro is) so that we catch it on CI (there are plenty of empty blocks generated during our tests) without incurring the cost in production.

The cost of a single integer comparison and `if` statement per block is probably too small to be measurable.  I want to run it on main net every time we sync, so if a too-small block ever does get into the main network, hopefully we will eventually notice the log message and redefine `SMT_MIN_BLOCK_SIZE`.  If we only ever run it on testnets, then it fails of its purpose, which is to inform us when a too-small block exists on mainnet.

mvandeberg:
The live/historic checks still seem unnecessary to me. It is the same thing as writing code that checks if we ever get a 16 bit integer in a 32 bit integer field. By definition of the language, it cannot happen. Even if the block were corrupt, the struct will always be STEEMIT_MIN_BLOCK_SIZE because it has to be by the C++ language spec. What we checking for is a violation of the programming language, not a bug in steemd.

mvandeberg:
And we are not just testing a single integer comparison. We are do raw pack on the block. That is not a cheap operation.

theoreticalbts:
> And we are not just testing a single integer comparison. We are doing raw pack on the block

We have to do this anyway for the maximum size check.

libraries/chain/database.cpp
theoreticalbts: I don't want to introduce a merge conflict with `develop`.

mvandeberg: Please make an issue for 0.19.X so that we can fix this once the merge is complete.

mvandeberg:
Ok. We should still make `calculate_min_size` save the block size as a static variable.

libraries/protocol/block.cpp
theoreticalbts: Why?

theoreticalbts:
> We should still make calculate_min_size save the block size as a static variable

Why?

libraries/protocol/block.cpp
mvandeberg: It is a static method and returns the same value every time. The returned value should also be static. Yes, we assert that a constant is equal to this value and then compare against the constant. Just because we are implementing this correctly now does not mean a future PR won't screw it up. These are things that easily slip through the cracks. I'd rather this method be done correctly and save us the hassle of potential problems in the future.

theoreticalbts:
The name of the function contains the word `calculate`, which means literally, figure it out by doing calculations.  If for some reason this function ends up being used in performance sensitive code, it should be obvious that (1) for performance reasons, the value needs to be cached, and (2) a method whose name includes the word `calculate` does not use any kind of caching, instead it CALCULATES the value.

The whole point of this PR is, instead of using caching, make the "cached" value available as a `#define` constant.  The recommended approach for any code that needs that constant is to use `STEEMIT_MIN_BLOCK_SIZE`, not call `calculate_min_size()`.  The only reason `calculate_min_size()` exists is to document how `STEEMIT_MIN_BLOCK_SIZE` was created.

I think I'm going to move `calculate_min_size()` to the unit tests, so it's clear that it's only for testing.

theoreticalbts:
> the struct will always be STEEMIT_MIN_BLOCK_SIZE because it has to be by the C++ language spec.

This is not true.  We are not checking the size of the struct (which itself can vary between compilers due to packing).  We are checking:

- (a) The *serialization* of a block
- (b) Whose fields are *initialized to values* which will result in the block being accepted.

Part (a) is defined by the FC serialization layer, not the compiler.  A change or bug in the FC serialization code could cause a block to be shortened beyond the minimum length.

Part (b) is defined by Steem itself.  For example, Steem has the following rule:

- The blockchain will not accept a block header with a `witness` field less than three characters (as defined by `STEEMIT_MIN_ACCOUNT_NAME_LENGTH`).

Even if you take FC serialization to be a "system level" component that we must assume to be bug-free (just as we must assume the compiler to be bug-free or we'd never be able to get anything done), part (b) is clearly a Steemd specific constraint, and a bug in steemd could conceivably cause it to be violated (for example if a block with a 0, 1 or 2 character `witness` field somehow got into the live blockchain).

Issue #1418, opened 2017-08-22
Implement SIGSEGV handler in steemd


theoreticalbts:
We need a config file option, enabled by default, which will cause `steemd` to install a `SIGSEGV` handler and dump a backtrace.

The Dockerized version should compile with line number information and use `addr2line` to get data.

Information [here](https://stackoverflow.com/questions/77005/how-to-generate-a-stacktrace-when-my-gcc-c-app-crashes) and [here](http://panthema.net/2008/0901-stacktrace-demangled/).

PR #1422, opened 2017-08-23
Enable backtrace on segfault #1418

theoreticalbts:
PR for #1418

### DIFF
--- a/libraries/app/application.cpp
+++ b/libraries/app/application.cpp
@@ -43,6 +43,7 @@
 #include <fc/rpc/api_connection.hpp>
 #include <fc/rpc/websocket_api.hpp>
 #include <fc/network/resolve.hpp>
+#include <fc/stacktrace.hpp>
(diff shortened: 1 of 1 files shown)


mvandeberg:
Please make a PR in fc so that we always point to `master` in fc.

theoreticalbts:
@mvandeberg I thought about doing that, but currently FC `master` has a bunch of parallel-api stuff, which I believe is non-backward-compatible.

mvandeberg:
I tired to make as much of that backwards compatible as possible. Did you try merging locally and seeing if it works?

mvandeberg:
I don't like having a branch that does not point to `fc/master` but there does appear to be some incompatibilities with `fc/master` and non-appbase steem branches. Considering that we will vendor in fc before the appbase release, I am going to less this one slide and merge with fc pointing to a branch.

PR #1427, opened 2017-08-24
1322 get block api

mariusztrela:


mariusztrela:
Corrections will be in another branch

Issue #1407, opened 2017-08-21
Not getting past hf5 for new chains


jnordberg:
When creating a new chain the active hardfork stays at hf 5. Noticed this since the `account_create_with_delegation` op fails with `_db.has_hardfork( STEEMIT_HARDFORK_0_17__818 ): Account creation with delegation is not enabled until hardfork 17`

```
------------------------------------------------------

            STARTING STEEM NETWORK

------------------------------------------------------
initminer public key: STX8Tu6Mx5iXjfVjQgfsTwRMJCTxcQGp6UShTFWi2Q5pr6bLPaguf
chain id: 79276aea5d4877d9a25892eaa01b0adf019d3e5cb12a97478df3298ccdd01673
blockchain version: 0.19.1
------------------------------------------------------
2901421ms th_a       main.cpp:172                  main                 ] parsing options
2901421ms th_a       main.cpp:174                  main                 ] initializing node
2901422ms th_a       main.cpp:176                  main                 ] initializing plugins
2901422ms th_a       application.cpp:1140          initialize_plugins   ] Initializing plugin account_by_key
2901422ms th_a       account_by_key_plugin.cpp:256 plugin_initialize    ] Initializing account_by_key plugin
2901422ms th_a       application.cpp:1140          initialize_plugins   ] Initializing plugin account_history
2901422ms th_a       application.cpp:1140          initialize_plugins   ] Initializing plugin follow
2901422ms th_a       follow_plugin.cpp:360         plugin_initialize    ] Intializing follow plugin
2901422ms th_a       application.cpp:1140          initialize_plugins   ] Initializing plugin market_history
2901422ms th_a       market_history_plugin.cpp:177 plugin_initialize    ] market_history: plugin_initialize() begin
2901422ms th_a       market_history_plugin.cpp:192 plugin_initialize    ] bucket-size [15,60,300,3600,86400]
2901422ms th_a       market_history_plugin.cpp:193 plugin_initialize    ] history-per-size 1000
2901422ms th_a       market_history_plugin.cpp:195 plugin_initialize    ] market_history: plugin_initialize() end
2901422ms th_a       application.cpp:1140          initialize_plugins   ] Initializing plugin raw_block
2901422ms th_a       application.cpp:1140          initialize_plugins   ] Initializing plugin tags
2901422ms th_a       tags_plugin.cpp:525           plugin_initialize    ] Intializing tags plugin
2901422ms th_a       application.hpp:105           register_api_factory ] name: tag_api
[...]

PR #1430, opened 2017-08-24
Add STEEMIT_INIT_HARDFORK config option

jnordberg:
This allows for a new chain to be started with all the hardforks applied

Refs #1407


jnordberg:
Also ref #1278

mvandeberg:
Our test net code handles this differently.

PR #1431, opened 2017-08-24
Add STEEMIT_INIT_SBD_SUPPLY config option

jnordberg:
Allows you to create an initial steem dollar supply on a new chain

Refs #1278


mvandeberg:
This breaks the invariant `virtual_supply = steem_supply + sbd_supply * price_feed` because the price feed is not defined at genesis.

PR #1437, opened 2017-08-25
1322 get block new api

mariusztrela:


mariusztrela:
pull request is referenced to bad branch

PR #1438, opened 2017-08-25
Issue 1322 - get block API

mariusztrela:


mariusztrela:
I launched:
curl --data '{"jsonrpc": "2.0", "method": "call", "params": ["block_api","get_block", {"block_num":23} ], "id": 1}' http://127.0.0.1:9876/rpc
It works...

Issue #1340, opened 2017-08-07
Chain API


mvandeberg:
We should create a chain api that can push blocks and transactions.

mvandeberg:
Two api calls should be `push_block` and `push_transaction`. Exceptions thrown during those calls should be caught by the API and returned as a valid response rather than as a jsonrpc error.

Return type is probably a struct with a `bool success`, and `optional< string > error`;

vogel76:
This new API shall be part of  new module under plugin/apis/chain_api or part of chain plugin ?

mvandeberg:
plugins/apis/chain_api

mkochanowicz:
Need info how new api/api_plugin could/should be tested in dev env.

mkochanowicz:
#1441

PR #1441, opened 2017-08-25
create chain_api/chain_api_plugin

mkochanowicz:
#1340

### DIFF
--- a/libraries/plugins/apis/chain_api/chain_api.cpp
+++ b/libraries/plugins/apis/chain_api/chain_api.cpp
@@ -8,14 +8,14 @@ namespace detail {
 class chain_api_impl
 {
    public:
-      chain_api_impl() : db( appbase::app().get_plugin<steemit::plugins::chain::chain_plugin>().db() ) {}
+      chain_api_impl() : _chain( appbase::app().get_plugin<chain_plugin>() ) {}
 
       DECLARE_API(
          (push_block)
          (push_transaction) )
 
-   public:
-      steemit::chain::database& db;
+   private:
+      chain_plugin& _chain;
 };
 
 DEFINE_API( chain_api_impl, push_block )
@@ -26,7 +26,7 @@ DEFINE_API( chain_api_impl, push_block )
 
    try
    {
-      db._push_block(args);
+      _chain.accept_block(args.block, args.currently_syncing);
       result.success = true;
    }
    catch (const fc::exception& e)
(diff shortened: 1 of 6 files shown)


libraries/plugins/apis/chain_api/chain_api.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+DEFINE_API( chain_api_impl, push_block )
+{
+   push_block_return result;
+
+   result.success = false;
+
+   try
+   {
+      db._push_block(args);
```
mvandeberg: Call accept_block on the chain_plugin

libraries/plugins/apis/chain_api/chain_api.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+DEFINE_API( chain_api_impl, push_transaction )
+{
+   push_transaction_return result;
+
+   result.success = false;
+
+   try
+   {
+      db._push_transaction(args);
```
mvandeberg: Call accept_transaction on the chain_plugin

libraries/plugins/apis/chain_api/chain_api.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+   JSON_RPC_REGISTER_API(
+      STEEM_CHAIN_API_PLUGIN_NAME,
+      (push_block)
+      (push_transaction) );
+}
+
+DEFINE_API( chain_api, push_block )
+{
+   return _my->db.with_write_lock( [&]()
```
mvandeberg: We do not need this lock because chain_plugin locks properly.

libraries/plugins/apis/chain_api/chain_api.cpp
```diff
@@ -0,0 +1,103 @@
...
+   {
+      return _my->push_block( args );
+   });
+}
+
+DEFINE_API( chain_api, push_transaction )
+{
+   return _my->db.with_write_lock( [&]()
```
mvandeberg: Do not need this lock either

libraries/plugins/apis/chain_api/include/steemit/plugins/chain_api/chain_api.hpp
```diff
@@ -9,7 +9,12 @@ namespace steemit { namespace plugins { namespace chain {
 
 namespace detail { class chain_api_impl; }
 
-typedef steemit::chain::signed_block push_block_args;
+struct push_block_args
+{
+   steemit::chain::signed_block block;
+   bool                         currently_syncing;
```
mvandeberg: Default initialize this field to false. It is used to determine what log level the chain plugin should use when pushing the block.

mkochanowicz:
done

PR #1456, opened 2017-08-29
Initial SMT operation skeleton

theoreticalbts:
Skeleton of some SMT operations with evaluators.


libraries/chain/smt_evaluator.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <steemit/chain/util/reward.hpp>
+
+#include <steemit/protocol/smt_operations.hpp>
+
+namespace steemit { namespace chain {
+
+void smt_setup_evaluator::do_apply( const smt_setup_operation& o )
+{
+   FC_ASSERT( _db.has_hardfork( STEEMIT_HARDFORK_0_21 ), "SMT functionality not enabled until hardfork 20" );
```
mvandeberg: Create an define for STEEMIT_SMT_HARDFORK as STEEMIT_HARDFORK_0_21 so we can easily delay SMTs if needed.

theoreticalbts:
How does this look?

PR #1459, opened 2017-08-29
Bump version to 0.19.2

mvandeberg:


sneak:
regarding #1465 - IMO this is the wrong approach; do not test for the specific private key of the user - just test for a string that fits the format of any private key in WIF.

ultimately, though - does this code really belong in steemd? I think this should stay out of the daemon and stay firmly in the UI.

It certainly should not block a release in any case, @theoreticalbts.

PR #1461, opened 2017-08-30
Add missing case to memo protection code

theoreticalbts:


theoreticalbts:
Closing, need to create new separate PR's for `master` and `develop` to un-bork CI.

sneak:
IMO this is the wrong approach; do not test for the specific private key of the user - just test for a string that fits the format of *any* private key in WIF.

sneak:
(ultimately, though - does this code really belong in `steemd`? I think this should stay out of the daemon and stay firmly in the UI.)

PR #1465, opened 2017-08-30
Add missing case to memo protection code (master branch, take 2)

theoreticalbts:
Supersedes #1461


mvandeberg:
There has been no discussion on this or #1466 in several months. I think it was decided that if we want to add more checks against keys, we should simply do a regex check rather than against specific keys.

PR #1462, opened 2017-08-30
Add missing case to memo protection code (develop branch)

theoreticalbts:
Identical to #1461 but against `develop`

PR #1466, opened 2017-08-30
Add missing case to memo protection code (develop branch, take 2)

theoreticalbts:
Supersedes #1462


mvandeberg:
Is this (and #1465) something we still want to add?

I think @sneak had a point in #1459

> do not test for the specific private key of the user - just test for a string that fits the format of any private key in WIF.

I think this PR should be closed and an issue created to refactor this to a broad WIF check.

mvandeberg:
There has been no discussion on this or #1465 in several months. I think it was decided that if we want to add more checks against keys, we should simply do a regex check rather than against specific keys.

Issue #1457, opened 2017-08-29
Create step-by-step example of adding an operation to Steem


theoreticalbts:
We have some new backend devs who will be helping implement operations.  To help explain how to do this, I will add a new operation as an example, showing each step along the way.

Kiwonik:
Questions regarding further implementation of `smt_elevate_account_operation`:
1. It seems to us that the `fee` member of `smt_elevate_account_operation` is redundant in terms of actual amount, that is established by evaluator on its own (obtained from db). However the member still carries important information about the asset_symbol_type the client wants to use. Thus the `asset fee` can be reduced to `asset_symbol_type fee_symbol`.
2. Is the actual fee transfer supposed to be a part of `smt_elevate_account_evaluator::do_apply` code? Or rather a prior separate transfer operation that must be only identified in the account history in said `do_apply`. Assuming the former, shall we use `transfer_operation` with its evaluator, or reuse (aka copy-paste) the code of `transfer_evaluator::do_apply` or rather put its body into separate function called from both transfer and elevate evaluator?

theoreticalbts:
> the fee member of `smt_elevate_account_operation` is redundant in terms of actual amount, that is established by evaluator on its own (obtained from db)

Right now the fee is constant, but later we may create a process for changing the fee (based on witness votes).  If the fee isn't given in the operation, and let's say the fee is raised someday from $1000 to $5000, a user creating an account may have their UI show the fee will be $1000 -- which is accurate at the time -- and then be very upset when the transaction he authorized gets into the blockchain a few blocks later (after the change takes effect) and $5000 is deducted from his account.

What should happen in this case is the transaction signed by the user should say "I authorize paying a fee of $1000" and then if the blockchain requires a higher fee at the moment the transaction is processed, the transaction will be rejected and never happen.

The second issue is that the user's account history should show how much was paid when they elevated in the past.  If the fee is implicitly contained in the database, we would have to create a second, virtual operation to document how much was deducted.  This is possible but more complex, in our design of operations we sometimes deliberately include a few extra bytes of redundant information if it makes the operation a better fit for our architecture.

theoreticalbts:
> Is the actual fee transfer supposed to be a part of smt_elevate_account_evaluator::do_apply code?

Yes, it's just not implemented yet.

[...]

PR #1467, opened 2017-08-30
Step by step documentation of new operations #1457

theoreticalbts:
PR for #1457.  This PR is a work in progress, not ready for merging yet.

### DIFF
--- a/tests/tests/smt_tests.cpp
+++ b/tests/tests/smt_tests.cpp
@@ -14,8 +14,98 @@
 using namespace steemit::chain;
 using namespace steemit::protocol;
 using fc::string;
+using boost::container::flat_set;
 
 BOOST_FIXTURE_TEST_SUITE( smt_tests, clean_database_fixture )
+
+BOOST_AUTO_TEST_CASE( elevate_account_validate )
+{
+   try
+   {
+      smt_elevate_account_operation op;
+      op.account = "@@@@@";
+      op.fee = ASSET( "1.000 TESTS" );
+      STEEMIT_REQUIRE_THROW( op.validate(), fc::exception );
+
+      op.account = "alice";
+      op.validate();
+
+      op.fee.amount = -op.fee.amount;
+      STEEMIT_REQUIRE_THROW( op.validate(), fc::exception );
+
+      // passes with MAX_SHARE_SUPPLY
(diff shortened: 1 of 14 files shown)


theoreticalbts:
Some TODO items:

- Needs to be rebased after #1464 #1469 are merged to `develop`
- Needs to check for non-sufficient funds
- Needs to actually change balances
- Unit test needs to check balance has changed

tests/tests/smt_tests.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "../common/database_fixture.hpp"
+
+using namespace steemit::chain;
+using namespace steemit::protocol;
+using fc::string;
+
+BOOST_FIXTURE_TEST_SUITE( smt_tests, clean_database_fixture )
+
+BOOST_AUTO_TEST_CASE( elevate_account )
```
mvandeberg: This does not match the test structure for other ops. Each op has a validate, authority, and apply unit test.

Claim reward balance is a good example of how this should be formatted.

[validate](https://github.com/steemit/steem/blob/master/tests/tests/operation_tests.cpp#L5781-L5834)
[authorities](https://github.com/steemit/steem/blob/master/tests/tests/operation_tests.cpp#L5836-L5859)
[apply](https://github.com/steemit/steem/blob/master/tests/tests/operation_tests.cpp#L6044-L6140)

Kiwonik: As this request is a blocker to further SMT development we'd like to take it over and implement it in dedicated `smt-develop` branch, that is already created and has the `1457-step-by-step` merged to. Do you mind us going for it?

libraries/protocol/include/steemit/protocol/config.hpp
```diff
@@ -43,6 +42,8 @@
 #define STEEMIT_OWNER_AUTH_HISTORY_TRACKING_START_BLOCK_NUM 1
 #else // IS LIVE STEEM NETWORK
 
+#define STEEMIT_BLOCKCHAIN_VERSION              ( version(0, 19, 0) )
```
mvandeberg: Changes in develop mean that we do not need to have a testnet version and mainnet version that are different.

theoreticalbts:
Fixed the build, it was missing `#include` file and wrong index name

PR #1485, opened 2017-08-31
Update README.md

mvandeberg:
Update the README for 0.19.2 including links to the improved quickstart guide.


theoreticalbts:
Misspelled commnad line argument:  `--shard-file-dir=/path`

theoreticalbts:
We should probably also have a PR for `develop`.

mvandeberg:
When develop gets merged to master, it will all get handled. Right now, anything that doesn't need to be in both right now should just stay where it is at. Besides, appbase is going to require documentation updates of its own.

Issue #1312, opened 2017-08-07
Plugin pointer type consistency


mvandeberg:
Check for `shared_ptr`/`unique_ptr` type, naming, and visibility consistency across plugins.

mariusztrela:
done

PR #1490, opened 2017-09-01
Issue #1312 - plugin pointer type consistemcy

mariusztrela:
#1312


mvandeberg:
Merge conflicts

theoreticalbts:
When do we use `shared_ptr` vs `unique_ptr`?  In this patch some use `unique_ptr` (`account_by_key_plugin`, `account_history_plugin`) while others use `shared_ptr` (`block_api`, `chain_api`, `condenser_api`).

I thought the point of the issue was to pick one of these two pointer types and use it for `my` member of all plugins.

mvandeberg:
I think following the impl pattern should all be private unique pointers.

The APIs should be shared pointers because, unlike FC APIs, these APIs are singletons.

mariusztrela:
Done.

PR #1491, opened 2017-09-01
api_message_object for DB message_object - fix #1318

### LINKED ISSUE #1318: API object naming
Check that API object naming is consistent.

`api_*_object` for a DB object that would be returned in an API.

mkochanowicz:


mvandeberg:
If the private message plugin was the only place where this was a problem, then I am inclined to simply close the issue without a merge. The private message plugin has never been used in production and is really only sticking around because I have been too lazy to delete it.

Issue #1179, opened 2017-06-08
Add block to free memory message


mvandeberg:
When getting a free memory message live, you know the block number, but when reindexing there is only a range. It would be nice to have exact block numbers so we can accurately track memory usage over time.

A reindex does show us free memory every 100k blocks, but the extra data points would be nice.

vogel76:
Could you point us to some place in the code related to free memory messaging to have some startpoint for further analysis ?

mvandeberg:
In database.cpp anywhere `get_free_memory` is called.

mariusztrela:
"...there is only a range"
What do you mean by saying 'range'?
There is only info abouts blocks:  "double( cur_block_num * 100 ) / last_block_num << "%   " << cur_block_num << " of " << last_block_num"

"...extra data points would be nice"
What points? Do you want to remove 'if( cur_block_num % 100000 == 0 )' condition ?

mvandeberg:
I wrote the original message poorly.

On reindex we get free memory every 100000 blocks. It would be nice on the generic "Free memory is now" messages to have the head block num.

mvandeberg:
Earlier blocks we have more data points because memory is not used quickly. Now, 100k can eat up 6-7Gb. It would be nice to have those block numbers to better track our memory growth.

mariusztrela:
Done, but this issue is not assigned to any project.

PR #1500, opened 2017-09-04
Issue #1179 - memory message improved.

mariusztrela:
#1179

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -182,7 +182,7 @@ void database::reindex( const fc::path& data_dir, const fc::path& shared_mem_dir
             if( cur_block_num % 100000 == 0 )
                std::cerr << "   " << double( cur_block_num * 100 ) / last_block_num << "%   " << cur_block_num << " of " << last_block_num <<
                "   (" << (get_free_memory() / (1024*1024)) << "M free)\n";
-            apply_block( itr.first, skip_flags );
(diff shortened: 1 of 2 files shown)


libraries/chain/database.cpp
```diff
@@ -2497,7 +2497,7 @@ void database::set_flush_interval( uint32_t flush_blocks )
 
 //////////////////// private methods ////////////////////
 
-void database::apply_block( const signed_block& next_block, uint32_t skip )
+void database::apply_block( const signed_block& next_block, uint32_t skip, uint32_t current_block_num )
```
mvandeberg: Calculating the block num is cheap. Use next_block.block_num()

libraries/chain/include/steemit/chain/database.hpp
```diff
@@ -422,7 +422,7 @@ namespace steemit { namespace chain {
       private:
          optional< chainbase::database::session > _pending_tx_session;
 
-         void apply_block( const signed_block& next_block, uint32_t skip = skip_nothing );
+         void apply_block( const signed_block& next_block, uint32_t skip = skip_nothing, uint32_t current_block_num = 0 );
```
mvandeberg: This is bad. Many places that currently call apply_block would now have broken error messages. Future code may begin to rely on that value instead of grabbing the block num from the block itself and break because of it.

mariusztrela:
done

PR #1502, opened 2017-09-04
Update CORS headers

jnordberg:
Add Access-Control-Allow-Methods and Access-Control-Allow-Headers headers

@jredbeard


mvandeberg:
When approved by Justin, I will merge.

jnordberg:
Verified that the config works

Issue #1331, opened 2017-08-07
Document plugin startup and asynchronous behavior


mvandeberg:
We need documentation describing how plugin startup works as well as describing the asynchronous nature of APIs and best practices for API developers.

PR #1530, opened 2017-09-07
Document plugin startup and asynchronous behavior

mvandeberg:
PR for #1331


theoreticalbts:
This needs to be rebased on latest `develop`

PR #1535, opened 2017-09-08
Add reblog API in cli_wallet

VIM-Arcange:
Reblog operation should be exposed via cli_wallet


mvandeberg:
The follow plugin behavior is being migrated to an external service from steemd. While the cli wallet has easy access to the custom ops now, it will not for long. If you want to programmatically provide this behavior I recommend submitting the transaction via `sign_transaction`.

VIM-Arcange:
As we talked about at Steemfest, reopening this PR

Issue #1445, opened 2017-08-25
Clang-tidy linter


mvandeberg:
Integrate clang-tidy linter in our cmake build process.

I am creating four levels of linting.

OFF (Default) - Does nothing
LOW - Default level of linting
HIGH - Default plus some more useful checks
ALL - All checks

This requires CMake 3.6+ and Ubuntu 16.04 only comes with CMake 3.5, so it will not work in our Docker image unless we update what version the base image uses. The config code in CMake only allows linting if CMake is 3.6 or above.

PR #1537, opened 2017-09-08
Clang-tidy linter

mvandeberg:
PR for #1445


theoreticalbts:
@mvandeberg Not having a working `clang` setup, I have no way to check that this functions as advertised and doesn't break anything.  I just based my review on "the diff looks reasonable and it passes CI," as a `clang` user it's up to you to verify this patch is functional.

mvandeberg:
It does work. And it should only attempt to build with the linter if clang can be found. It is worth noting, that Ubuntu 16.04 does not have a new enough version of cmake required by this patch, so passing CI demonstrates it does not break anything that currently exists.

PR #1538, opened 2017-09-11
validate_invariants is now optional, not called by default - fix #1477

### LINKED ISSUE #1477: full node - unexpected, occassional slow start from saved state
Use case scenario:
Running full node with all bells and whistles
```
du -h /run/fullnode/shared_memory.bin
108G    /run/fullnode/shared_memory.bin
```
Performance notes: `shared_memory_file` is on `tmpfs`, backed by 64GB physical RAM, swapped to RAID1 on SSDs when needed. Dedicated machine, `Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz`
Replay finishes after 5 hours so it's relatively good for this job.
I'm stopping node. After short time (a minute or three) starting it back up again.
```
3322023ms th_a       block_log.cpp:130             open                 ] Log is nonempty
3322023ms th_a       block_log.cpp:139             open                 ] Index is nonempty
```
hangs at that ^ state for considerable amount of time
```
616596ms th_a       database.cpp:2569             show_free_memory     ] Free memory is now 12G
```
it continues roughly after 15 minutes (!)
huge swap in/out = huge disk IOs

Sometimes it goes much quicker (115 seconds). I don't have statistics what's more frequent because it's pretty rare scenario to shutdown a node, however it might also affect startup time in auto-scalling environments that are using saved state at creation of new instances.

mkochanowicz:


libraries/chain/database.cpp
```diff
@@ -123,7 +123,8 @@ void database::open( const fc::path& data_dir, const fc::path& shared_mem_dir, u
             undo_all();
             FC_ASSERT( revision() == head_block_num(), "Chainbase revision does not match head block num",
                ("rev", revision())("head_block", head_block_num()) );
-            validate_invariants();
+            if (_validate_invariants)
```
mvandeberg: I'm not sure I like using an instance variable for this when it is only used in this one function. I think adding it is a parameter to open with default initialization to false makes more sense.

PR #1539, opened 2017-09-11
cli_wallet should expose reblog API

VIM-Arcange:
Added reblog declaration to header
This PR is in relation with PR #1535


VIM-Arcange:
Code now compile correctly but I get the following error when calling the reblog API from recompiled cli_wallet:

unlocked >>> reblog "arcange" "some-username" "some-permlink" false
```
itr != _by_name.end(): no method with name 'reblog'
```

How to add the 'reblog' function to the list of valid methods?

mvandeberg:
The follow plugin behavior is being migrated to an external service from steemd. While the cli wallet has easy access to the custom ops now, it will not for long. If you want to programmatically provide this behavior I recommend submitting the transaction via `sign_transaction`.

Issue #1321, opened 2017-08-07
check_reflect.py should look at plugin objects


mvandeberg:
With new namespacing of plugins we need to be sure that check_reflect.py is looking in all the right places for correct reflection of objects.

mvandeberg:
Below is the output of check_reflect from Jenkins. Does this look correct?

```
ok:
steemit::app::candle_stick
steemit::app::discussion
steemit::app::order_history_item
steemit::app::tag_index
steemit::app::vote_state
steemit::chain::account_authority_object
steemit::chain::account_history_object
steemit::chain::account_object
steemit::chain::account_recovery_request_object
steemit::chain::block_summary_object
steemit::chain::change_recovery_account_request_object
steemit::chain::comment_content_object
steemit::chain::comment_object
steemit::chain::comment_vote_object
steemit::chain::convert_request_object
steemit::chain::db_schema
steemit::chain::decline_voting_rights_request_object
steemit::chain::dynamic_global_property_object
steemit::chain::escrow_object
steemit::chain::feed_history_object
steemit::chain::hardfork_property_object
steemit::chain::limit_order_object
steemit::chain::liquidity_reward_balance_object
steemit::chain::object_schema_repr
steemit::chain::operation_object
steemit::chain::operation_schema_repr
steemit::chain::owner_authority_history_object
steemit::chain::reward_fund_object
steemit::chain::savings_withdraw_object
steemit::chain::shared_authority
steemit::chain::smt_token_object
steemit::chain::transaction_object
steemit::chain::util::comment_reward_context
steemit::chain::vesting_delegation_expiration_object
steemit::chain::vesting_delegation_object
steemit::chain::withdraw_vesting_route_object
steemit::chain::witness_object
steemit::chain::witness_schedule_object
steemit::chain::witness_vote_object
steemit::plugins::account_by_key::get_key_references_args
steemit::plugins::account_by_key::get_key_references_return
steemit::plugins::account_by_key::key_lookup_object
steemit::plugins::account_history::api_operation_object
steemit::plugins::account_history::get_account_history_args
steemit::plugins::account_history::get_account_history_return
steemit::plugins::account_history::get_ops_in_block_args
steemit::plugins::account_history::get_ops_in_block_return
steemit::plugins::account_history::get_transaction_args
steemit::plugins::block_api::api_signed_block_object
steemit::plugins::block_api::get_block_args
steemit::plugins::block_api::get_block_header_args
steemit::plugins::block_api::get_block_header_return
[...]

PR #1540, opened 2017-09-11
Remove unused debug API types #1321

theoreticalbts:
PR for #1321


theoreticalbts:
Fixed typo in ticket name and commit text, it's removing types, not methods.

Issue #1529, opened 2017-09-07
Rename steemit namespace to steem


mvandeberg:
The steemit namespace should be changed to steem every where that it occurs.

mvandeberg:
All steemit build targets should be renamed to steem. (i.e. steemit_chain -> steem_chain)

vogel76:
Does this change shall also affect directories (ie `libraries/protocol/include/steemit`) ?

mvandeberg:
Yes

PR #1546, opened 2017-09-12
1529 steemit rename

vogel76:
Issue #1529 - all occurrences of 'steemit' inside source code (including CMakeLists.txt and doc files) replaced to 'steem'. Also few build scripts modified, to correctly handle reflection etc as referenced types are placed now inside steem namespace.


vogel76:
If possible please review and merge this PR soon, due to large number of changes actually in every file.

theoreticalbts:
I read that entire diff, everything looks good

Issue #1297, opened 2017-08-07
Old config options


mvandeberg:
All old config options should still work, but output deprecated warning messages.

mvandeberg:
For the time being, I have decided the chain plugin's config namespace will remain as is. The chain plugin exists as a small anomaly in its current iteration. This will be broken up better when each plugin has a memory mapped file and we begin breaking up consensus data into separate plugins.

PR #1547, opened 2017-09-12
Old config options

mvandeberg:
PR #1297


theoreticalbts:
I'm not really enthusiastic about the amount of code duplication in these commits.  But given that we'll probably simply remove it in the following release, it is probably not worth the effort to do this in a "cleaner" way.

mvandeberg:
I will create an issue to remove old config options in 0.20.X

Issue #1562, opened 2017-09-13
Run tests in read-only chroot


theoreticalbts:
Let's tighten down the CI environment to catch bugs of the form of #1561 (unit tests attempting to write to locations they shouldn't).

Some notes on this issue:

```
# See explanation at https://unix.stackexchange.com/questions/128336/why-doesnt-mount-respect-the-read-only-option-for-bind-mounts
mkdir -p steem-ro
mount -t none -o bind steem steem-ro
mount -t none -o bind,remount,ro steem steem-ro
chroot steem-ro tests/chain_test
```

Note, we will need to `ldd chain_test` and then copy the libraries needed by the `chain_test` binary into the chroot.  This process can be simplified by statically linking `steemd`.  The static build PR https://github.com/steemit/steem/pull/1446 needs to be ported to `develop`.

theoreticalbts:
`chroot` may be overkill, it could be simpler to just `chmod -R a-w build`

mvandeberg:
Closing, this is a part of #1561

Issue #1561, opened 2017-09-13
Unit tests throw error if witness_node_data_dir exists


theoreticalbts:
I'm getting the following line repeated many times on unit test output of `develop` branch, ending with:

```
Error parsing command line: the argument ('false') for option 'required-participation' is invalid
/home/theoretical/src/steem/libraries/appbase/include/appbase/application.hpp(78): fatal error in "Plugin& appbase::application::get_plugin() const [with Plugin = steem::plugins::chain::chain_plugin]": std::runtime_error: unable to find plugin: chain
```

The output ends with

```
*** 120 failures detected in test suite "Master Test Suite"
```

Obviously CI doesn't suffer from this problem.  Some more information:  I was getting this error running unit tests on my local machine, in a long-lived build directory that has been used to build many prior versions of Steem in-tree (with very frequent `make` commands and occasional `make clean` commands).

I noted a `witness_node_data_dir` exists in this directory, with a `config.ini` that contains the line `required-participation = false`.  Renaming this directory allows the tests to proceed without error, but the directory is re-created at startup.

- The CI environment should execute unit tests in a read-only chroot (with `/tmp` mounted read-write).  This should cause the code that is attempting to create `witness_node_data_dir` to fail.  I will file a separate ticket for this.
- The code that is attempting to create `witness_node_data_dir` should be fixed.

theoreticalbts:
An informative backtrace can be produced by these commands:

```
mkdir empty-ro-dir
chmod a-w empty-ro-dir
cd empty-ro-dir
gdb --args ../tests/chain_test -t '*/asset_test'
catch throw
run
bt
```

Here is the backtrace:

```
#1  0x000000000348cd40 in boost::filesystem::detail::create_directories(boost::filesystem::path const&, boost::system::error_code*) ()
#2  0x0000000002bf84a7 in boost::filesystem::create_directories (p=...) at /usr/include/boost/filesystem/operations.hpp:523
#3  0x0000000003451315 in appbase::application::write_default_config (this=0x40d5aa0, cfg_file=...) at libraries/appbase/application.cpp:189
#4  0x00000000034505ef in appbase::application::initialize_impl (this=0x40d5aa0, argc=1, argv=0x7fffffffe018, autostart_plugins=std::vector of length 3, capacity 3 = {...})
    at libraries/appbase/application.cpp:119
[...]

PR #1563, opened 2017-09-13
Make build dir read-only before running tests #1562

theoreticalbts:
PR for #1562.  This should not pass CI until a fix for #1561 is available, and should not be merged until it passes CI.

When such a fix is available, rebase this PR on the fix.  Then if CI passes for the rebased PR, the fix is valid, and both #1561 and #1562 can be closed once the rebased PR is merged to `develop`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,6 +53,7 @@ RUN \
         -DSKIP_BY_TX_ID=ON \
         .. && \
     make -j$(nproc) chain_test test_fixed_string plugin_test && \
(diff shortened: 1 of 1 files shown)


mvandeberg:
I thought this branch was not supposed to pass CI

theoreticalbts:
Good point.  I wonder why it does?

mvandeberg:
What is the status of this? It should fail CI right now and is not, but has not been touched in over 2 months.

mvandeberg:
Closing until we begin working on this again.

Issue #1517, opened 2017-09-06
Implement smt_set_runtime_parameters_operation


theoreticalbts:
- Implement the operation
- Implement unit tests
- Record parameters in `smt_token_object`
- Unit tests to check parameters have intended functionality

PR #1578, opened 2017-09-15
Issue #1517 - smt_set_runtime_parameters_operation

mariusztrela:
Issue #1517


theoreticalbts:
Here are my thoughts:

- It is clear that `curve_id` needs to be moved out of `chain`, since now there is an operation field with type `curve_id`.
- I don't think that `utilities` is the correct place to put `curve_id`, it is part of Steem-specific design, not "library" functionality.
- `curve_id` should be placed in `protocol` instead.
- We should not add `utilities` to `target_link_libraries` for `protocol`.
- Fixes to new code (especially small patches) do not need to be separate commits.  We encourage `git rebase -i` and `git push -f` to get rid of such commits in branches / PR's.

@mvandeberg What do you think?

mvandeberg:
I agree. As soon as a data type exists in an operation, its binary serialization is now a part of the blockchain protocol and it is a first class member of the protocol. `protocol` should be able to stand on its own without link libraries other than fc and boost.

libraries/chain/smt_evaluator.cpp
```diff
@@ -81,9 +81,57 @@ void smt_set_setup_parameters_evaluator::do_apply( const smt_set_setup_parameter
...
+      });
+   }
+};
+
 void smt_set_runtime_parameters_evaluator::do_apply( const smt_set_runtime_parameters_operation& o )
 {
    FC_ASSERT( _db.has_hardfork( STEEM_SMT_HARDFORK ), "SMT functionality not enabled until hardfork ${hf}", ("hf", STEEM_SMT_HARDFORK) );
+
+   const auto* _token = _db.find< smt_token_object, by_control_account >( o.control_account );
+   FC_ASSERT( _token, "Any token does not exist for ${account}",("account", o.control_account) );
```
goldibex: I'm not sure what "any token does not exist" means. i realize this isn't user facing, but I feel like error messages should be self-documenting.

libraries/protocol/smt_operations.cpp
```diff
@@ -192,7 +192,42 @@ void smt_cap_reveal_operation::validate()const {}
...
+      FC_ASSERT(  ( param_windows.cashout_window_seconds > ( param_windows.reverse_auction_window_seconds + SMT_UPVOTE_LOCKOUT ) ) &&
+                  ( param_windows.cashout_window_seconds < SMT_VESTING_WITHDRAW_INTERVAL_SECONDS ), "Cashout window second must be valid" );
+
+      FC_ASSERT( ( param_windows.reverse_auction_window_seconds + SMT_UPVOTE_LOCKOUT ) >= 0, "Reverse auction window seconds must be valid" );
+   }
+
+   void operator()( const smt_param_vote_regeneration_period_seconds_v1& vote_regeneration ) const
+   {
+      FC_ASSERT(  ( vote_regeneration.vote_regeneration_period_seconds > 0 ) &&
+                  ( vote_regeneration.vote_regeneration_period_seconds < SMT_VESTING_WITHDRAW_INTERVAL_SECONDS ), "Vote regeneration must be valid" );
```
goldibex: Again, I feel like error messages should be descriptive. this assertion's text is a repetition of the assertion itself.

libraries/protocol/smt_operations.cpp
```diff
@@ -192,7 +192,42 @@ void smt_cap_reveal_operation::validate()const {}
...
+   smt_set_runtime_parameters_operation_visitor(){}
+
+   typedef void result_type;
+
+   void operator()( const smt_param_windows_v1& param_windows ) const
+   {
+      FC_ASSERT(  ( param_windows.cashout_window_seconds > ( param_windows.reverse_auction_window_seconds + SMT_UPVOTE_LOCKOUT ) ) &&
+                  ( param_windows.cashout_window_seconds < SMT_VESTING_WITHDRAW_INTERVAL_SECONDS ), "Cashout window second must be valid" );
+
+      FC_ASSERT( ( param_windows.reverse_auction_window_seconds + SMT_UPVOTE_LOCKOUT ) >= 0, "Reverse auction window seconds must be valid" );
```
goldibex: Again, I feel like error messages should be descriptive. this assertion's text is a repetition of the assertion itself.

libraries/protocol/smt_operations.cpp
```diff
@@ -192,7 +192,42 @@ void smt_cap_reveal_operation::validate()const {}
...
+struct smt_set_runtime_parameters_operation_visitor
+{
+   smt_set_runtime_parameters_operation_visitor(){}
+
+   typedef void result_type;
+
+   void operator()( const smt_param_windows_v1& param_windows ) const
+   {
+      FC_ASSERT(  ( param_windows.cashout_window_seconds > ( param_windows.reverse_auction_window_seconds + SMT_UPVOTE_LOCKOUT ) ) &&
+                  ( param_windows.cashout_window_seconds < SMT_VESTING_WITHDRAW_INTERVAL_SECONDS ), "Cashout window second must be valid" );
```
goldibex: Again, I feel like error messages should be descriptive. this assertion's text is a repetition of the assertion itself.

theoreticalbts:
In addition to the comments above, merge conflicts now exist and need to be fixed.

Issue #1597, opened 2017-09-21
Account History Options Incorrect


mvandeberg:
The legacy options `history-whitelist-ops` and `history-blacklist-ops` are incorrectly defined as `whitelist-ops` and `blacklist-ops` in the account history plugin. This needs to be updated.

PR #1598, opened 2017-09-21
Account History Options Incorrect

mvandeberg:
PR for #1597


theoreticalbts:
Should also remove text `Deprecated in favor of...`

Issue #1607, opened 2017-09-29
`get_typename` errors with clang 4


mvandeberg:
Problems with `get_typename` were revealed when linting in earlier versions of clang. With clang 4, they are now errors. Luckily, this was recently fixed in fc and just needs the submodule just needs to be updated.

PR #1608, opened 2017-09-29
`get_typename` errors with clang 4

mvandeberg:
PR for #1607


theoreticalbts:
Hard to evaluate, but it passes CI, looks reasonable, and the old code was clearly broken.

Issue #1568, opened 2017-09-14
JSON RPC Tests


mvandeberg:
The json rpc plugin should be tested for conformance with the json rpc 2.0 spec.

http://www.jsonrpc.org/specification

All "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", and "RECOMMENDED" directives of the spec should fail the unit test if the requirements are not met.

All  "MAY" and "OPTIONAL" directives should be tested, but not fail the test. Print a warning if such a test fails.

vogel76:
At first please look for some tools able to do such verification..

PR #1621, opened 2017-10-04
1568 json rpc tests

mariusztrela:
Issue #1568


mvandeberg:
I realized that a couple of the examples I had given to you of API calls that should work were wrong. This led you to add an incorrect parsing case in the jsonrpc plugin. This was not your fault, but mine. I wanted to test a couple of cases for the condenser API to ensure that it is working as expected with the jsonrpc plugin. This all looks good now. Thanks for doing all of the heavy lifting and making adding test cases easy!

@theoreticalbts Because I am involved in development of this branch now, please review.

theoreticalbts:
The bool variable and the first `FC_ASSERT` condition in `process_params` are confusing.  They look like this:

```
bool call_mode = method == "call";

FC_ASSERT( ( call_mode && request.contains( "params" ) ) || !call_mode );

if( call_mode ) { ... }
else { ... }
```

But should be replaced with equivalent code that is easier to understand:

```
if( method == "call" )
{
   FC_ASSERT( request.contains( "params" ) );
   ...
}
else { ... }
```

mvandeberg:
```
if( method == "call" )
{
   FC_ASSERT( request.contains( "params" );
   ...
}
else { ... }
```

@mariusztrela Can you make the changes?

theoreticalbts:
Additionally the `rpc_id()` function looks too long and convoluted, some of the variant unboxing and error handling is unnecessary.

I think there is also at least one strange bug, that should probably be covered in the unit tests:  It looks like the code will unbox string values that are parseable as integers, i.e. a request with `{"id" : "1234"}` would return a response with `{"id" : 1234}` which seems wrong.

This function should use a more direct, simple approach, like this:

```
auto it = request.find("id");
if( it != request.end() )
{
   const fc::variant& request_id = it->value();
   switch( request_id.get_type() )
   {
      case int64_type:
      case uint64_type:
      case string_type:
         response.id = request_id;
         break;
      default:
         response.error = json_rpc_error( JSON_RPC_INVALID_REQUEST, "Only integer value or string is 
allowed for member \"id\"" );
   }
}
```

The assignment `response.id = request_id` simply copies the `variant` instance, no conversion to integer or string needs to be performed by the caller.

mvandeberg:
When I was developing the plugin the id parsing worked properly and returned and integer string as a string.

However, we should have a unit test for it to ensure that a refactor did not change those semantics accidentally.

I thought I had tried an implementation similar to your suggestion using `get_type` and for some reason it did not work. However, I was working on a lot of code at the time. It is likely I am not remembering some detail correctly.

theoreticalbts:
I'm also not sure of error handling when you call `(*call)(func_args);`, it seems the `catch( fc::assert_exception& e )` may catch assertions from either inside or outside the call.  You should have different `catch` blocks for `process_params()`, and `(*call)(func_args)`.

theoreticalbts:
Finally, `const auto request = message.get_object()` is unnecessary object copy, please use a const reference instead.

mariusztrela:
Done

libraries/plugins/json_rpc/json_rpc_plugin.cpp
```diff
@@ -154,11 +144,26 @@ namespace detail
...
+                  }
+
+                  try
+                  {
+                     if( call )
+                        response.result = (*call)( func_args );
+                  }
+                  catch( fc::assert_exception& e )
+                  {
+                     response.error = json_rpc_error( JSON_RPC_INVALID_CALL, e.to_string(), fc::variant( *(e.dynamic_copy_exception()) ) );
```
mvandeberg: Not sure invalid call is the right name here. This is an error during call.

tests/plugin_tests/json_rpc.cpp
```diff
@@ -87,16 +87,16 @@ BOOST_AUTO_TEST_CASE( basic_validation )
       make_request( request, -32600 );
 
       request = "{\"jsonrpc\": \"2.0\", \"method\": \"xyz\", \"params\": [], \"id\": 1}";
-      make_request( request, -32601 );
+      make_request( request, -32002 );
```
mvandeberg: To prevent this type of change in the future, we should move the error code defs to the jsonrpc plugin header.

mariusztrela:
Done

PR #1623, opened 2017-10-05
SMT tests refactoring.

mariusztrela:
New class 'smt_database_fixture'.


mvandeberg:
I like this. It will allow us to make changes to the database_fixture without creating SMT merge conflicts.

Before release, any SMT specific additions should be reintegrated with clean_database_fixture.

Issue #1590, opened 2017-09-18
Implement benchmarking system to measure steemd reindex performance


vogel76:
Before any serious refactorings will start, definitely benchmarking system must be created, offering performance measure for each done build.
Prequisistes:

1.  steemd must offer an option to stop reindex at given block number, to make possible limiting execution time during benchmark
1. introduce functionality which allows gathering statistic data related to
- number of processed blocks per `s` during reindexing. 
- number of created state objects (i.e. account_objects etc),
- allocated memory
Such data collection shall be done in intervals (ie every 1k processed blocks), to show potential time/memory allocation change
Maybe existing `blockchain_statistic_api` could be enhanced and reused for this purpose.

vogel76:
Such benchmarking process shall operate on already downloaded blockchain, to avoid time violation caused by network traffic.

mvandeberg:
We should be careful using the `blockchain_statistic_plugin` in this manner because it creates objects and has a non-trivial runtime. It would be easy enough to have some counters in chainbase that count creations/modifications/deletions for each object type.

vogel76:
Please add options:
1. stop-replay-at-block <block-no> which shall lead steemd to stop and exit after reaching given block number.
1. collect-benchmark-results <out-file> which shall print benchmark results on the stdout, together with messages related to replay progress. Additionally, XML/JSON file shall be produced containing:
  - benchmark data header containing: operation name, operation start time
  - potentially multiple data collections containing range of blocks the data have been gathered for. Collected data shall contain CPU time, Real time values, Current mem. allocation, Peak mem. allocation
  - one final data collection containing total time measure for whole operation incl. memory report as described above

Gandalf-the-Grey:
blocks per second value would change over time during the reindexing
getting average for every 100k blocks processed might be a good idea

Gandalf-the-Grey:
oh, also transactions per second (along with blocks per second) would be a valuable info

vogel76:
Implementation done, memory report has been verified and corrected as specified here:
https://github.com/steemit/steem/issues/1651
@sneak  Please move it to any person being responsible for adding a benchmark scenario (at replay) to CI and do collection of results for further analysis.

mvandeberg:
The next step in this issue is adding some automated benchmarking.

Removing myself and adding to Backlog

mvandeberg:
For now, we won't add the benchmarking to CI. Closing.

PR #1624, opened 2017-10-05
New options "stop-replay-at-block" & "collect-benchmark-results" ready to merge.

Kiwonik:
Enhanced chain plugin with "stop-replay-at-block" & "collect-benchmark-results" command line options #1590

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -143,7 +143,8 @@ void database::open( const fc::path& data_dir, const fc::path& shared_mem_dir, u
    FC_CAPTURE_LOG_AND_RETHROW( (data_dir)(shared_mem_dir)(shared_file_size) )
 }
 
-void database::reindex( const fc::path& data_dir, const fc::path& shared_mem_dir, uint64_t shared_file_size )
+void database::reindex( const fc::path& data_dir, const fc::path& shared_mem_dir, uint32_t stop_replay_at, 
+                        uint64_t shared_file_size, uint32_t* last_block_number /*= nullptr*/ )
 {
    try
    {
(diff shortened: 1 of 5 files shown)


libraries/plugins/chain/chain_plugin.cpp
```diff
@@ -54,6 +57,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
    cli.add_options()
          ("replay-blockchain", bpo::bool_switch()->default_value(false), "clear chain database and replay all blocks" )
          ("resync-blockchain", bpo::bool_switch()->default_value(false), "clear chain database and block log" )
+         ("stop-replay-at-block", bpo::value<uint32_t>(), "Stop and exit after reaching given block number")
+         ("collect-benchmark-results", bpo::value<uint32_t>(), "Print time and memory usage every given number of blocks")
```
mvandeberg: This option name did not make sense to me until I read the description. Perhaps `benchmark-interval` would be a better name?

mvandeberg: Just saw that's what you called the variable. Yeah, let's use this name.

libraries/chain/database.cpp
```diff
@@ -143,7 +143,8 @@ void database::open( const fc::path& data_dir, const fc::path& shared_mem_dir, u
    FC_CAPTURE_LOG_AND_RETHROW( (data_dir)(shared_mem_dir)(shared_file_size) )
 }
 
-void database::reindex( const fc::path& data_dir, const fc::path& shared_mem_dir, uint64_t shared_file_size )
+void database::reindex( const fc::path& data_dir, const fc::path& shared_mem_dir, uint32_t stop_replay_at, TBenchmark benchmark,
+                        uint64_t shared_file_size, uint32_t* last_block_number /*= nullptr*/ )
```
mvandeberg: Let's make `last_block_number` the return type of `reindex`. We try to avoid passing pointers wherever possible.

It is a minor pet peeve, but let's put the new args at the end and default initialize them so that the previous method signature is still valid.

Kiwonik: Committed and pushed all requested changes. Ready to merge. #1590

mvandeberg:
Build was failing on OS X. Pushed fix. It looks like FC_REFLECT does not like `int`, and `long`. As a practice, we should avoid using those types wherever possible and use explicitly sized types.

PR #1631, opened 2017-10-09
Chain id command line arg

mariusztrela:
Chain ID can be set from command line.


mvandeberg:
Implement this in the chain plugin. There is no reason for appbase to know anything about Steem.

Also add the changes to the testnet branch because that is the feature branch that these changes are for. If you want to make the changes incremental and then merge to that branch via an approved PR, I would be happy to review to code.

libraries/protocol/include/steem/protocol/types.hpp
```diff
@@ -148,6 +148,11 @@ namespace steem {
          friend bool operator == ( const extended_private_key_type& p1, const extended_private_key_type& p2);
          friend bool operator != ( const extended_private_key_type& p1, const extended_private_key_type& p2);
       };
+
+      extern std::string chain_id_name;
```
mvandeberg: Can this not be a global variable?

I think everywhere that needs the chain id probably has access to the database. Make it a settable parameter in the database. When the chain_plugin opens the database, set it to STEEM_CHAIN_ID if it is not set via CLI, or the passed arg.

The one place that this might not work well is the p2p code. If needed, add a `get_chain_id()` virtual method to the node delegate and implement it in the p2p plugin.

mariusztrela: Question 1:
Where should I get info about `STEEM_CHAIN_ID`, where I use wallet ? It is necessarily in `sign_transaction` function in `wallet.cpp`.
My idea is, to add additional method to `condenser_api` for example: `get_chain_id_name` and calling `_remote_api->get_chain_id_name()`, when we are first time in `sign_transaction`.

Question 2:
In file `sign_transaction.cpp` in `main` function we havent't access to `database` object. I think, I should create `fc::sha256` object independently from `database` object.

mvandeberg: 1. Add a command line arg to the wallet as well.

2. Keep the default STEEM_CHAIN_ID.

One purpose of the chain ID is to segment transactions between chains via signature. It would be bad if the cli wallet automatically worked if you connected to GOLOS. Right now you need to recompile for it to work. This is to add the ability for the user to configure the same binary to work on compatible forks. The sign_transaction utility is not widely used to the best of my knowledge, so I think it is safe keeping it working with just the default chain id. This is only going to be used for test nets, not for the main net. The chain id for the main net is set and will never change.

mariusztrela:
Done

libraries/appbase/include/appbase/application.hpp
```diff
@@ -88,6 +88,9 @@ namespace appbase {
 
          boost::asio::io_service& get_io_service() { return *io_serv; }
 
+         const std::string& get_chain_desc() const{ return chain_desc; };
```
mvandeberg: No, nothing about the blockchain should be in appbase. Appbase is application agnostic.

libraries/protocol/include/steem/protocol/config.hpp
```diff
@@ -33,12 +32,14 @@
 #define STEEM_ACCOUNT_RECOVERY_REQUEST_EXPIRATION_PERIOD  fc::seconds(12)
 #define STEEM_OWNER_UPDATE_LIMIT                          fc::seconds(0)
 #define STEEM_OWNER_AUTH_HISTORY_TRACKING_START_BLOCK_NUM 1
+
+#define STEEM_CHAIN_ID_NAME "testnet"
```
mvandeberg: Don't worry about the seed for the chain id. By definition, the chain ID is a sha256 hash. The cli option should just be a hash.

programs/steemd/main.cpp
```diff
@@ -39,30 +39,33 @@ string& version_string()
...
+      std::cerr << "chain id: " << appbase::app().get_chain_desc() << "\n";
       std::cerr << "blockchain version: " << fc::string( STEEM_BLOCKCHAIN_VERSION ) << "\n";
       std::cerr << "------------------------------------------------------\n";
 #else
       std::cerr << "------------------------------------------------------\n\n";
       std::cerr << "            STARTING STEEM NETWORK\n\n";
       std::cerr << "------------------------------------------------------\n";
       std::cerr << "initminer public key: " << STEEM_INIT_PUBLIC_KEY_STR << "\n";
-      std::cerr << "chain id: " << std::string( STEEM_CHAIN_ID ) << "\n";
+      std::cerr << "chain id: " << appbase::app().get_chain_desc() << "\n";
```
mvandeberg: Display STEEM_CHAIN_ID for main net.

programs/steemd/main.cpp
```diff
@@ -39,30 +39,33 @@ string& version_string()
...
-   {
 #ifdef IS_TEST_NET
       std::cerr << "------------------------------------------------------\n\n";
       std::cerr << "            STARTING TEST NETWORK\n\n";
       std::cerr << "------------------------------------------------------\n";
       auto initminer_private_key = steem::utilities::key_to_wif( STEEM_INIT_PRIVATE_KEY );
       std::cerr << "initminer public key: " << STEEM_INIT_PUBLIC_KEY_STR << "\n";
       std::cerr << "initminer private key: " << initminer_private_key << "\n";
-      std::cerr << "chain id: " << std::string( STEEM_CHAIN_ID ) << "\n";
+      std::cerr << "chain id: " << appbase::app().get_chain_desc() << "\n";
```
mvandeberg: Don't display chain id for test net if if means making appbase chain id aware.

mvandeberg: Besides, at this point, nothing would be set in appbase anyway.

programs/util/sign_transaction.cpp
```diff
@@ -49,7 +51,7 @@ int main(int argc, char** argv, char** envp)
       tx_signing_result sres;
       sres.tx = sreq.tx;
       sres.digest = sreq.tx.digest();
-      sres.sig_digest = sreq.tx.sig_digest(STEEM_CHAIN_ID);
+      sres.sig_digest = sreq.tx.sig_digest( _chain_id );
```
mvandeberg: When you revert the chain id name change, this file should be unchanged.

mariusztrela: Done

Issue #1640, opened 2017-10-11
total_vesting_shares > STEEM_MAX_SHARE_SUPPLY


mvandeberg:
This first happened around block 900k and nothing has broke. I don't believe this is a threat to the blockchain, but an oddity. If I recall correctly, STEEM_MAX_SHARE_SUPPLY was not intended as an upper bound on asset magnitude, but only on the STEEM supply. However, there are a couple of places in the code that use it incorrectly and violate assumptions.

```
libraries/plugins/apis/tags_api/tags_api.cpp:   auto tidx_itr = tidx.lower_bound( boost::make_tuple( tag, parent, share_type( STEEM_MAX_SHARE_SUPPLY ) )  );
```

With the current voting semantics, the max rshares a post can have is 1/50th of the number of VESTS, which is 7822080302916958.32 and is greater than the max share supply by ~8x. So if 1/8th of VESTS vote on a single post, it may not be returned properly by the tags plugin.

```
libraries/protocol/asset.cpp:      price price::max( asset_symbol_type base, asset_symbol_type quote ) { return asset( share_type(STEEM_MAX_SHARE_SUPPLY), base ) / asset( share_type(1), quote); }
libraries/protocol/asset.cpp:      price price::min( asset_symbol_type base, asset_symbol_type quote ) { return asset( 1, base ) / asset( STEEM_MAX_SHARE_SUPPLY, quote); }
```

This is not an problem because we do not have an extreme vesting share price and we do not index objects by vesting share price. It is highly unlikely for such a price to exist, but because there is not an upper bound on VESTS, it is theoretically possible and thus violates the constraint of max and min price.

This is an immediate issue only because some of the asset changes for SMTs have assumed that STEEM_MAX_SHARE_SUPPLY is an upper bound for the magnitude of an asset. This assumption is incorrect and resulted in the inability to serialize the total vesting shares to a String.

The purpose of this issue is to fix the NAI asset String serialization and determine if the other locations where this assumption is violated need to be refactored as well.

mvandeberg:
This particular issue only affects string serialization. The logic for the asset amount is the same with NAI support and without. Restoring the previous serialization logic but utilizing the `asset_symbol_type` serialization logic should fix the problem and retain serialization semantics.

theoreticalbts:
It is clear that we need a separate constant for the maximum number of satoshis of an asset allowed.  I will set this to `2^63-1` for now, we may come up with a tighter bound later.

[...]

PR #1641, opened 2017-10-11
Fix asset string serialization

mvandeberg:
PR for #1640


theoreticalbts:
PR #1647 is intended to replace this PR.  If #1647 PR is merged, then this PR should be closed and not merged.

Issue #1625, opened 2017-10-05
Implement supply functions for SMT's


theoreticalbts:
We need to be able to adjust SMT balances and track SMT's in supply accounting.  EDIT:  This work is complex and involves several moving parts and subtle issues.  So let's start things off with a big-picture discussion of what parts of Steem need to change to support arbitrary asset balances, and file separate issues for specific tasks as necessary.

theoreticalbts:
Let's start with a historical review.  Originally Steem was not designed to have multiple asset types, so an account's STEEM, SBD and VESTS balances exist as specific fields on the account object.  Likewise, the savings balances exist as specific fields.

So I started to write down questions, it seems clear that maybe each of these questions should be its own issue.  So I will put checkboxes here, to be checked as each issue is closed.  I will also create issues for each item, and edit this post to include issue numbers of each issue.

- [ ] It is clear that SMT balances must be implemented as a separate object indexed by `account_name`, `smt_symbol` fields.
- [ ] Should STEEM, SBD and VESTS balances continue to be fields on the account object, or should they be refactored to be separate objects as well?
- [ ] Currently SBD interest processes whenever an SBD balance changes.  Should this behavior continue?
- [ ] Should we refactor SBD balance into a struct like this?  It would allow unification of handling of SBD balance inside / outside savings.

```
struct account_sbd_balance
   asset             sbd_balance;
   uint128_t         sbd_seconds;
   time_point_sec    sbd_seconds_last_update;
   time_point_sec    sbd_last_interest_payment;
};
```

- [ ] Is the `price` structure and market handling system able to handle multiple different kinds of orders?  I think the answer is "yes" since much of the code is borrowed from BitShares, which can.  But I'm not sure, and it certainly needs tested.
- [ ] Should we move overflow checks to balance adjustment functions?  It seems like this is a no-brainer refactoring to make it harder to create overflow bugs.
- [ ] I'm thinking about writing a compile-time bounded integer type which automatically bounds checks the result of any arithmetic operation.  Should we do this?
- [ ] What is the global upper bound on the number of satoshis of an asset that can exist?
[...]

PR #1647, opened 2017-10-12
Add STEEM_MAX_SATOSHIS constant #1640 #1625

### LINKED ISSUE #1640: total_vesting_shares > STEEM_MAX_SHARE_SUPPLY

theoreticalbts:
PR which should fix #1640.  Also related to #1625.  If this PR is merged, then PR #1641 should be closed and not merged.


mvandeberg:
I would still like #1641 to be considered. The code is much easier to read and does not rely on tedious buffer offsets.

theoreticalbts:
> The code is much easier to read and does not rely on tedious buffer offsets.

The code in  #1641 does not rigorously define acceptable input, and is relatively loose in what it will accept, including some unusual or undefined behavior.  For example if one of the substrings is long enough that the value overflows integer type, or if you have string such as `"100.12 34"`, or `"-123.-45"`, or code with multiple spaces between the amount and the symbol.  I think code that processes strings character-by-character is clearer, especially in terms of its behavior on exceptional input.  And since this method will get un-sanitized strings supplied directly from the public Internet, we should be very careful about our handling of exceptional inputs.

This PR's code is certainly faster, since it avoids the creation of temporary `string` and `stringstream` objects.

Issue #1661, opened 2017-10-16
Rebroadcast Mem Pool


mvandeberg:
To ensure that a local transaction in the mem pool does not get stalled, we want to implement rebroadcasting. Every 3 blocks, each node will rebroadcast all known pending transactions. For nodes that aren't having connection issues, this would consist of rebroadcasting their inventory as though it were new and peers would opt not to request said transactions. For nodes struggling with connections, this may appear as new transaction and they would request it.

theoreticalbts:
Let's not close this ticket until we've merged PR's for both `master` and `develop`

mvandeberg:
This one is going to be a bit trickier in develop. It is pretty hacked together in master because of how much we use globals.

mvandeberg:
This has not been done for appbase.

PR #1662, opened 2017-10-16
Rebroadcast Mem Pool

mvandeberg:
PR for #1661

### DIFF
--- a/libraries/app/application.cpp
+++ b/libraries/app/application.cpp
@@ -59,6 +59,9 @@
 
 #include <boost/range/adaptor/reversed.hpp>
 
+// Every 3-6 blocks, rebroadcast. Do this randomly to prevent simultaneous p2p spam.
+#define REBROADCAST_RAND_INTERVAL() 3 + ( rand() % 4 )
+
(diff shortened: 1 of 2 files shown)


libraries/app/application.cpp
```diff
@@ -351,6 +365,9 @@ namespace detail {
          }
          _chain_db->show_free_memory( true );
 
+         _next_rebroadcast = _chain_db->head_block_num() + 3 + rand() % 4;
```
goldibex: I'd keep this code identical with line 243. It's a candidate for refactoring after a couple more uses and keeping its expression consistent should help other developers recognize that later.

libraries/app/application.cpp
```diff
@@ -351,6 +365,9 @@ namespace detail {
...
+         _rebroadcast_con = _chain_db->applied_block.connect( [&]( const signed_block& b ){ rebroadcast_pending_tx(); } );
```
goldibex: Forgive my total ignorance of C++, but it is it necessary for the lambda to have access to all the variables in the enclosing scope?

Issue #1660, opened 2017-10-16
Increase Active Ignored Request Timeout


mvandeberg:
When a node requests a transaction or block, there is a 1 second timeout before disconnecting and asking another peer. We know there are circumstances in which full API nodes can't respond within one second, causing all peers to disconnect from them.

We want to increase this value to 6 to give ample time for the p2p code to respond to the request.

theoreticalbts:
Let's not close this ticket until we've merged PR's for both `master` and `develop`

PR #1663, opened 2017-10-16
Increase Active Ignored Request Timeout

mvandeberg:
PR for #1660


libraries/net/node.cpp
```diff
@@ -1370,7 +1370,9 @@ namespace graphene { namespace net { namespace detail {
         // 1 second seems reasonable.  When we get closer to our eventual target of 1 second
```
goldibex: I'd probably change this comment as well in light of the change to 6 seconds

Issue #1655, opened 2017-10-13
Block size can eclipse p2p max message size.


mvandeberg:
In protocol https://github.com/steemit/steem/blob/master/libraries/protocol/include/steemit/protocol/config.hpp#L228

The max block size is set to 375 MiB.

In the p2p code https://github.com/steemit/steem/blob/master/libraries/net/include/graphene/net/config.hpp#L39

The max message size is 2 MiB.

It is possible for a block that is accepted via consensus to be rejected via the p2p code.

The max block size is never actually checked against an upper bound when updating witness properties

https://github.com/steemit/steem/blob/master/libraries/protocol/include/steemit/protocol/steem_operations.hpp#L394-L400
https://github.com/steemit/steem/blob/master/libraries/chain/steem_evaluator.cpp#L64-L108

Furthermore, the block size is only checked against `STEEM_MAX_BLOCK_SIZE` if you are producing.
https://github.com/steemit/steem/blob/master/libraries/chain/database.cpp#L828

Is 375 MiB a reasonable max block size?
Should we increase the p2p max message size to the max block size?

What needs to change ASAP.

- [ ] Increase p2p max message size to 375 MiB.
- [ ] Soft fork to ensure witness votes max block size is less than 32 MiB. 
- [ ] When generating a block, check block size against witness voted block size.

In a hard fork we need to enforce the max block size check in consensus.

theoreticalbts:
> the block size is only checked against STEEM_MAX_BLOCK_SIZE if you are producing.

This is a problem and needs to be fixed.

> needs to change ASAP

I don't know that we have to rush an emergency hardfork out for this, see my reasoning below.

> Should we increase the p2p max message size to the max block size?

Given that the current p2p protocol transmits an entire block in a single message, the p2p message size should be greater than or equal to the max block size, plus whatever wrapping goes around the block to form the p2p message.

Instead of increasing the p2p max message size to the max block size, let's decrease the max block size to the max p2p message size, because the p2p message size is a reasonably sized value, and the max block size is unreasonably large.

> Is 375 MiB a reasonable max block size?

Certainly not.

[...]

PR #1664, opened 2017-10-17
Max Block Size Soft Fork

mvandeberg:
PR for #1655


theoreticalbts:
This patch is a step in the right direction, but incomplete.

- We should also check `dgpo.maximum_block_size` against the size of incoming blocks, currently we only check against generated blocks.
- We should also include code to make these hard (not soft) requirements as of HF20.
- We should also include `#warning` stating that we should, if possible, make the restrictions retroactive once HF20 has occurred.

mvandeberg:
> We should also check dgpo.maximum_block_size against the size of incoming blocks, currently we only check against generated blocks.

This is a hardfork then.

> We should also include code to make these hard (not soft) requirements as of HF20.

I will make the issue part of 0.20.0, but I'd rather not ship future hard fork code in this release. We have not done so in the past and I don't particularly want to start now.

> We should also include `#warning` stating that we should, if possible, make the restrictions retroactive once HF20 has occurred.

I can add a TODO `#pragma message`

PR #1665, opened 2017-10-17
Add rate limit zone to nginx steemd proxy

jnordberg:
Set the limit to 100 requests per with a +50 burst. Could probably be lowered.


jnordberg:
That's correct, closing this.

Issue #1567, opened 2017-09-14
API Testing


mvandeberg:
We have no testing of our APIs. At the very least, we could use sanity checks that the API stack works properly and returns sane results.

vogel76:
Maybe it is worth to create dedicated API (or use debug_plugin_api) to implement a helper API offering functionality to iterate over all registered APIs in the application.
Thus, there will be easily controlled:
- possibility to call any API method
- availability of all needed APIs being registered in the application.
In further steps, dedicated tests shall be done for specific APIs, what could be not so straightforward, since some of them can need proper data imported/generated from blockchain.
For such 2nd step, dedicated testing schedule must be defined, especially which APIs shall be tested first.

mvandeberg:
One idea that @theoreticalbts and I had several months ago was to do all of the api testing in python. @theoreticalbts had been working on an approach to reproducing a base blockchain state using the debug node API interfacing through Python.

Whatever you want to do to test the APIs, I am fine with. Why don't you create some infrastructure and write a simple test for `database_api.get_dynamical_global_properties`. We can review the approach with a small example and if it looks good we can move forward with that approach for the rest of the APIs.

vogel76:
Probably this python library can be helpful to make tool executing tests.
https://github.com/svanoort/pyresttest

mkochanowicz:
I added some unit-tests using simple PyRestTest tool (https://github.com/svanoort/pyresttest).

- Tests can be defined in YAML/JSON config files.
- Request body could be templated (Python template string) or even loaded from file.
- Validators could extract tested parts of response body.
- Logic is simply extendable (extractors, comparators, validators).
- It works on Linux & Mac.
- Looks like easy adoptable as part of smoketest after build with Jenkins.

I added also simple extension: new validator comparing extracted result with pattern file. If comparison failed validator generates output file to decide what to do (create new pattern or change comparison method).

There are several tests for `database_api` - most of methods without args.

Tests are in `python_scripts/tests/api_tests`. To run them:
- Install PyRestTest.
- Launch steemd.
- Set `RPC_PORT` in `run_api_tests.sh` the same as in steemd's `config.ini`.
- Execute `run_api_tests.sh` in `python_scripts/tests/api_tests` dir.

Note: existing patterns (*.json.pat) are generated from steemd launched in TESTNET with `run_api_tests.sh` set to false.

mkochanowicz:
Now single test in form:
```
- test:
  - name: "list_witnesses"
[...]

PR #1666, opened 2017-10-17
add simple API test using pyresttest tool #1567

mkochanowicz:


vogel76:
1) If you have some thoughts related to task(s) please include them at task specification, to make clear expectations.  
2) Proposed framework allows filtering/formatting parts of output before comparison. Thus it is always possible to include in pattern only part of output, being critical for verification. Because of above not any change in API implementation (specifically *given method*) will lead test to fail, but only changes having significant impact on produced results. The nature of regression tests is need to maintain their set what is of course always a time cost. On the other side we have hard proof things works as expected. 
Additionally specifying a pattern in a file has only advantages - allows to write some automaton (which actually in some way this testing framework is) to produce the patterns (ie test generator based on request/results sent by condenser to backend ?). Embedding expected result in a code has only disadvantages - it looks cool for very simple testcases, where for complex ones leads to embedding output in textual form in the code what is hard to maintain. It also increases time needed to create test and also requires knowledge of programming language for person creating the tests. Proposed form needs just simple test editing skills.
3) That is true, at current state test description is verbose but it can be easily improved by using YAML anchors (feature similar to inheritance), what leads to specifying common parts of structure defaults in one place and easily reuse. Proposed test suite description will be enhanced by this feature to show more compacted form.

python_scripts/tests/api_tests/database_api/get_config.json.pat
```diff
@@ -0,0 +1,158 @@
+{
```
mvandeberg: Is there any way to programmatically generate this file? A lot of this does not change, but if anything does, it is tedious to make the change in N different places where the data is duplicated.

mvandeberg:
I am not against this testing framework, but I do have a couple of questions about this approach.

Because results are in json, can we convert these to python objects and compare the values of the fields? Perhaps this is what the suite does already. That would get around some of the problems that @theoreticalbts voiced. Adding fields or changing the order of fields would not break the unit test if that were the case.

python_scripts/tests/api_tests/database_api/get_config.json.pat
vogel76: Pattern file can be easily produced by renaming the output file which testing framework generates. Of course it needs to be verified and accepted by test maintainer.

vogel76:
@mvandeberg Comparison is done at JSON object level, using custom-defined filtering/comparing function specified at Python level. Produced textual .out and .pat files are just additional effects of such comparison (.out) and lead to significantly simpler analysis of failed/new test result - it is much more faster and easier to compare (and see) text files instead of debugging code to see differences at object level.
Moreover, having patterns in textual files, allows to massively modify/process patterns (ie by some dedicated script) when after some code modifications some changes are expected in patterns.

vogel76:
Just to link PR & issue together: This PR is associated to issue #1567
Also opinion please.

python_scripts/tests/api_tests/database_api/request_template.json
```diff
@@ -0,0 +1,6 @@
+{
+    "jsonrpc": "2.0",
+    "id": $test_id,
+    "method": "call",
```
mvandeberg: Change this to the new call format assuming the replacement works.

```
"method":"$api.$method"
"params":$args
```

mvandeberg:
My leaning is that if the comparison is done at JSON or Python object level, then this is checking the actual result of the API call rather than the format and @theoreticalbts's concerns regarding copy pasting output to pass tests is lessened.

To prove this, what happens if you do not include all of the fields on the `get_dynamic_global_properites` test? Does it still pass? I think it needs to so that we can extend APIs without breaking the tests. Obviously, we should go back and update them, and we will, but this will lead to less of the copy pasting.

mvandeberg:
This is a good start, but I really want to run these tests on a node that has some activity so that we can fully test the rest of the calls. Many return no data until there is some activity on the chain. This should probably use steem-python to push transactions.

mkochanowicz:
> To prove this, what happens if you do not include all of the fields on the get_dynamic_global_properites test? Does it still pass?

It depends on comparator. Currently used comparator requires response is **equal** to pattern. To pass the case you described, we need comparator which requires response **contain** pattern - it is not a problem to add new comparator. The question is: pattern should not be updated after new fields added?

mvandeberg:
It should, but should not necessarily result in a failure.

Our philosophy behind using struct args and return types is that we can add fields without changing existing behavior.

If we decide to change an API call in such a way that it can return more fields when certain arguments are present we should have a test for the new feature and ensure that the current behavior remains the same with the existing test.

python_scripts/tests/api_tests/database_api/request_template.json
mkochanowicz: done

mkochanowicz:
OK, so I see two passes of tests:

- check current behavior remains the same (new fields should not cause failure)

- check response is **equal** to pattern (and generate json.out if not equal to helps produce new pattern)

The first should be run with Jenkins, the second could be run on dev machine to produce new patterns after fields added. I don't know only how to force developers to change patterns in such scenario (to be sure they run the second after add a new field).

mvandeberg:
Assuming that swapping the order of fields in the output files results in a passing test, I am comfortable with this approach. One thing this test does not do that it needs to is a way to set state and test that a blockchain with some state returns the correct values from the APIs. This can be done initially by using the debug node to create a 100 or so height blockchain with various ops in it that can then be queried via API and tested with the plugin. This branch is getting close to what we need, but isn't there yet.

mvandeberg:
Is this correct? I see 400 files changed and 300,000 additions?

mkochanowicz:
Yep - it's correct. 2 files per each test + a few configuration, bash, py files.

mvandeberg:
How am I supposed to review 300,000 additions?

mkochanowicz:
Where do you see this number? I see 400 files. Additions - maybe lines? There are only new json, yaml, py, bash files - no any changes in `steem` code.

mvandeberg:
screen shot 2018-01-16 at 10 05 29

mvandeberg:
I am seeing many copies of the same test to get dynamic global properties, all calling the exact same method, all expecting the exact same result. What exactly is the point of that?

vogel76:
The main goal of this branch is creation of system being able to run regressions (nightly) or on demand. The most of tests which looks stupid, redundant and meaningless is just an effect of try to **generate** some basic testsuite, which can cover API functionality used by condenser calls - they were automatically generated by spawning steemd with enabled json-rpc dump and playing with steemit website pointed to condenser talking to such traced steemd. 
Most of such tests looks like a trash, by design they are not done as systematic testsuite which should cover corner cases for each API function etc. Their goal is to have fast check to see that steemd reports the same for same condenser request until some changes are expected. Then, such tests could be thrown away and replaced by new (similary generated) suite.
In opposite, when there is no expected change, any difference reported by this testsuite is a signal of just uncovered bug.

mvandeberg:
Can we at least go through and remove duplicate tests? They accomplish nothing and make the test suite appear to be more complete than it is.

mkochanowicz:
@mvandeberg You are probably referring to `condenser_api` tests. They are created by clicking on website, so that's why some of them are duplicated. But ok, I removed all of them.

mvandeberg:
I see a script that allows comparison between two nodes. I don't see documentation on how to test against defined expected results. I also do not see documentation on how to get pyresttest. Before that can be integrated into any form of automated test, this needs to be added. I'd like to have a discussion on our API testing philosophy. What should that look like, how do we guarantee consistency across time, etc? These are perhaps more ops related questions. For now, I am merging this. Will create a new issue to answer these questions.

Issue #1671, opened 2017-10-19
Trim Account History


mvandeberg:
To keep state smaller, we will only keep 30 days of account history or 30, whichever is more.

If node operators want to keep the current functionality, an option `--history-disable-pruning` will be added which will turn off this behavior.

KINGdotNET:
Yes please keep it slim and small for faster access.
Thank you.

steemypete:
What happens if requesting further, does just send 30 or uses a slower access method to grab and return it?

mvandeberg:
No, this limits the account history accessible via API to 30 days or 30, whichever is more. The API call to access history remains unchanged.

KINGdotNET:
@steemypete I was about to ask the same question. 
Thank you @mvandeberg for a quick answer

mvandeberg:
Pruning can be disabled if you want to run it on your node.

We also have plans to move account history available through our public APIs outside of steemd and will return full account history.

PR #1672, opened 2017-10-19
Prune Account History

mvandeberg:
PR for #1671

### DIFF
--- a/libraries/chain/include/steemit/chain/history_object.hpp
+++ b/libraries/chain/include/steemit/chain/history_object.hpp
@@ -91,13 +91,6 @@ namespace steemit { namespace chain {
                member< account_history_object, uint32_t, &account_history_object::sequence>
             >,
             composite_key_compare< std::less< account_name_type >, std::greater< uint32_t > >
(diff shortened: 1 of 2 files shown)


theoreticalbts:
Ideally we could do this with the existing index.  Can you do `itr = lower_bound( make_tuple( item, 0 ) )` to get a pointer to the last entry for the current account and iterate from there as necessary?

Pretty sure `sequence - seq_itr->sequence > 30` needs to be rewritten as `sequence > seq_itr->sequence + 30` due to potential integer overflow.

theoreticalbts:
- If account `a` has a history item with a `sequence` number of zero, `lower_bound(make_tuple(a, 0))` will return that element.  Since `seq_itr` is decremented immediately, such an element can never be deleted by this code.

- `seq_itr` needs to be checked against `seq_idx.begin()` on each loop iteration.

theoreticalbts:
As we discussed, it's not possible for the iterator to go off the end, and also it must be the case that `sequence >= seq_itr->sequence`.

Issue #1676, opened 2017-10-20
Internal Reference to SMTs should be asset_symbol_type


mvandeberg:
Several places currently have the NAI as the internal reference to an SMT. The NAI is intended as an external only String representation. The `asset_symbol_type` as is, is the correct internal identifier to use.

Issue #1653, opened 2017-10-13
Define and implement testing scenarios related to smt_create_operation


vogel76:
Testing scenarios mostly depend on issue and related questions in the comment inside :
https://github.com/steemit/steem/issues/1554

vogel76:
BLOCKED by  issue https://github.com/steemit/steem/issues/1554

Kiwonik:
1. (Positive) Create multiple SMT with the same control account.
   * Use same or different precision (including 0)
2. (Negative) Create SMT with precision too big.
3. (Negative) Create conflicting tokens with
  a) Same NAI and precision.
  b) Same NAI but different precision.
3. (Negative) Hijacked NAI - used to create SMT by different users with
  a) Same precision
  b) Different precision
4. (Positive) Re-launch - after failed ICO the creator tries to reuse previously acquired NAI
   * This scenario is allowed currently by whitepaper "The symbol will remain reserved to the specified control_account. However, in order to launch the token, an smt_create_operation must be issued and the smt_creation_fee must be paid again."
   * Implementing requires that `smt_create_operation` evaluation recognizes the re-launch and reuses existing smt_token_object.
   * What to do with not refunded contributions left over from failed launch?

Kiwonik:
Implemented cases 1 to 5. Number 6 (SMT relaunch) has to wait until launch phase operations are implemented.

Kiwonik:
Additional operations tests (e.g. `set_setup_parameters_apply`) should be modified to use `smt_database_fixture::create_invalid_smt` and `smt_database_fixture::create_smt_3` similarily to `setup_emissions_apply`. This however is blocked by #1738

PR #1679, opened 2017-10-23
Internal references to SMTs switched to asset symbol type

Kiwonik:
Full asset_symbol_type is now used instead of "public" nai string representation #1676

### DIFF
--- a/libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
+++ b/libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
@@ -25,11 +25,6 @@ public:
       c( *this );
    }
 
-   uint32_t get_nai() const
-   {
-      return symbol.to_nai();
-   }
-   
    // id_type is actually oid<smt_token_object>
    id_type           id;
 
@@ -69,7 +64,7 @@ public:
    uint8_t              rel_amount_denom_bits = 0;
 };
 
-struct by_nai;
+struct by_symbol;
 struct by_control_account;
 
 typedef multi_index_container <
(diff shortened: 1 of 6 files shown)


Kiwonik:
I started to push changes for issue #1653 in this branch, because they need the code waiting here to be merged.

libraries/chain/include/steem/chain/account_object.hpp
```diff
@@ -124,6 +126,10 @@ namespace steem { namespace chain {
          }
 
          asset effective_vesting_shares()const { return vesting_shares - delegated_vesting_shares + received_vesting_shares; }
+#ifdef STEEM_ENABLE_SMT
+         /// Return SMT token object controlled by this account identified by its symbol. May return nullptr!
+         const smt_token_object* get_controlled_smt( const asset_symbol_type& smt_symbol, database& db ) const;
```
mvandeberg: This appears to be beyond the scope of this PR.

mvandeberg:
Instead of pushing more changes for another branch on this branch, branch #1653 off this branch and push changes to it. Then, when this branch is merged, rebase 1653 on `develop`. A PR directly from 1653 would probably also work.

If I requested changes to this branch, then rebasing 1653 on this branch to incorporate the changes would also work.

As the developer, it is your responsibility to make the job of review as easy as possible. Either through in code documentation, or documentation in git hub. Adding extraneous code to the PR makes my job harder for 2 reasons.

1. I have to determine what code is for what spec.
2. If the branch is now a work in progress, I don't want to merge code that is actively in development.

libraries/chain/include/steem/chain/account_object.hpp
vogel76: What is the problem here ? To avoid duplication of code referencing the SMT in each place we have  added method to the account.

mvandeberg: No problem with the code.

Please see my comment above. It is difficult to review a PR when there code for other issues mixed in.

Kiwonik:
I created the subbranches and moved commits appropriately. There's a single comment in this branch / PR now. Ready to merge again.

PR #1686, opened 2017-10-23
Prune Account History for Develop

mvandeberg:
Dev PR for #1671


theoreticalbts:
Command line flag has the wrong sense, when `history-disable-pruning` is specified then `_prune` will be set to `true` which causes pruning to occur.

mvandeberg:
This is a problem for #1672 as well

PR #1688, opened 2017-10-23
No ops filtering for appbase

mvandeberg:
Disable ops filtering and bump shared memory for develop Replaces #1685


theoreticalbts:
LGTM but @jredbeard should approve this also.

mvandeberg:
Lowered shared mem size to 120G as per @jredbeard's request in Slack

PR #1690, opened 2017-10-23
Expose more p2p configuration to config file

theoreticalbts:
I'm not sure how to route the newly created parameters into the command line / config file actually being used by our nodes.  Is it some black magic involving Dockerfile and env vars?


libraries/app/application.cpp
```diff
@@ -997,6 +1016,7 @@ void application::set_program_options(boost::program_options::options_descriptio
    configuration_file_options.add_options()
          ("p2p-endpoint", bpo::value<string>(), "Endpoint for P2P node to listen on")
          ("p2p-max-connections", bpo::value<uint32_t>(), "Maxmimum number of incoming connections on P2P endpoint")
+         ("p2p-parameters", bpo::value<uint32_t>(), "P2P network parameters")
```
mvandeberg: This should be set to some default value with the options that currently exist.

The type does also not match the parsing logic in `resest_p2p_node`.

I would make a struct for the parameters with the default values set to the defaults for the config.
For the default value, construct the default initialized struct, and convert to string.

theoreticalbts:
Hopefully CI should work now

libraries/app/application.cpp
```diff
@@ -997,6 +1016,7 @@ void application::set_program_options(boost::program_options::options_descriptio
    configuration_file_options.add_options()
          ("p2p-endpoint", bpo::value<string>(), "Endpoint for P2P node to listen on")
          ("p2p-max-connections", bpo::value<uint32_t>(), "Maxmimum number of incoming connections on P2P endpoint")
+         ("p2p-parameters", bpo::value<string>()->default_value("{}"), "P2P network parameters")
```
mvandeberg: The default value needs to mean something. The only documentation of what should go here, is the source code itself. This needs to be the default allocation `node_configuration`.

libraries/net/include/graphene/net/node_configuration.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <fc/crypto/elliptic.hpp>
+#include <fc/network/ip.hpp>
+
+namespace graphene { namespace net {
+
+struct node_configuration
+{
+   node_configuration() {}
+
+   fc::ip::endpoint listen_endpoint;
```
mvandeberg: I don't think we need to endpoint listed here. All of the behavior to determine if the node should listen or not is handled in `application.cpp`.

theoreticalbts:
> I don't think we need to endpoint listed here. All of the behavior to determine if the node should listen or not is handled in `application.cpp`.

No, it is used heavily in `listen_to_p2p_network()` [here](https://github.com/steemit/steem/blob/48eef814950f7adaeffd0695d99615e4452e50b0/libraries/net/node.cpp#L4565-L4657).  Maybe this is the wrong place to put it, but doing so would require a bigger refactoring of the p2p code that should be its own issue.

mvandeberg:
Can we confirm that the p2p-endpoint listen semantics are unchanged with this patch?

theoreticalbts:
Since unspecified members do nothing, don't specify it in the config file and you should be fine.

theoreticalbts:
Perhaps I should make a more specific version of the above comment.  The `--p2p-parameters` option processing code [here](https://github.com/steemit/steem/blob/85a455cf968afd0548d846be56cb4cb1baa34f60/libraries/app/application.cpp#L141-L147) converts the user-specified JSON to a `variant_object` and passes that `variant_object` to `set_advanced_node_parameters`.  Which as you can see [here](https://github.com/steemit/steem/blob/85a455cf968afd0548d846be56cb4cb1baa34f60/libraries/net/node.cpp#L5001) uses `from_variant()` to set the `_node_configuration` member of `node_impl`.

One interesting subtle point of the behavior of `from_variant` to convert a `variant_object` called `vo` to a reflected struct instance `o` is that the reflection-generated implementation of `from_variant( vo, o )` will not change the corresponding field of `o` if the corresponding key `k` is not included in `vo`.

As a consequence, if the `listen_endpoint` field is not specified as a dictionary key in the config file JSON, then the `listen_endpoint` will not be overwritten by the new call to `set_advanced_node_parameters`.  Meaning if the user does not specify `listen_endpoint` field in `p2p-parameters`, then the `set_advanced_node_parameters` call will not modify `_node_configuration.listen_endpoint`.  So `_node_configuration.listen_endpoint` will end up being initialized the same way it was before this patch.

mvandeberg:
There is still something about this patch that doesn't feel good...
I understand this is the quickest way of enabling cli args for the p2p code, but because of how boost program options handles arguments, either the entire node config struct will commented out or be there. And conveying to a node operator to comment out certain fields in the json object seems a bit weird.

I could see it working if the entire config file were json, but because this is mixed config file...

I will approve this because it is needed, but I really don't want to.

theoreticalbts:
Let's merge it and if there is an improvement to the syntax or how / where it is stored, that can be a separate PR.

Issue #1702, opened 2017-10-24
Develop: get_state discussion_idx truncating


mvandeberg:
The discussion_idx in get_state is truncating content paths.

The culprit is `auto key = d.author +"/" + d.permlink;`

To prevent any implicit conversions, the `author` field of the `api_comment_object` should be changed to a `string` and `auto key` defined as `string key`.

mvandeberg:
This issue was not solved. Upon further investigation, the `discussion_index` struct was duplicated in `state.hpp` and `condenser_api.hpp` but was changed in `condenser_api.hpp` to be a vector of `tag_name_type`, which is a fixed length string and accounts for the truncation. It looks like `state.hpp` is not needed as all classes in it are duplicated in their respective APIs. The following changes should be made.

1. Remove `state.hpp`
2. Revert the change from`tag_name_type` to `string`.

There is one other occurrence of `tag_name_type` in an API object that is new, `tag_index`. But because the struct is specifically supposed to be a vector of tags, this is the correct usage of the type.

PR #1704, opened 2017-10-24
Develop: get_state discussion_idx truncating

mvandeberg:
PR for #1702


theoreticalbts:
- First change (`reply_ref`) only changes whitespace
- ~~Should be `string(d.author) + "/" + d.permlink`~~ This criticism is obsolete because the patch changes `author`, `parent_author` to type `string`

PR #1707, opened 2017-10-25
Revert "Prune Account History for Develop"

mvandeberg:
Reverts steemit/steem#1686

This merge is causing problems in CI for whatever reason. Reverting it while I can diagnose the problem.


mvandeberg:
It is a problem with CI.

Issue #1710, opened 2017-10-26
Encapsulate SMT finding in dedicated method of account_object.


Kiwonik:
Processing (evaluation) of every post-creation SMT operation requires doing common actions to start with - the provided account supposed to be controlling the provided SMT symbol must be found and confirmed to actually be the SMT's control account. To keep the code evaluating post-creation SMT operations neat and clean encapsulate these common finding / checking code in a dedicated method of `account_object`.

Kiwonik:
The code implementing this issue is dependent on code that implements issue #1676

PR #1716, opened 2017-10-26
Encapsulate smt finding in dedicated method of account object

Kiwonik:
Added a method to keep the code evaluating post-creation SMT operations neat and clean #1710


theoreticalbts:
This method takes three parameters, let's call them `account`, `sym`, `db`.  I can think of three different possible syntaxes for a function with these parameters:

```
account.get_controlled_smt( sym, db );   // (1)
db.get_controlled_smt( account, sym );   // (2)
get_controlled_smt( db, account, sym );   // (3)
```

- We should not do (1), since it creates an unnecessary circular dependency (`database` already depends on `account_object`, with this change `account_object` now depends on `database`).
- (2) is how many existing small "helper" functions are structured.  This is bad and we want to move away from it, because the `database` class is very large, very hard to fully understand since it contains many small snippets from mostly unrelated subsystems, and takes a very long time to compile.
- (3) is what I would recommend.

Also, more minor issues:

- This function's behavior only accesses the `name` field of the `account_object`.  So instead of passing in `account_object`, just pass in the account name.  This might allow us to avoid looking up an `account_object` in future call sites where this function will be used.
- In this codebase, a function that returns `nullptr` when the result is not found is called `find_whatever()`.  A method that throws exception when the result is not found is called `get_whatever()`.  So this method should be renamed to `find_controlled_smt()` since it returns `nullptr` when it is not found.

As a consequence of the above, we might also implement `get_controlled_smt( db, account, sym )` which essentially is `result = find_controlled_smt( db, account, sym ); FC_ASSERT( result != nullptr ); return *result;`.  Then in `common_pre_setup_evaluation()` the first four lines can be rewritten as two lines:

```
   const account_object& account = _db.get< account_object, by_name >( control_account );
   const smt_token_object& smt = get_controlled_smt( db, account, symbol );
```

We can omit the `nullptr` checks because this check occurs in the `get_` function.

vogel76:
1. Your note (1): Placing `get_controlled_smt` method in `account_object` does not add any circular dependency to database:
-  All these classes reside in same module (chain)
- database.hpp does not contain include account_object.hpp, so it could be even possible to define this method directly in class body, by adding `#include <steem/chain/database.hpp>` to the account_object.hpp directly.  We didn't decide to do it, since being experienced in development of really large and longterm C++ projects, we know that it could pollute headers and make problems in the future. Thus we decided to use forward declaration & implementation in .cpp what makes things safe according to future changes.
1. Your note (3): Defining external function as you proposed is bad solution, since it makes harder code maintenance and understanding. The one of features supported by account_object is to provide association to smt_token_object being controlled via given account. Since C++ language allows OOP, given method was added instead of pseudo C approach utilizing unassociated structures and functions. 
1. Your minor notes: I agree, this method should not return nullptr, but shall to throw exception in case there is no matching SMT to given asset_symbol_type. This will be fixed ASAP.

Kiwonik:
> 2. Your minor notes: I agree, this method should not return nullptr, but shall to throw exception in case there is no matching SMT to given asset_symbol_type. This will be fixed ASAP.

Implemented and committed.

mvandeberg:
Because an account can now create multiple SMTs, I believe the approach of this method is no longer valid. @youkaicountry, can you confirm and close this PR if that is the case?

Kiwonik:
Actually, this approach works with any number of SMTs created by given account. It just returns the smt_token_object corresponding to given asset_symbol_type.

youkaicountry:
Definitely should still work with multiple SMTs per account since it finds the SMT by symbol, then verifies the given account controls it. If it's still useful then we will keep it.

Kiwonik:
It is useful, actually SMT's #1653 uses it.

mvandeberg:
I am just thinking that the only operations that really need to use this are ones that SMT creators use. And of those, the only op that needs to access the account object is `smt_create_operation` to pay the fee. Once the SMT is created, if you have the creator as a field of the SMT object, then you should never need to access the SMT object again. This would mean this method is either inefficient to use in most cases, and/or will only be used in one evaluator, so creating a helper method for just that doesn't seem worth it.

I am good with merging this, but if it is only used for one or two things, I think it should be removed and placed in the evaluator itself.

youkaicountry:
That sounds reasonable. If this function isn't going to be used much, then it would probably be best to scrap it.

Kiwonik:
@mvandeberg @youkaicountry The method will be used by all operations issued by SMT creators that modify SMT settings including but not limited to those impacting SMT ICO (with the exception of `smt_create_operation` actually). For each of the operations we will need to start evaluation with 1. Checking that the user that issued the operation is the one who created/controls the SMT and 2. Accessing the SMT object to modify SMT setting/state stored in it. This is exactly what this method does and so its call will be a part of these operations evaluators: `smt_setup_emissions_operation`, `smt_set_setup_parameters_operation`, `smt_set_runtime_parameters_operation`, `smt_setup_operation`, `smt_cap_reveal_operation`.

mvandeberg:
`smt_create_operation` requires a lookup of the account object to pay the fee. Past that, there is no reason to look up the account object to access the SMT object because there is currently no information in the account object that is needed to evaluate the other ops. All you need to do is look up an SMT by NAI and confirm the owner matches the op (exactly like you do). While, yes, the method you have written does so, it does so inefficiently because it requires calling the method **on** the account object, which I just established serves no other purpose other than providing this method.

What I would do is this, add a composite key to the SMT object (NAI, owner, id) called `by_nai_owner`. Then, the logic to check the existence of an NAI becomes

```c++
auto smt = _db.find< smt_object, by_nai_owner >( boost::make_tuple( op.nai, op.owner ) );
FC_ASSERT( smt != nullptr, "The SMT ${nai} does not exist or ${account} does not control the SMT.", ("nat",op.nai)("account",op.owner) );
```

If you want more detailed errors messages regarding ownership vs non-existence, then the three line version currently used would be required.

As per theoreticalbts's recommendation above, I would also recommend option (3). Preferably, inlined and local scope to `steem_evaluator.hpp` if it is not needed elsewhere.

> get_controlled_smt( db, account, sym );   // (3)

Kiwonik:
I went with the option (3), inlined and local scope to `smt_evaluator.cpp` as we found the only use of this method is in SMT operations evaluation.

Kiwonik:
@youkaicountry Ready to merge.

youkaicountry:
@Kiwonik Just to be safe, I would recommend making `db` in `get_controlled_smt` and `_db` in `common_pre_setup_evaluation` both const.

Kiwonik:
@youkaicountry Done. Ready to merge again.

PR #1717, opened 2017-10-26
adds broadcaster/broadcastsync low memory node for paas/beanstalk

jredbeard:
This allows adding the flag `IS_BROADCAST_NODE` to launch a low memory node that has the sole purpose of handling things from `network_broadcast_api` to be routed through jussi. They will be launched as the beanstalk apps `broadcaster` and `broadcastsync`.


mvandeberg:
@jredbeard Is this still something we are doing? Do you want this merged?

theoreticalbts:
Why are these kind of implementation details of Steemit's internal infrastructure in this repo?

mvandeberg:
Necessary evil for the time being. We want to move details such as this to their own repo. But I don't believe we are using these node types, so the PR may be closable. @jredbeard Can you confirm?

jredbeard:
While I agree that we should move any implementation details out into their own repo eventually, it also has not been the highest priority among other tasks. This isn't C++ and is all in it's own directory, so it shouldn't effect anyone either way.

We are not using 'broadcaster' nodes at the moment but we'd like this functionality to be available in case we do again. I think we should still merge this in.

PR #1722, opened 2017-10-26
Make perform_vesting_share_split() a no-op if it happens at genesis

theoreticalbts:
When creating a testnet, I've encountered some trouble with total `VESTS` exceeding `STEEM_MAX_SATOSHIS` if the vesting split occurs when only `1.000 TESTS` is vested.  The issue can be fixed by refusing to perform the split hardfork on the genesis block.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -4088,7 +4088,14 @@ void database::perform_vesting_share_split( uint32_t magnitude )
 {
    try
    {
-      modify( get_dynamic_global_properties(), [&]( dynamic_global_property_object& d )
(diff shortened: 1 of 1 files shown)


mvandeberg:
This impacts how the unit tests execute and has changed values such that the tests now fail.

mvandeberg:
If you are using the same HF trigger as #1723, then that explains the behavior. The genesis hardfork application in #1278 applies the hardfork before everything else (when there are 0 satoshi in the vesting fund) where as `process_hardforks` occurs after the first block reward.

mvandeberg:
Replaced by #1748

PR #1723, opened 2017-10-26
Apply all hardforks in the past at genesis

theoreticalbts:
If we generate a genesis block after some of the hardfork(s) have been scheduled, those hardforks take effect in the genesis block.


mvandeberg:
I am inclined to close this PR to use #1278 for the reasons stated in #1722

mvandeberg:
Replaced by #1748

Issue #68, opened 2016-05-25
cli_wallet can't deal with multiple accounts in authorities


xeroc:
The account `streemian` looks like this:

```
  "owner": {
    "weight_threshold": 1,
    "account_auths": [[
        "xeroc",
        1
      ]
    ],
    "key_auths": [[
        "STM7H3YNtDHtGMP5yqA4hGxWDzCnB9TPwb7ocbHzoT2ePLzG6bBpi",
        1
      ]
    ]
  },
```

The other permissions only hold a key.
When I try to transaction from that account, I get:

```
10 assert_exception: Assert Exception
it != approving_account_lut.end(): 
    {}
    th_a  wallet.cpp:598 operator()

    {"ops":[["withdraw_vesting",{"account":"streemian","vesting_shares":"22189715.573908 VESTS"}]],"sigs":[]}
    th_a  transaction.cpp:270 verify_authority
```

abitmore:
I'm afraid that CLI doesn't support "account authorities" so far. There is a TODO in the [code](https://github.com/steemit/steem/blob/v0.4.3/libraries/wallet/wallet.cpp#L577):
```
      /// TODO: recursively check one layer deeper in the authority tree for keys
```
BTW same issue found in BitShares CLI.

xeroc:
With recent events, can we have this fixed please?!

xeroc:
bug seems to be fixed in 
https://github.com/steemit/steem/commit/a6684ce

bacchist:
Not sure whether this is the same issue, or another one... (sorry for the formatting)

[...]

PR #1726, opened 2017-10-27
Verify Signatures API Call - Appbase

mvandeberg:
PR for #1674

### DIFF
--- a/libraries/plugins/apis/database_api/database_api.cpp
+++ b/libraries/plugins/apis/database_api/database_api.cpp
@@ -4,6 +4,8 @@
 #include <steem/plugins/database_api/database_api_plugin.hpp>
 
 #include <steem/protocol/get_config.hpp>
+#include <steem/protocol/exceptions.hpp>
+#include <steem/protocol/transaction_util.hpp>
 
 namespace steem { namespace plugins { namespace database_api {
 
@@ -60,6 +62,7 @@ class database_api_impl
          (get_potential_signatures)
          (verify_authority)
          (verify_account_authority)
+         (verify_signatures)
 #ifdef STEEM_ENABLE_SMT
          (get_smt_next_identifier)
 #endif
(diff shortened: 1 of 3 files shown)


theoreticalbts:
- Why is `authority::key` not supported?
- Contains unrelated debug code in `follow_plugin.cpp`
- In `verify_authority()`, it seems needlessly complicated to pass in a callback which is immediately called, can we perhaps just pass in `required_active`, `required_owner`, `required_posting`, and `other`?   Perhaps these four fields should be in their own `struct`?

It's functionally correct, so get rid of the debug code in `follow_plugin.cpp` and this is mergeable.  The other two seem like maybe separate PR's.

mvandeberg:
Why would we make an API call that verifies the signature of a single key (or set of keys)? There are many other crypto libraries that can do that. The purpose of this call is to leverage Steem account authorities and multi-sig to verify signatures.

mvandeberg:
FYI, this was already merged to `master` in #1725 

Any changes requested here need to be duplicated for `master` and merged there as well.

mvandeberg:
I am going to generalize the `opertation_get_required_authority_visitor` to work for both a static variant and a base type that supports the interface so that we can specify all auth types in a struct that behaves like an operation. This will remove the weird lambda semantics.

There is a fine line we need to walk with APIs between making the APIs easy enough to use while not performing needless calculations in the API.

theoreticalbts:
Minor issues:

- Merge conflicts need to be addressed.
- The [recursive visitor invocation](https://github.com/steemit/steem/blob/f5913e27fed77d0f572fab704bea69b56f09bf03/libraries/protocol/include/steem/protocol/operation_util.hpp#L58) could be simplified to `v.visit(*this);` ?

Major issues:

We're refactoring some of the most important core consensus logic here:  Who's allowed to do what transactions.  I'd like to see the consensus changes here split out into their own commits, with a description in the ticket of the transformations that were applied to existing code to arrive at the new code.  Then the reviewer can verify it is "correct-by-construction" by checking that (a) the specified sequence of transformations yields the code in the commit, and (b) the specified sequence of transformations does not affect the functionality of the code.

This refactoring doesn't seem like it actually simplifies the code.  The problem, discussed in more detail in my comment yesterday on our oldest open ticket #68 , is that there are three, or sometimes four, parallel things that need to be passed around when dealing with authority handling functions.  If we are going to open this particular can of worms -- a (badly needed IMHO) refactoring of the internal API's for consensus authority logic -- we need to think carefully about how it should be designed and implemented.  I think the branch attached to #68 is a better approach.  We should consider the possibility of merging that branch (which does not change consensus), and then (once it is adequately reviewed and tested) using it as a base for a correct-by-construction refactoring of the consensus logic.

I've started and abandoned several attempts to refactor the consensus logic, eventually I decided that the best bet is to wait for the #68 branch to be merged, and then use that branch as the basis for a refactoring the consensus logic.  While this is certainly a critical review, I'm just as critical of my own code; #68 is the product of several "ok-let's-refactor-this, type-type-type, hmm-this-really-isn't-working-out-i-guess-i-should-just-nuke-it-and-start-over" cycles.

mvandeberg:
We have some collision of priorities here. While I don't disagree that some refactoring is necessary, we are about to run into a problem that other backend services are planning on using this API to authenticate users. Delaying the API and delaying the release of these services to do the refactor may not be an option.

If I fix the merge conflict, and restructure the commits with a bit more documentation, is this PR mergable at all? Or will it get blocked due to the desire to refactor the authority code in its entirety?

theoreticalbts:
> Delaying the API and delaying the release of these services to do the refactor may not be an option.

If we're going to refactor the consensus level authority API's, then we need to take our time and do it right.  The consensus level authority code is absolutely critical to the security of the blockchain.

As a logical consequence, we need to choose between the following options for implementing this API call:

- (a) Implement the API call without refactoring the consensus level authority code
- (b) Push back the release of the API call until after the consensus level authority code has been refactored

If this RPC call is needed sooner rather than later, then (b) is the less desirable option.  So let's go with option (a).

mvandeberg:
#1841 contains the refactor of the consensus code. This PR will be updated to contain just the API changes.

mvandeberg:
#1841 needs to merged first

mvandeberg:
@theoreticalbts This is ready for review.

theoreticalbts:
The OP of #1674 says, "To match behavior of steemd transaction signing, it will fail on duplicate signatures, but will throw an exception on that failure rather than silently fail by returning false."  The code in this ticket returns `false`.  I suggest adding a comment to #1674 (or perhaps editing the OP) to explain the behavior is actually returning `false`.

mvandeberg:
That is not correct, there is a `STEEM_ASSERT` on the duplicate signature check that mirrors the behavior of transaction signature verification.

Issue #1736, opened 2017-11-01
Add account name to get-dev-key output


youkaicountry:
The json dictionary output of get-dev-key should include the account names in the associated dictionaries, for better reference.

youkaicountry:
PR #1739

PR #1739, opened 2017-11-02
Added account_name to the json output of get_dev_key. #1736

youkaicountry:
Adds a new field with the account name to the output of get_dev_key.


theoreticalbts:
Looks reasonable.

Issue #1747, opened 2017-11-09
Apply all hardforks in the past at genesis


theoreticalbts:
Right now we have two competing implementations of this functionality, both of which break things:  #1723 and the [1278-steemit-testnet branch](https://github.com/steemit/steem/tree/1278-steemit-testnet).

In the case of #1723, it is not usable without #1722 which breaks many unit tests.

In the case of the branch,

- No PR has been created from this branch, so its developers do not consider it ready to merge
- It fails CI (but only one test)
- It contains unrelated changes (mostly configuration constant changes)

So I'm going to write a clean PR to do pre-update at genesis

mvandeberg:
Should have been closed a year ago.

PR #1748, opened 2017-11-10
Apply all hardforks in the past at genesis #1747

theoreticalbts:
If this PR is merged, please close PR's #1722 #1723.


theoreticalbts:
I started by porting the 1278 branch, then realized later I didn't need it, but left it in.  Since the flag isn't actually used, let's take it out of there for now.  If we ever want it, we can always add it again.

PR #1752, opened 2017-11-12
fix documentation build

piem:
see bitshares/bitshares-core#409 and bitshares/bitshares-core#474


theoreticalbts:
Looks good to me.

PR #1753, opened 2017-11-12
avoid compiler warnings when using g++-6

piem:


theoreticalbts:
Here's what needs to happen with this code:

- Some of these commits need further work, as detailed below.
- The mergeable commits should be rebased on `develop`. 
- Someone with commit privileges on this repo will need to make an internal PR against `develop` for these changes, so they can go through CI.

> avoid throwing exception in destructors

The problem here is that your change will lose the information that an exception is thrown.  Let's do something more noisy here than merely logging the exception.  The "proper" thing to do is to have the destructor set a flag accessible to some outer context and then have the outer context check the flag and throw an exception, but this seems complicated, especially given that the outer context is provided by Boost macros, not code we control.

But this is in the unit tests.  The unit tests' primary function is to be run by the CI script to make sure all is well with every commit.  So let's simply `exit(1)`.  It's not as ideal as the proper solution, but it is easy to implement and would meet the desired requirements (cause the test process to exit unsuccessfully if an exception is thrown in one of these destructors, fix warnings on g++ 6).

> check symbols of each input parameters

It took me a while to realize that this commit doesn't change anything.  These lines are merely recording for the payout amount, not actually part of consensus (despite being in `database.cpp`).  It must so happen that `curator_sbd_value.symbol` and `beneficiary_value.symbol` must always be SBD at this point, so the newly added `if` statements must always be passing.  Because if they weren't SBD, then the asset addition would fail.  And because this is per-block processing, the entire block would be discarded, and we would quickly notice the whole network was stopped.

So really this commit doesn't actually change any functionality, the conditions of the two added `if` statements are always `true`.  But I think it's clearer to simply change the indentation and add a comment stating that these values must be SBD at this point.

Some additional notes:

> avoid overflow [g++-6]

This changes the bandwidth calculation if the compiler's truncating this expression to 32 bits.  If this was a consensus change, then we'd have to gate this with a hardfork (basically you'd have to do the new calculation in newly created fields, keeping the old calculation in the old fields, then switchover to ignoring the old fields once the hardfork becomes enabled.)  But since bandwidth is soft consensus only, the change stands as-is.

Also, the bandwidth computation should converge to almost the same value for current blocks regardless of conditions far in the past, so changing its initialization here shouldn't affect much.

> libraries/wallet/CMakeLists.txt: add PWD to PERLLIB before running generate_api_documentation.pl

Duplicate of #1752

> programs/util/get_dev_key.cpp: show help if less than 2 parameters are given
> programs/util/get_dev_key.cpp: avoid undefined variable

The CLI processing of this tool has been updated on `develop`.  While these commits do address deficiencies, they may have merge conflicts.

piem:
hi @theoreticalbts. ok, thanks for the comments, I will rebase on `develop` and split in separate PRs for clarity.

mvandeberg:
Closing this PR in favor of the aforementioned multiple PRs.

piem:
This PR was splitted into #1752, #1755, #1756, and #1757.

- `libraries/wallet/CMakeLists.txt: add PWD to PERLLIB before running ge` 6e612e6
  see PR #1752
- `libraries/chain/database.cpp: check symbols of each input parameters,` b3e17ce
   see PR #1757, sure, I see now there is a big `FC_ASSERT` in `asset +=` of course. I just thought more input validation wouldn't hurt. I removed the if altogether and added a comment.
- `tests/common/database_fixture.cpp: avoid throwing exception in destru` 4ca0826
  see PR #1755, added calls to `exit(1)` after `FC_CAPTURE_AND_LOG` (`return` should be reached otherwise)
- `libraries/plugins/witness/witness_plugin.cpp: avoid overflow [g++-6] ` f9e7cda 
  not needed, better solution in develop
- `programs/util/get_dev_key.cpp: avoid undefined variable              ` ef6a34b
  not needed, already in develop
- `programs/util/get_dev_key.cpp: show help if less than 2 parameters ar` 0039478
  see PR #1756 
- `tests/tests/operation_tests.cpp: avoid multiline comment [g++-6]     ` 24b2e68
  not needed, better solution in develop

PR #1754, opened 2017-11-13
fix documentation build

mvandeberg:
Local PR for #1752


theoreticalbts:
Shouldn't this be against `develop`?

mvandeberg:
The original PR was against `master`.

mvandeberg:
What is preventing this from being merged?

PR #1756, opened 2017-11-13
get_dev_key.cpp: show help if less than 2 parameters are given

piem:
also remove the (debugging?) first line shown before the help


mvandeberg:
Replaced with #1774

PR #1758, opened 2017-11-13
get_state discussion_idx truncating

mvandeberg:
PR for #1702


theoreticalbts:
I'm surprised that header is unused.  But if it compiles and passes CI, of course it must be, so let's get rid of it.

mvandeberg:
The header exists in `app` on `master`. Part of the appbase refactor broke some the API calls in to the appropriate plugin. (i.e. `get_discussions_by_trending` to `tags_api`) The header was broken up, but never deleted. This lead to confusion of the actual problem in issue.

PR #1759, opened 2017-11-13
#963 appbase implementation of fork monitoring plugin with unit test

mvandeberg:
Local PR for #1715

Related #963

### DIFF
--- a/libraries/plugins/fork_monitor/CMakeLists.txt
+++ b/libraries/plugins/fork_monitor/CMakeLists.txt
@@ -1,0 +1,17 @@
+file(GLOB HEADERS "include/steem/plugins/fork_monitor/*.hpp")
+
+add_library( fork_monitor_plugin
+             fork_monitor_plugin.cpp
+             ${HEADERS} )
+
+target_link_libraries( fork_monitor_plugin chain_plugin )
+target_include_directories( fork_monitor_plugin
+                            PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" )
+
+install( TARGETS
+   fork_monitor_plugin
+
+   RUNTIME DESTINATION bin
+   LIBRARY DESTINATION lib
+   ARCHIVE DESTINATION lib
+)
--- a/libraries/plugins/fork_monitor/fork_monitor_plugin.cpp
+++ b/libraries/plugins/fork_monitor/fork_monitor_plugin.cpp
@@ -28,50 +28,33 @@ namespace detail {
 
          void on_block( const signed_block& b );
 
-         std::map< uint32_t, std::vector< signed_block > > _last_blocks;
+         std::map< uint32_t, signed_block > block_map;
          chain::database& db;
    };
 
    void fork_monitor_plugin_impl::on_block( const signed_block& b )
    {
-
      auto lib = db.get_dynamic_global_properties().last_irreversible_block_num;
-
-      if( b.block_num() > lib ) // Should not be needed, but just in case
-         _last_blocks[ b.block_num() ].push_back( b );
-
-      if( _last_blocks.find( lib ) != _last_blocks.end() )
-      {
-         auto& blocks = _last_blocks[ lib ];
-         fc::optional< signed_block > included_block = db.fetch_block_by_number( lib );
-         std::set< signature_type > printed_blocks;
-         auto it = std::begin(blocks);
-
-         /* remove duplicate orphan blocks */
-         while (it != std::end(blocks)) {
-            if( it->witness_signature != included_block->witness_signature
-                && printed_blocks.find( it->witness_signature ) != printed_blocks.end() )
-            {
-                it = blocks.erase(it);
+     if( b.block_num() > lib ) { // Should not be needed, but just in case
+         /* Check for deviant block */
+         signed_block old_block;
+         map<uint32_t, signed_block>::iterator it = block_map.find(b.block_num());
+         if(it != block_map.end()){
+            old_block = it->second;
+            if(old_block.witness_signature != b.witness_signature){
+               ilog( "Orphaned Block ${n} --- Data:${d}", ("n", old_block.block_num() )("d", old_block) );
+               block_map[ b.block_num() ] = b;
             }
-            else {
-              printed_blocks.insert( it->witness_signature );
-              ++it;
+         }
+         else {
+            block_map[ b.block_num() ] = b;
+            if(block_map.size() > 21){
+               it = block_map.find(lib);
+               block_map.erase ( block_map.begin(), it );    // erasing up to lib
             }
-          }
-
-          for( auto& orphan : blocks )
-          {
-            if( orphan.witness_signature != included_block->witness_signature )
-            {
-              ilog( "Orphaned Block ${n} --- Data:${d}", ("n", lib )("d", orphan) );
-          }
-        }
-         _last_blocks.erase( lib );
+         }
       }
-
    }
-
 } // detail
 
 fork_monitor_plugin::fork_monitor_plugin() : _my( new detail::fork_monitor_plugin_impl() ) {}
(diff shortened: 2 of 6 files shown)


theoreticalbts:
The algorithm of this PR is not totally clear.  But after some reading, I think I understand the intent of this patch:

- (a) Maintain the invariant that `_last_blocks` contains, for each height above LIB, a vector of blocks seen with that height.
- (b) Once an entry in `_last_blocks` "sunsets" below LIB, it is printed on the console.
- (c) If an already-existing block would be inserted into the vector, then de-duplicate it.

Let's think about these two questions:

- Are requirements (b) and (c) desirable features?
- Are requirements (b) and (c) implemented correctly?

The answer to the second question is "no."  Well, I'm not entirely sure if requirement (c) is implemented correctly, I'm still not sure what the reason is behind the `it->witness_signature != included_block->witness_signature` check.

But as for requirement (b), quickly skimming the code shows that it only acts at the LIB height, and doesn't consider that LIB sometimes doesn't advance, and other times advances by more than 1 block.  Consider the following sequence of events, where Alice and Bob are witnesses:

- Alice produces block number 101
- Bob, unaware of Alice's block, produces a different block number 101
- The other witnesses all build on Alice's block 101, nobody builds on Bob's block 101
- The network advances on Alice's fork until LIB = 100, no orphans are registered
- The next block results in LIB advancing from 100 directly to 102, there is never a point in time at which LIB = 101
- Since the code in this plugin only examines blocks at the LIB height, it does not ever examine Bob's block.

theoreticalbts:
So I'm thinking the patch should maybe use a different algorithm, maybe something like this:

- Change `_last_blocks` to be `map< uint32_t, signed_block >`
- If a block already exists in `_last_blocks` at the same height as the new block, print the old block as an orphan
- Add the new block to `_last_blocks` at its height (block number)

We also need a cleaning step to keep from filling memory with old blocks, so whenever the first height in `_last_blocks` is below LIB, we should delete the first block from `_last_blocks` (and repeat the check until the first height is above LIB, since multiple blocks may need to be deleted).

This algorithm doesn't satisfy requirement (b), for example if we have the following sequence of events:

- Block 100 builds on block 99
- Block 101a builds on block 100
- Block 101b builds on block 100
- Block 102b builds on block 101b
- Block 102a builds on block 101a
- Block 103a builds on block 102a
- There are no further forks, the network builds a single chain on 103a

Then even though 101a ends up in the main chain, my algorithm will still print it.  The OP algorithm appears to be designed to suppress this printing.  This complicates the OP algorithm since it must defer deciding whether to print 101a or 101b until LIB >= 101.  But I think it's diagnostically more useful to know that the network switched away from a fork and then switched back.  Additionally if the network gets into a near 50/50 split and LIB is not advancing, the deferred decision/printing will never occur, meaning the log messages won't be particularly helpful in analyzing the situation.

This algorithm also doesn't satisfy requirement (c).  Again, only highly unusual network behavior (such as a network split that remains exactly 50/50 for a long period) result in a significant number of duplicate blocks, and it might be useful in such a situation to know the exact order in which fork switches occurred.  Attempting to de-duplicate multiple instances of the same entry complicates the algorithm, is only relevant in unusual situations, and destroys potentially valuable diagnostic information about the exact sequence of fork switching events that occurred.

So it looks like maybe we should get rid of requirements (b) and (c), and go with the algorithm I propose here instead.

@mvandeberg @youkaicountry any comments on the above?

NateBrune:
Thanks @theoreticalbts for taking some time to really analyse what's going on here. 
Firstly I'd like to say that more or less you were correct with the intent of the algorithm. The goal is to maintain a list of seen blocks, and after an entry sunsets below the LIB, the seen blocks which fork off the main chain are logged. 

To address this:
>  Well, I'm not entirely sure if requirement (c) is implemented correctly, I'm still not sure what the reason is behind the it->witness_signature != included_block->witness_signature check.

That looks like a bug to me as well. I believe it's left over from an early implementation. I'll fix this but be advised that otherwise the de-duplication loop functions as I intended. The purpose of the de-duplication is that a block, 101b may be added to _last_blocks multiple times before the network decides to switch permanently to 101a. I took the liberty of removing duplications without being instructed to do so. While this may be useful for de-cluttering I can absolutely see the argument that it takes information out of the log that could be useful for debugging.

As for your second point: 
> doesn't consider that LIB sometimes doesn't advance, and other times advances by more than 1 block. 

Again spot on, I wasn't considering that when I wrote the algorithm. That can be mitigated with relative ease by just writing the print loop to output everything in _last_blocks up to the LIB.

If (b) and (c) aren't requirements, than I'm very much of the opinion that a variant of @theoreticalbts' proposed algorithm should be used. It should remove a lot of the complexity and it preserves chronology of the fork. The main reason behind the sunset design was to remove duplications because initially the idea was to store each orphan in a folder on disk. Later it was decided that an ilog would be satisfactory. If de-duplication isn't a concern, this algorithm can be rewritten to be cleaner and more useful.

mvandeberg:
I am fine with this implementation, but it will result in more log spamming. The original implementation would only print at block height when LIB is updated. This implementation will print the entire fork when we switch.

In the case of micro forks, this is not problematic, but in the case of heavy forking, this will increase logging significantly.

theoreticalbts:
The new architecture/code looks good to me for the most part.  The one problem is this code:

```
    if(block_map.size() > 21)
    {
        it = block_map.find(lib);
        block_map.erase( block_map.begin(), it );    // erasing up to lib
    }
```

The code above makes an assumption:

- If `block_map` contains more than `21` entries [1], then it must contain LIB.

This assumption is not true.  It is possible for LIB to be *arbitrarily far behind* the head block number [2]!  So you definitely need to handle the case when `lib` is not in `block_map`.

The cleaning step should probably just implement what I suggested in a previous comment:

> Whenever the first height in `block_map` is below LIB, we should delete the first block from `block_map` (and repeat the check until the first height is above LIB, since multiple blocks may need to be deleted).

The easiest way to implement this is with a `while()` loop, but there is probably some way to get it to work with `find()` / `erase()`.

[1] As a nitpick, we should use `STEEM_MAX_WITNESSES` instead of the "magic number" 21.

[2] If you have enough witnesses to keep the blockchain running (as specified by `required_active_participation` [3]) but not enough to advance LIB (as specified by `STEEM_IRREVERSIBLE_THRESHOLD`), the blockchain can grow to arbitrarily many blocks while LIB does not advance.

[3] Since `required_active_participation` is a local setting for each witness, at the most extreme case any single witness could unilaterally set `required_active_participation = 0` and start producing a chain containing only their own blocks!

NateBrune:
@theoreticalbts I disagree with your fundamental assessment of the code. Since it's the only branch of the code which removes elements from `block_map`, and it only removes up to the LIB, there is no point at which this assumption is made:
* If `block_map` contains more than 21 entries, then it must contain LIB.

I'm willing to implement the exact strategy you've outlined here, but I don't like it very much.

> Whenever the first height in block_map is below LIB, we should delete the first block from block_map (and repeat the check until the first height is above LIB, since multiple blocks may need to be deleted).

After reviewing the exact code I still think it's sane. The size condition is arbitrary and you're right, but seeing as this only erases up to the lib at all times as I understand it would still be functional in these scenarios you've outlined. To be more verbose and accurate we could do something like this
```
if( b.block_num() > lib ) {...} //insert
else {
   block_map[ b.block_num() ] = b;
   it = block_map.find(lib+1);
   if(it != block_map.end() && it != block_map.begin())
      block_map.erase ( block_map.begin(), it );    // erasing up to and including lib
}
```
Would that be satisfactory? I think it would be okay even in the worse case you outlined in footnote [3].
There are a few different ways to implement this but I believe using the standard modifiers is the best option.

mvandeberg:
This has not been touched in several months months and was never updated based on review.

PR #1783, opened 2017-11-17
Retain JSON RPC on more errors

### LINKED ISSUE #1776: JSON RPC can lose request ID
In the following code:

``` c++
try
{
   const auto& request = message.get_object();

   rpc_id( request, response );
   if( !response.error.valid() )
      rpc_jsonrpc( request, response );
}
catch( ... )
```

If there is an unexpected error in `rpc_jsonrpc`, then the request ID is lost. If the ID is successfully parsed, under no circumstances should it ever be lost. The solution is adding another `try/catch` around `rpc_jsonrpc` so that if an unexpected error occurs in that function, the ID is not lost and can be returned with the response.

mvandeberg:
Closes #1776

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,7 @@ RUN \
         .. && \
     make -j$(nproc) chain_test test_fixed_string plugin_test && \
     ./tests/chain_test && \
+    ./tests/plugin_test -t json_rpc && \
     ./programs/util/test_fixed_string && \
     cd /usr/local/src/steem && \
     doxygen && \
@@ -76,6 +77,7 @@ RUN \
         .. && \
     make -j$(nproc) chain_test test_fixed_string plugin_test && \
     ./tests/chain_test && \
+    ./tests/plugin_test -t json_rpc && \
     ./programs/util/test_fixed_string && \
     cd /usr/local/src/steem && \
     doxygen && \
@@ -99,6 +101,7 @@ RUN \
         .. && \
     make -j$(nproc) chain_test plugin_test && \
     ./tests/chain_test && \
+    ./tests/plugin_test -t json_rpc && \
     mkdir -p /var/cobertura && \
     gcovr --object-directory="../" --root=../ --xml-pretty --gcov-exclude=".*tests.*" --gcov-exclude=".*fc.*" --gcov-exclude=".*app*" --gcov-exclude=".*net*" --gcov-exclude=".*plugins*" --gcov-exclude=".*schema*" --gcov-exclude=".*time*" --gcov-exclude=".*utilities*" --gcov-exclude=".*wallet*" --gcov-exclude=".*programs*" --output="/var/cobertura/coverage.xml" && \
     cd /usr/local/src/steem && \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -76,6 +77,7 @@ RUN \
         .. && \
     make -j$(nproc) chain_test test_fixed_string plugin_test && \
     ./tests/chain_test && \
+    ./tests/plugin_test -t json_rpc && \
     ./programs/util/test_fixed_string && \
```
theoreticalbts: Why aren't we running all plugin tests?

tests/db_fixture/database_fixture.cpp
```diff
@@ -671,13 +671,41 @@ fc::variant json_rpc_database_fixture::get_answer( std::string& request )
...
    int64_t answer_code;
 
    if( is_fail )
    {
+      if( id.valid() && code != JSON_RPC_SERVER_ERROR && code != JSON_RPC_INVALID_PARAMS )
+      {
+         BOOST_REQUIRE( answer.get_object().contains( "id" ) );
+         check_id_equal( answer[ "id" ], *id );
+      }
+
```
theoreticalbts: It looks like the new code will forward the ID when these values of `code` occur, so there are at least some cases where we do want tests to enforce that the ID is forwarded when these errors occur.  Let's just require the caller to set `id` to empty if we want to skip this check in a specific test.

mvandeberg: We can't. Setting the ID to empty is against the jsonrpc standard.

Dockerfile
mvandeberg: Some are broken. Because the plugin tests were never ran, they weren't updated alongside other code. I'll file a ticket to fix them, but I don't believe it should be a blocker for this PR.

mvandeberg: For this, I was mistaken. All plugin tests pass. Will change in the Dockerfile

theoreticalbts:
I'm not sure I was clear about my suggestion.  We have the following `if` statement in the `review_answer()` fixture method:

```
if( id.valid() && code != JSON_RPC_SERVER_ERROR && code != JSON_RPC_INVALID_PARAMS )
{
   BOOST_REQUIRE( answer.get_object().contains( "id" ) );
   check_id_equal( answer[ "id" ], *id );
}
```

Let's change the condition to simply `if( id.valid() )`

My above comment was noting that *if* this change causes some test which calls `review_answer()` to improperly fail, *then* we should rewrite that test to call `review_answer()` with the `id` parameter of the `review_answer()` method unset.

However, as it happens, I tried to make this change and `plugin_test` passes.  So all we have to do is add the one-line change of the `if` condition to `if( id.valid() )`, and this PR will be ready to merge.

theoreticalbts:
Looks like you went above and beyond and got rid of a flag too!  That's pretty cool.  As soon as CI is done, this is clear to merge.

mvandeberg:
I wanted to be sure the tests adhere to the json rpc spec. The only case we shouldn't check the ID is if the ID is invalid in the request, which returns that error. The if statement effectively now says, if there was not an error parsing the id locally and not an error parsing the id in the plugin, then we check that they match.

PR #1790, opened 2017-11-21
Fix JSON RPC tests framework

### LINKED ISSUE #1788: JSON RPC plugin test does not follow steemd testing practices
The set up of the JSON RPC test is weird and does not follow other test practices. The most egregious of these mistakes is not using a temporary directory.

https://github.com/steemit/steem/blob/develop/tests/db_fixture/database_fixture.cpp#L623-L659

However, that should be handled by refactoring the test to use the existing database test fixture in an expected manner. Rather than connecting and sending the rpc requests over a local socket, the test should send strings directly to the plugin via the `call` method.

The best practice rules informing this issue are:

- Steemd tests should not have consequences on the host system
- Steemd unit tests should, to the best of their ability, test just the code in question

mvandeberg:
Closes #1788

Should be reviewed/merged after #1783


theoreticalbts:
Looks good, let's merge it.

Issue #1554, opened 2017-09-13
Rename smt_elevate_account -> smt_create


theoreticalbts:
This operation needs to be renamed.

theoreticalbts:
In addition to renaming I'm going to add the field for the NAI and serialization code to put it in an asset identifier.

vogel76:
Following doubts must be clarified:
1) Does single control account can create multiple SMTs (having different NAIs) ? This is reasonable according to separation from controlled account and SMT as it was proposed initially.
2) When given NAI shall be reserved and stored in `smt_creation_operation` ? Operation evaluation stage is reasonable, especially the spec describes that NAI is allocated by blockchain.
3) How SMT symbol uniqueness shall be determined ? SMT ids share global space. `asset_symbol_type::validate` actually checks correctness of given SMT symbol according to NAI/precision encoding rules. Which rules shall be used to determine SMT uniquenes:
a) whole encoded asset_symbol_type is used to check it (so pair NAI, precision determines uniqueness)
b) part of encoded asset_symbol_type holding a NAI is only used. This disallows to create 2 SMTs having same NAI and different precision

mvandeberg:
1. A single account can create multiple SMTs each having a different NAI.

2. The NAI is stored in `smt_creation_operation` when the user creates the application. In this case, we assume the user is some application providing a GUI to an actual human. The NAI is retrieved through an API call, `get_next_smt_identifier`.

`get_next_smt_identifier` takes a single argument, the desired precision of the SMT and returns a list of available NAIs. One of those NAIs is chosen and used in the operation. In #1633 I proposed creating a method called, `claim_smt_identifier` that takes a reference to the database and the NAI. The method will return true on a successful claim and false otherwise. Assert the return is true. This only updates the list of potential NAIs, but it ensures that the requested NAI will not be used by another SMT. The evaluator still needs to create the SMT object with the correct identifier.

For purposes of testing, define a valid NAI and have `claim_smt_identifier` return true only on the valid NAI. That way you can create valid and invalid ops and have the helper method respond accordingly. It will be implemented in a different issue.

3. The helper methods that determine the pool of potential NAIs will handle NAI uniqueness. It is not something that this evaluator needs to be concerned with. If `claim_smt_identifier` returns true then the NAI was already determined to be unique. 

The NAI without the precision digit will be unique. The precision digit only exists to represent the entirety of the symbol.

Kiwonik:
[...]

PR #1804, opened 2017-11-27
Remaining implementation of issue 1554

Kiwonik:
Added dummy get_smt_next_identifier implementation that allows for early use of smt creation in dedicated testnet and completes issue #1554


theoreticalbts:
Looks reasonable.

PR #1806, opened 2017-11-27
Refactor database::open / database::reindex #1805

### LINKED ISSUE #1805: Refactor database open() and reindex() methods
We've added enough arguments to `open()` and `reindex()` that we're getting problems like [this](https://github.com/steemit/steem/pull/1785).

Let's use a `struct` for the arguments so they must be named, it should make potential issues like #1785 less likely in the future.

theoreticalbts:
Closes #1805. Closes #1785.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -99,12 +99,12 @@ database::~database()
    clear_pending();
 }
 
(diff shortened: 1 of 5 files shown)


mvandeberg:
Looks reasonable

PR #1809, opened 2017-11-27
Remove remove_object #1138

### LINKED ISSUE #1138: Remove remove_object method from chainbase
This method is more properly implemented as an index extension.  Related:  #683 #854

theoreticalbts:
Closes #1138


mvandeberg:
This isn't breaking anything, so I don't see any reason why we shouldn't clean this up if we can.

PR #1812, opened 2017-11-28
Move sufficient funds checks to adjust_balance #1811

### LINKED ISSUE #1811: Move sufficient balance checks to adjust_balance
In a number of places we have checks that require sufficient funds for operations that can (with the account holder's permission) remove STEEM / SBD from an account (for example, transfer, creating a market order, vesting, escrow).

For SMT's we will be adding a lot of new operations that can remove STEEM / SBD from an account.  Instead of repeating these checks on each operation, let's move them to `adjust_balance()`, so newly created operations are checked for sufficient funds *by default* (if they use `adjust_balance()` instead of directly editing balance objects).

If we don't require the programmer to create sufficient funds checks manually for each operation, we reduce the possibility they will be incorrectly omitted for any operation (which would obviously be an enormous problem if exploited, it would allow anyone to print STEEM).

This is obviously very important to get right, so I'll tag the PR for review by two people.  Please review this one very carefully.

theoreticalbts:
Closes #1811

Tagging this for the SMT project because that's the reason I was looking into this, but really it's actually just core functionality.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -498,7 +498,6 @@ void database::pay_fee( const account_object& account, asset fee )
    if( fee.amount == 0 )
       return;
 
-   FC_ASSERT( account.balance >= fee );
    adjust_balance( account, -fee );
    adjust_supply( -fee );
 }
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -498,6 +498,7 @@ void database::pay_fee( const account_object& account, asset fee )
    if( fee.amount == 0 )
       return;
 
+   FC_ASSERT( account.balance >= fee );
    adjust_balance( account, -fee );
    adjust_supply( -fee );
 }
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -3520,7 +3520,6 @@ void database::adjust_reward_balance( const account_object& a, const asset& delt
 
 void database::adjust_supply( const asset& delta, bool adjust_vesting )
 {
-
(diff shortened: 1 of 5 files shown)


theoreticalbts:
Rebased on current develop, added fix for #1825.  This PR now closes #1825 .

mvandeberg:
Are we sure that all operations modify balances via `adjust_balance`?

If that is not the case this fix could inadvertently fork the blockchain. (If a fork occurred it would be the result of initially buggy code)

theoreticalbts:
> Are we sure that all operations modify balances via adjust_balance?

No, but all the places that I removed checks subsequently call `adjust_balance()`.

> If that is not the case this fix could inadvertently fork the blockchain. (If a fork occurred it would be the result of initially buggy code)

The only case in which it could fork the blockchain is if someone is actively exploiting a bug that allows negative balances.  I suppose I could gate the `adjust_balance()` changes by HF (of course I would also have to re-add the checks I removed).

theoreticalbts:
Perhaps this is more to your liking?  I made the following changes:

- Made the checks in `adjust_balance()` conditional on HF20
- Re-added the old checks in a separate commit (so the commit which removed them is easily accessible to be re-applied after HF20 occurs)
- Added a similar check to `adjust_supply()`

If you view the full diff for the PR, it should be much easier to review than the individual commits, since two of the commits cancel each other out.

mvandeberg:
Yeah, let's do this. I am pretty certain we would be safe with the original PR, but this guarantees we won't have problems. Especially considering the original issue was in the test fixture, not live code.

Issue #1642, opened 2017-10-12
Implement SMT balance objects


theoreticalbts:
SMT's need balances.  Each balance should be an object which maps `(account_name, symbol)` to an amount.  The [BitShares version](https://github.com/cryptonomex/graphene/blob/master/libraries/chain/include/graphene/chain/balance_object.hpp) is basically the same, excluding some fields related to BitShares vesting which are largely irrelevant to Steem.

According to usecontext needs defined in #1682, following definition is proposed:
```
enum account_balance_target : short
{
   REGULAR,
   SAVINGS
};
template <enum account_balance_target BalanceTarget>
class account_balance_object
{
   account_name_type owner;
   asset balance;
   account_balance_target balanceTarget() const
   {
      return BalanceTarget;
   }
};
```

This object will need to be indexed by `(owner, balance.symbol)`.  This will need to use accessor based indexing with `const_mem_fun` to extract `balance.symbol`

- [x] Define class for account_balance_object, take care for correct constructor `template <class Constructor, class Allocator> account_balance_object(Constructor&&, allocator< Allocator >)`

- [x] Define a method which retrieves asset_symbol_type from embedded balance - be useful for ordering op as also for further usage

- [x] Define an external index type and add its instance to chain::database: `void database::initialize_indexes()`

Issue #1682, opened 2017-10-23
Implement SMT transfers


theoreticalbts:
The existing `transfer_operation` must be extended to support SMT's.

- [x] Allow SMT usage to `void transfer_operation::validate() const`

- [x] Extend `database::get_balance` to read balance from correct place (see #1642)

- [x] Extend `database::adjust_balance`

vogel76:
@youkaicountry Does `escrow_transfer_operation` shall also support SMT transfers ?

youkaicountry:
@vogel76 We currently have no plans to extend SMT support to `escrow_transfer_operation`. It's pretty much unused, without even UI support.

vogel76:
@youkaicountry Does SMT assets shall be also stored for given account on:
- savings_balance,
- reward_steem_balance
?

youkaicountry:
@vogel76 There will be no `savings_balance` for SMTs.
`reward_steem_balance` will have an SMT equivalent, there will be an issue created for it in the future.
This issue is only related to extending `transfer_operation` to support SMTs.
With the minimum amount of infrastructure possible, the operaton should be able to accept SMTs for transfer, and then transfer the balance."

vogel76:
@youkaicountry This is very important information that `reward_steem_balance` will hold SMTs too, since it will impact #1642 - we need to be able to choose correct index to store/retrieve assets for given account or appropriately define index key to distinguish use context (regular-balance, rewards-balance). Any proposal for such solution ?

theoreticalbts:
An account has two fields called `balance` and `reward_steem_balance`.  They both represent STEEM owned by the account, but they are different "balance types".

We have two basic options:

# Option 1: Balance types in separate objects

- The two types go in different objects, for example `account_balance_object` and `reward_balance_object`.
- Each object type is indexed by `(account, symbol)`.

# Option 2: Balance types distinguished by enum

- Both balance types are represented by `balance_object`
- Add a new integer (or `enum`) field to `balance_object` identifying which balance type it is
- The `balance_object` is indexed by `(account, symbol, balance_type)`

There are competing long-term concerns here.  Option 1 is a better fit for modular blockchain, but Option 2 allows code to be deduplicated and allows generic API's for working with balances.

[...]

PR #1813, opened 2017-11-28
1642 implement smt balance objects

vogel76:
Tests passed, things done as described in issue #1642

### DIFF
--- a/libraries/chain/include/steem/chain/smt_objects/account_balance_object.hpp
+++ b/libraries/chain/include/steem/chain/smt_objects/account_balance_object.hpp
@@ -1,0 +1,91 @@
+#pragma once
+
+#include <steem/chain/steem_object_types.hpp>
+#include <steem/protocol/smt_operations.hpp>
+
+#ifdef STEEM_ENABLE_SMT
+
+namespace steem { namespace chain {
+
+template <enum account_balance_target BalanceTarget, enum object_type ObjectType>
+class account_balance_object : public object< ObjectType, account_balance_object<BalanceTarget, ObjectType> >
+{
+   account_balance_object() = delete;
+
+public:
+   typedef account_balance_object<BalanceTarget, ObjectType>   _this_type;
+
+   template <class Constructor, class Allocator>
+   account_balance_object(Constructor&&, allocator< Allocator >)
+   {
+      c( *this );
+   }
+
+   // id_type is actually oid<account_balance_object<BalanceTarget, ObjectType> >
+   typename object< ObjectType, _this_type >::id_type id;
+   account_name_type                                  owner = "@@@@@";
+   asset                                              balance;
+
+   account_balance_target balanceTarget() const
+   {
+      return BalanceTarget;
+   }
+
+   asset_symbol_type get_symbol() const
+   {
+      return balance.symbol;
+   }
+};
+
+struct by_owned_symbol;
+
+typedef multi_index_container <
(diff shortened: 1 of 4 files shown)


libraries/chain/include/steem/chain/smt_objects/account_balance_object.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace steem { namespace chain {
+
+/**
+ * Class responsible for holding non STEEM balance (SMT one) for given account.
+ * Since multiple kinds of balances are possible (atm regular and savings) this
+ * class is parameterized to distinguish their types. 
+ */
+template <enum object_type ObjectType>
+class account_balance_object : public object< ObjectType, account_balance_object<ObjectType> >
```
mvandeberg: I would consider combining the normal balance and savings balance in the same object. It may be a little more space, but it probably makes the implementation a bit easier.

libraries/chain/include/steem/chain/smt_objects/account_balance_object.hpp
```diff
@@ -0,0 +1,90 @@
...
+   template <class Constructor, class Allocator>
+   account_balance_object(Constructor&&, allocator< Allocator >)
+   {
+      c( *this );
+   }
+
+   // id_type is actually oid<account_balance_object<ObjectType> >
+   id_type             id;
+   /// Name of the account, the balance is held for.
+   account_name_type   owner = "@@@@@";
```
mvandeberg: I am not sure why we are setting the default value for account name type when the default `account_name_type()` is already an invalid account name.

youkaicountry: Looks like this happens in other places, we can file an issue to fix it.

mvandeberg: It isn't wrong, per se, but is not needed and does not match other places where account names are present in objects. I think this was picked up because some of our out of source documentation uses it as a place holder for an arbitrary account.

theoreticalbts:
I envisioned two possible ways of implementing this, as discussed in a comment on #1682.

What's implemented here is a third alternative I never considered.

It basically looks "from the outside" like Option 1 (Balance Types in Separate Objects) if you just use the `typedef` types `account_regular_balance_object` and `account_savings_balance_object` but "from the inside" it has some of the benefits of Option 2 (Balance Types Distinguished by Enum).  But since it doesn't combine the two object types into a single "table" or add an extra field to each object, it should be slightly better performance than Option 2 (since we have smaller tables of smaller objects).

libraries/chain/include/steem/chain/smt_objects/account_balance_object.hpp
vogel76: This part can be performance (especially memory consumption) sensitive since we are adding separate index for getting basic account data as its balance. Having limited mem. consumption and  make index choosing simply by compile time type-system is best what we can achieve here. The question is how much possible is to have account(s) without SMT savings...

mvandeberg:
We can eventually add an enum for reward balances. I've come around to the implementation.

theoreticalbts:
Sorry I never got back to this yesterday.  The implementation works, so let's use it, at least for now.

mvandeberg:
That is why I merged it yesterday ;)

Issue #409, opened 2016-09-15
Potential unexpected behavior while calculating median feed?


abitmore:
While calculating median feed, need to compare feed prices to each other (code [here](https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/database.cpp#L2910-L2912),[here](https://github.com/steemit/steem/blob/v0.14.1/libraries/chain/database.cpp#L2923)).

Format of feed prices can be either `{base:"xxx STEEM", quote:"yyy SBD"}` or `{base:"xxx SBD", quote:"yyy STEEM"}`, both are valid. Currently, most if not all witnesses are using the latter format. However, it's possible that some will use the other format in the future, which will cause unintended behavior while comparing due to [the comparison code](https://github.com/steemit/steem/blob/develop/libraries/chain/protocol/asset.cpp#L93-L96):

```
      bool operator < ( const price& a, const price& b )
      {
         if( a.base.symbol < b.base.symbol ) return true;
         if( a.base.symbol > b.base.symbol ) return false;
         if( a.quote.symbol < b.quote.symbol ) return true;
         if( a.quote.symbol > b.quote.symbol ) return false;
```

It's not a big deal as long as majority of active witnesses/miners use same format.

mvandeberg:
I believe there was originally a restriction on base and quote symbols for the price feed that was inadvertently removed. It would be easy enough to add the requirement back in.

mvandeberg:
This was fixed in hardfork 14 [here](https://github.com/steemit/steem/blob/master/libraries/chain/steem_evaluator.cpp#L79-L82).

mvandeberg:
Don't know what I was thinking when I closed this issue, but the referenced code does not fix the problem.

mvandeberg:
Will be fixed in #1620

mvandeberg:
HF 20 will require the base to be SBD and the quote to be STEEM.

mvandeberg:
This is from a quick check that most witnesses publish their price feed in this format. The change should impact the fewest number of people possible.

Issue #320, opened 2016-08-25
Increase STEEMIT_IRREVERSIBLE_THRESHOLD


theoreticalbts:
As part of #319 I checked the actual value of `STEEMIT_IRREVERSIBLE_THRESHOLD`. It is 51%.   As noted in #319, increasing `STEEMIT_IRREVERSIBLE_THRESHOLD` is necessary to safeguard against misbehaving witnesses (either due to bug, misconfiguration or malice).  I propose increasing this number to 80%.

Transactions will take slightly longer to get irreversible status, but increase safety for e.g. exchanges.  If #319 is implemented, this will not matter too much, as this should improve operating latency to 2-3 blocks when at least STEEMIT_IRREVERSIBLE_THRESHOLD witnesses are up.

bytemaster:
67% is the threshold indicated by byzantine generals

theoreticalbts:
Citation for 67% number?

theoreticalbts:
The intention here is for the blockchain to continue as long as 14 of the 19 elected witnesses are signing off.  The numbers need to be set so that this happens.

theoreticalbts:
14 / 19 elected witnesses signing off -> 16 / 21 of slots signing off; i.e. we tolerate at most 5 failures.  Considering a situation like this:

    0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20
    1  1  1  1  1  1  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2

we need to set last irreversible block to 1, but if the 5 slot changes to 2 then we should set it to 2.  So the critical slot is 5, calculated as `offset` by [this line](https://github.com/steemit/steem/blob/6500bb65eb6282866f3c6f356cbcbe09fd03c7bf/libraries/chain/database.cpp#L3196).  This means that, in order to implement the 5-failures-ok, stop-for-6 intent, we need to set STEEMIT_IRREVERSIBLE_THRESHOLD between 71.42% and 76.19% (inclusive); setting to 75% is a nice round number in this range.

theoreticalbts:
Merged.

PR #1818, opened 2017-11-28
1620 flexible witness parameters

### LINKED ISSUE #1620: Flexible witness parameter operation
SMT's specify some additional witness parameters.  Rather than add to the existing witness update operation, which requires the user to specify values, it makes more sense to have a `steem_set_runtime_parameter` operation, similar to the SMT version, which allows witnesses to set the desired parameter values without being required to set old parameters.

In particular, this will give us a path to completely deprecate the old operation, which is important because its validation checks are tricky and non-rigorous since they need to support broken past operations which made it into the blockchain in ancient times.

mvandeberg:
Closes #1620 #409


theoreticalbts:
This operation should require the active key, not the block signing key:

- Existing similar operations (`witness_update_operation`, `feed_publish_operation`) require the active key
- Including the block signing key requires either a substantial refactoring of the authority framework, or including the block signing key in the op.

mvandeberg:
It is using the signing due to witness requests in #320

I have discussed it directly with witnesses and have received incredible support for using the signing key on this operation. The authority structure for this operation is done correctly given our infrastructure.

theoreticalbts:
I don't really like it.  My instinct is to say "just make the operation require the active key."  But I'm willing to compromise, so let's think about how this should work.

Ideally you shouldn't need to include the key in the operation, the blockchain should "just know" witnesses' block signing keys so the key doesn't need to be included in the op, similar to how for e.g. a transfer op it "just knows" the user's active key, and the active key doesn't need to be included in the op.

But that would require refactoring the authority infrastructure [1].  While I would love to do that, I don't think we have enough free developer time to do that in time for HF20.

So we need a hack to make it work with the current design, that involves putting the key somewhere in the op.  The current patch puts the key in a field called `current_signing_key`.

Let's use a slightly different hack.  Instead of a field called `current_signing_key`, add an `extensions` field [2], and put `current_signing_key` in the extension, and enforce the following constraints:

- Statelessly require the extension to be present or the op fails `vaildate()`.
- Statelessly require signing by the key specified in the extension
- Statefully enforce the key specified in the extension is the specified witness's current signing key

It's basically just a small refactoring of the current code, to convert a field to an extension.  The strength of putting it in an extension is that if we (at a much later date) upgrade our infrastructure so that specifying the block signing key is unnecessary and it "just knows," we can remove the extension and have smaller ops.

[1] I started to do that in the branch for #68, but that code will require some more work and testing to get to the point where we can use its new infrastructure for consensus.

[2] We've agreed all new ops should have an extensions field regardless of whether the initial implementation provides any extensions (usually they don't, because for a brand-new op, if you discover needed functionality you can just add a field since there's no backward compat concern.)

theoreticalbts:
AFAICT nowhere does this code check that `current_signing_key` is equal to the witness's current block production key.  It should be enforced in the code and checked in a unit test, that `eve` can't simply set `owner = alice` and `current_signing_key = eve_public_key` and then modify `alice`'s witness parameters...

Also maybe we should call this something shorter, `signing_key` or even just `k`, since it is going to be included in every op, every byte we save here makes a difference to blockchain size.

theoreticalbts:
In the case where `get_required_authorities()` is supposed to return an unsatisfiable authority, we should require signing by `STEEM_NULL_ACCOUNT`.  The reason `STEEM_NULL_ACCOUNT` exists is to be the canonical unsatisfiable authority.

theoreticalbts:
I see the test doesn't allow unknown keys.  If that is the case, we might as well use `static_variant` which is more idiomatic and simplifies the code.

I was thinking the entire purpose of setting it up as a `map< string, vector<char> >` was to allow older clients to ignore certain witness parameters.

For example, let's say we add a new parameter whose only effect is to disallow a certain set of transactions, for example, a new witness parameter that's a dust threshold for transfers called `transfer_dust_threshold`.  With your code, `transfer_dust_threshold` would have to be implemented in a hardfork, because the unrecognized name `transfer_dust_threshold` would trigger an exception (this is explicitly tested in the unit tests).

However, the architecture of having `flat_map< string, vector<char> >` allows clients that aren't aware of the `transfer_dust_threshold` parameter to interoperate.  So `transfer_dust_threshold` can be implemented as a witnesses-only upgrade, other nodes don't have to upgrade.

The reason it has to be a `flat_map< string, vector<char> >` to allow this interoperability is that the `vector<char>` wrapper allows old clients to have a limited understanding of parameters created in the future (!!!) -- specifically, the `vector<char>` serialization tells the length (number of `char`).  If we tried to use a `static_variant`, the old version has to throw when it encounters an unknown tag, because it doesn't know how long the unknown tag's serialization is, so there's no way it can parse the rest of the ops in the transaction.

theoreticalbts:
So TLDR version of the above comment, we have two choices:

- (a) Alter the test to enforce that `extranious_property` [1] *does not* throw an exception [2].
- (b) Use `static_variant` instead of `flat_map<string, vector<char> >`.

Option (a) may sometimes allow us to implement future parameters as witness-only upgrades.  Option (b) is more similar to how we handle this situation in other parts of Steem, and is more space efficient (in terms of blockchain size).

[1] Should be `extraneous_property`, spelling.
[2] Of course you need to also change the consensus code to pass the new test.

theoreticalbts:
Which option to choose?  Both options have merits, but I'm leaning toward option (b) right now.

I'm thinking maybe we should just make it a `static_variant`, and if (in the future) we want to create a witnesses-only upgrade to implement witness parameters whose sole effect is to forbid certain transactions, we can implement those parameters using a custom op interpreted by the witness plugin.

mvandeberg:
I actually think (a) is the better choice.

A `static_variant` works best if we have distinct types. We already have two fields for `public_key_type`. Which means that in order to differentiate them, we need a wrapper. It is more space efficient, but adds development complexity.

I think the idea of a witness only update to add parameters makes sense. I can see this being useful to set soft parameters related to bandwidth without needing a custom op for those specific parameters.

theoreticalbts:
OK.  I don't have strong feelings about (b), and (a) is probably less code to write at this point.

theoreticalbts:
OK, looks like this is ready to go.

PR #1826, opened 2017-11-29
Change assert() to FC_ASSERT(), add exceptions #1815

### LINKED ISSUE #1815: Change assert() in database::match() to FC_ASSERT()
There are lots of `assert()` statements [in database::match()](https://github.com/steemit/steem/blob/e58333c8ea7b33c7399484aa7a8e6913a766a666/libraries/chain/database.cpp#L3219-L3273).  This ticket proposes changing them to `FC_ASSERT()`.  (The difference is that `assert()` is only triggered in debug builds.)  Some things to keep in mind about this kind of change:

- (a) In general, recently the core devs' policy has been that it is acceptable to convert `assert()` to `FC_ASSERT()` statements for conditions that should never happen.  The slight performance penalty is worth extra checking, defense in depth against potential bugs or exploits.
- (b) Newly-added asserts must "fail safe" with regard to block production, `FC_ASSERT()` in per-block processing could shut down the blockchain, so we can't put any there.
- (c) Newly-added asserts must "fail safe" with regard to consensus changes, any newly added `FC_ASSERT()` must be gated by a hardfork (which can be removed once the hardfork passes, assuming nothing that triggered the newly added `FC_ASSERT()` got into the chain).

Since SMT's may exercise more corner cases of the matching engine, we should tighten these assertions in preparation -- ideally in HF20.

The implementation strategy for adding new `FC_ASSERT()` is fairly straightforward:

- Change to `FC_ASSERT()`, replay, see if anything on-chain triggers assertions
- If something triggers, we need to further analyze the issue
- If nothing triggers, move `FC_ASSERT()` to `if( has_hf_20 )` block
- Once HF20 passes, remove `if( has_hf_20 )` block, if we can replay then all is well (we can merge the removal of the `if( has_hf_20 )` check and do `FC_ASSERT()` unconditionally)

theoreticalbts:
Closes #1815


theoreticalbts:
Rebased, only conflict was a new `0_20` hardfork constant was added to `develop` since this branch branched.

Issue #1828, opened 2017-11-29
Benchmark Read/Write Lock Performance Impact


mvandeberg:
While investigating possible solutions for #1827, I was looking at different Boost lock types. There is no sharable recursive mutex. I found some interesting articles talking about the pros and cons of such approaches. The sharable mutex has been at the center of several debates and is being added to the C++17 standard, but has been rejected for several revisions of the standard library.

A couple of interesting articles:

http://joeduffyblog.com/2009/02/11/readerwriter-locks-and-their-lack-of-applicability-to-finegrained-synchronization/
http://permalink.gmane.org/gmane.comp.lib.boost.devel/211180

We should investigate more closely the performance impact of how we are using locks and if there are alternative synchronization mechanisms that have better performance.

PR #1830, opened 2017-11-30
Remove API Lock Recursion

### LINKED ISSUE #1827: Remove Appbase Lock Recursion
While investigating #1823, I realized that we needed to call `witness_api.get_account_bandwidth` in a loop. The accessible version of the API call acquires a read lock, but the read lock has already been acquired by `get_accounts` at this time. Chainbase uses an `interprocess_sharable_mutex`

According to the [Boost docs](http://www.boost.org/doc/libs/1_60_0/doc/html/interprocess/synchronization_mechanisms.html), `interprocess_sharable_mutex` is

> A non-recursive, anonymous sharable mutex that can be placed in shared memory or memory mapped files.

And earlier in the documentation it is stated that

> Non-recursive mutexes can't be locked several times by the same thread. If a mutex is locked twice by a thread, the result is undefined, it might throw an error or the thread could be blocked forever.

Obviously, this is bad and needs to change.

All of our API methods are already declared and defined via macros. I propose creating a read and write wrapper macro that handles the creating the lock. This new method signature would contain a boolean default initialized to false for whether to use a lock or not. The jsonrpc plugin will explicitly call this with true and internal calls (plugin to plugin) or APIs will implicitly avoid locking assuming that the calling thread is already locked. There will be a third macro for lockless APIs which define this field but mark at as unused. That is because the jsonrpc plugin will call all APIs with true, but not all APIs need to be locked (some are purely utility). This has the added benefit that all APIs at compile time need to be explicitly marked as read/write/lockless.

mvandeberg:
Closes #1827

### DIFF
--- a/libraries/plugins/apis/account_by_key_api/account_by_key_api.cpp
+++ b/libraries/plugins/apis/account_by_key_api/account_by_key_api.cpp
@@ -50,12 +50,6 @@ account_by_key_api::account_by_key_api(): my( new detail::account_by_key_api_imp
 
 account_by_key_api::~account_by_key_api() {}
 
-get_key_references_return account_by_key_api::get_key_references( const get_key_references_args& args )
-{
-   return my->_db.with_read_lock( [&]()
-   {
-      return my->get_key_references( args );
-   });
-}
+DEFINE_READ_APIS( account_by_key_api, (get_key_references) )
 
 } } } // steem::plugins::account_by_key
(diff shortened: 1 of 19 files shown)


theoreticalbts:
Looks reasonable.  I'm not sure about the macro names, or about using a `bool` parameter.  But that's a separate issue, the architecture's OK and the code should work.  Let's merge it.

mvandeberg:
The correct solution to this is to come up with a better locking solution that can handle an API calling another API in a well defined manner. See #1828

PR #1832, opened 2017-11-30
Implemented support for SMT transfers

Kiwonik:
Modified `adjust_balance`, `adjust_reward_balance` and `get_balance`. Refactored `savings` to `rewards`. Added simple test.  #1682

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -3405,7 +3405,36 @@ void database::clear_expired_delegations()
       itr = delegations_by_exp.begin();
    }
 }
+#ifdef STEEM_ENABLE_SMT
+template< typename smt_balance_object_type >
+void database::adjust_smt_balance( const account_object& a, const asset& delta, bool check_balance )
+{
+   //elog( "${a} ${b} ${c}", ("a", a.name) ("b", delta.amount) ("c", delta.symbol));
+   const smt_balance_object_type* bo = 
+      find< smt_balance_object_type, by_owned_symbol >( boost::make_tuple(a.name, delta.symbol) );
+   if( bo == nullptr )
+   {
+      const auto& new_balance = create< smt_balance_object_type >( [&]( smt_balance_object_type& smt_balance )
+      {
+         smt_balance.owner = a.name;
+         smt_balance.balance = delta;
+      } );
+      bo = &new_balance;
+   }
+   else
+   {
+      modify( *bo, [&]( smt_balance_object_type& smt_balance )
+      {
+         smt_balance.balance += delta;
+      } );
+   }
 
+   if( check_balance )
+   {
+      FC_ASSERT( bo->amount.value >= 0, "Insufficient SMT ${smt} funds", ("smt", delta.symbol) );
+   }
+}
+#endif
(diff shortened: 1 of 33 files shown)


youkaicountry:
Looks like there are merge conflicts though.

Kiwonik:
@youkaicountry I rebased the branch on develop's HEAD. Ready to merge again.

Kiwonik:
@vogel76 The code you commented is used by `adjust_balance` (modify) method, not by `get_balance` (check).

libraries/chain/database.cpp
```diff
@@ -3404,15 +3405,58 @@ void database::clear_expired_delegations()
       itr = delegations_by_exp.begin();
    }
 }
+#ifdef STEEM_ENABLE_SMT
+template< typename smt_balance_object_type >
+void database::adjust_smt_balance( const account_object& a, const asset& delta, bool check_balance )
```
mvandeberg: Only the account name is used in this method, so let's just pass the account name.

libraries/chain/database.cpp
```diff
@@ -3404,15 +3405,58 @@ void database::clear_expired_delegations()
...
+   {
+      modify( *bo, [&]( smt_balance_object_type& smt_balance )
+      {
+         smt_balance.balance += delta;
+      } );
+   }
 
+   if( check_balance )
+   {
+      FC_ASSERT( bo->balance.amount.value >= 0, "Insufficient SMT ${smt} funds", ("smt", delta.symbol) );
```
mvandeberg: This check is incorrect and should always be done, regardless of the existence of the account.

What if the delta is negative? Then a balance object with a negative balance will be created.

libraries/chain/database.cpp
```diff
@@ -3404,15 +3405,58 @@ void database::clear_expired_delegations()
...
+   }
+}
```
mvandeberg: This method should probably automatically remove balance objects if the resulting balance is 0.

libraries/chain/database.cpp
mvandeberg: I see it is passed from `adjust_balance`. Because SMTs are new, we don't need to gate it with a hardfork check.

mvandeberg:
This is challenging to review because `develop` was merged into the feature branch rather than the feature branch being rebased on `develop`. All of the commits that were added to `develop` since the base of the branch are now showing in the diff.

Kiwonik:
Indeed they are, even though I **did** do `git rebase develop` with the feature branch and not merging :(
I'll make another cherry-picked branch and add requested changes to it.

Kiwonik:
[PR 1842](https://github.com/steemit/steem/pull/1842) replaces this one.

youkaicountry:
Closing in favor of #1842 replacing this.

youkaicountry:
@mvandeberg @theoreticalbts What is the policy on closed PR branches. Should I delete it as I close the PR?

mvandeberg:
Only delete it if you know we don't need the branch. I periodically clean up old branches, so if one slips through, it will get cleaned up eventually.

PR #1840, opened 2017-12-04
Fix wrong amount of withdrawn VESTS

pmconrad:
When vesting withdrawals are processed and withdraw routes have been set for that account, several virtual operations are created - one for each withdraw route, and one for the rest. In that final vop for the rest, the amount of VESTS is wrong: it contains the full withdrawal rate and does not take into account that part of the rate has already been paid out through the withdraw routes.

Fortunately, the actual withdrawal uses the correct amount. The bug affects only the virtual operation.

This PR contains a simple fix.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -1471,7 +1471,7 @@ void database::process_vesting_withdrawals()
       if( to_withdraw > 0 )
          adjust_proxied_witness_votes( from_account, -to_withdraw );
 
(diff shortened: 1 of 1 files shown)


theoreticalbts:
I agree with this change.  @mvandeberg , since you wrote the vesting withdraw code, could you take a look at this PR?  If you approve, make your own instance of this PR so it goes through CI, and I'll approve it.

mvandeberg:
Yeah, I think this is reasonable.

PR #1841, opened 2017-12-04
Templated Authority Verification

mvandeberg:
Related to #1674 
Each commit was originally pushed individually as to prove successful CI on each change.

These are the consensus code changes required for #1674. The code in this branch should be equivalent to code in `develop` and should not change consensus behavior in any way. Reviewing this code should be focused on this and design of the refactor.

### DIFF
--- a/libraries/protocol/include/steem/protocol/operation_util.hpp
+++ b/libraries/protocol/include/steem/protocol/operation_util.hpp
@@ -10,6 +10,42 @@
 #include <string>
 #include <vector>
 
+namespace steem { namespace protocol {
+
+struct get_required_auth_visitor
+{
+   typedef void result_type;
+
+   flat_set< account_name_type >&        active;
+   flat_set< account_name_type >&        owner;
+   flat_set< account_name_type >&        posting;
+   std::vector< authority >&  other;
+
+   get_required_auth_visitor(
+         flat_set< account_name_type >& a,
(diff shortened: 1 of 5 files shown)


theoreticalbts:
Looks good.  Minor nitpicks:

- Clearer naming:  Maybe we should rename `template< typename AuthContainerType >` to `template< typename OperationType >`
- Clearer naming:  Maybe we should rename `template< typename ...Ts >` to `template< typename OperationTypes >`
- Clearer naming:  Maybe we should rename `template< typename T >` to `template< typename OperationType >`
- Code organization:  We should probably put visitor structs and template definitions in separate files from declarations.  Often, headers only need the declaration, and the body only needs to be included in a single compilation unit (`.cpp` file).  If we avoid unnecessary inclusion of template code, hopefully it will result in faster compile times and less compiler memory usage.  This was my original reason for having separate files `operation_util_impl.hpp` and `operation_util.hpp`.

None of these nitpicks are actual bugs or blockers.  The code is functionally correct, this sequence of commits clarifies what is being done and gives confidence that the code is equivalent to the existing code.  I don't know that it's a wise use of resources to spend time right now on the very minor issues above.  So I'm approving this PR.

mvandeberg:
I was trying to not name thing with operations because the API that will use this is not an operation. It is a conceptually higher level type that defines required authorities.

PR #1842, opened 2017-12-04
2nd take to merge SMT transfers code

Kiwonik:
As promised in [PR 1832](https://github.com/steemit/steem/pull/1832) I cherry picked the commits, put in a clean branch and added changes requested in that PR. Issue #1682

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -3407,31 +3407,44 @@ void database::clear_expired_delegations()
 }
 #ifdef STEEM_ENABLE_SMT
 template< typename smt_balance_object_type >
-void database::adjust_smt_balance( const account_object& a, const asset& delta, bool check_balance )
+void database::adjust_smt_balance( const account_name_type& a, const asset& delta )
 {
    //elog( "${a} ${b} ${c}", ("a", a.name) ("b", delta.amount) ("c", delta.symbol));
    const smt_balance_object_type* bo = 
-      find< smt_balance_object_type, by_owned_symbol >( boost::make_tuple(a.name, delta.symbol) );
+      find< smt_balance_object_type, by_owned_symbol >( boost::make_tuple(a, delta.symbol) );
+   // Note that SMT related code, being post-20-hf needs no hf-guard to do balance checks.
    if( bo == nullptr )
    {
+      // No balance object related to the SMT means '0' balance. Check delta to avoid creation of negative balance.
+      FC_ASSERT( delta.amount.value >= 0, "Insufficient SMT ${smt} funds", ("smt", delta.symbol) );
+      // No need to create object with '0' balance (see comment above).
+      if( delta.amount.value == 0 )
+         return;
+
       const auto& new_balance = create< smt_balance_object_type >( [&]( smt_balance_object_type& smt_balance )
       {
-         smt_balance.owner = a.name;
+         smt_balance.owner = a;
          smt_balance.balance = delta;
       } );
       bo = &new_balance;
    }
    else
    {
-      modify( *bo, [&]( smt_balance_object_type& smt_balance )
+      asset result = bo->balance + delta;
+      // Check result to avoid negative balance storing.
+      FC_ASSERT( result.amount.value >= 0, "Insufficient SMT ${smt} funds", ("smt", delta.symbol) );
+      // Zero balance is the same as non object balance at all.
+      if( result.amount.value == 0 )
       {
-         smt_balance.balance += delta;
-      } );
-   }
-
-   if( check_balance )
-   {
-      FC_ASSERT( bo->amount.value >= 0, "Insufficient SMT ${smt} funds", ("smt", delta.symbol) );
+         remove( *bo );
+      }
+      else
+      {
+         modify( *bo, [&]( smt_balance_object_type& smt_balance )
+         {
+            smt_balance.balance = result;
+         } );
+      }
    }
 }
 #endif
@@ -3441,9 +3454,10 @@ void database::adjust_balance( const account_object& a, const asset& delta )
 
 #ifdef STEEM_ENABLE_SMT
    // No account object modification for SMT balance, hence separate handling here.
+   // Note that SMT related code, being post-20-hf needs no hf-guard to do balance checks.
    if( delta.symbol.space() == asset_symbol_type::smt_nai_space )
    {
-      adjust_smt_balance< account_regular_balance_object >( a, delta, check_balance );
+      adjust_smt_balance< account_regular_balance_object >( a.name, delta );
       return;
    }   
 #endif
@@ -3559,9 +3573,10 @@ void database::adjust_reward_balance( const account_object& a, const asset& delt
 
 #ifdef STEEM_ENABLE_SMT
    // No account object modification for SMT balance, hence separate handling here.
+   // Note that SMT related code, being post-20-hf needs no hf-guard to do balance checks.
    if( delta.symbol.space() == asset_symbol_type::smt_nai_space )
    {
-      adjust_smt_balance< account_rewards_balance_object >( a, delta, check_balance );
+      adjust_smt_balance< account_rewards_balance_object >( a.name, delta );
       return;
    }   
 #endif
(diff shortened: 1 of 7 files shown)


theoreticalbts:
This is almost ready to merge, just a few minor issues:

- (a) `adjust_smt_balance()` is an internal helper method, it can't be called from outside `database.hpp` anyway.  So let's move it from a `database` member function to `database::adjust_smt_balance( a, delta )` to `adjust_smt_balance( db, a, delta )`.
- (b) It is a problem to use `ordered_non_unique` index in blockchain code [1].  It should be changed to `ordered_unique` or removed.
- (c) For the specific case of `by_account` index on `account_balance_object`, it should be removed, since any query which could be implemented with `by_account` could also be implemented with `by_owned_symbol`.
- (d) Rename `by_owned_symbol` to `by_owner_symbol` (index name should match field name).
- (e) `database_fixture::transfer_smt()` method looks identical to `database_fixture::transfer()` method.  Let's just have one method, `database_fixture::transfer()`.

[1] If you use a non-unique index, then it is perfectly okay so long as no consensus code iterates over the results.  However, historically it's occurred multiple times that an initial implementation of some feature is written with `ordered_non_unique` index, since at the time the index definition is created the implementor can't think of any query that requires iteration.  Then later it happens that the implementation is altered, improved or a new feature is implemented, but the index is never redefined.  This is very bad and can lead to hard-to-diagnose forks and node dropouts, as node behavior starts to depend on local conditions (shared memory size, set of plugins running, whether the node's replayed from genesis on its current run or has loaded state).  In the worst case (a majority of witnesses do things in a strange order) it can lead to an irreversible consensus state that relies on an undefined ordering (fortunately AFAIK this has never occurred on a production blockchain).

Kiwonik:
@theoreticalbts 
> (a) adjust_smt_balance() is an internal helper method, it can't be called from outside database.hpp anyway. So let's move it from a database member function to database::adjust_smt_balance( a, delta ) to adjust_smt_balance( db, a, delta ).

There are several private methods of `database` besides `adjust_smt_balance` and all of them can't be called from outside database class. What is so special about this one that it should be kicked out of the class? I've never heard of such practice.

Kiwonik:
@theoreticalbts
> (c) For the specific case of by_account index on account_balance_object, it should be removed, since any query which could be implemented with by_account could also be implemented with by_owned_symbol.

This specific index has been added to allow for easy listing of given account's SMT balances - a "wallet" of SMTs so to speak. When you implement it with by_owned_symbol it is my understanding that you get a loop instead. A loop by **every SMT ever created** to check whether a symbol/account balance exists. Is this how you want to implement it or am I missing something?

theoreticalbts:
> When you implement it with by_owned_symbol it is my understanding that you get a loop instead. A loop by every SMT ever created to check whether a symbol/account balance exists

No, you can iterate over a container in the order given by an index.  So if you want to iterate over all of Alice's SMT balances, you say something like:

```
account_name_type alice = "alice";
auto it = idx.lower_bound( alice, asset_id_type() );
while( (it != idx.end()) && (it->owner == alice) )
{
   // do something with *it
   ++it;
}
```

This loop will only process SMT's where Alice has a nonzero balance.  The [database_api plugin](https://github.com/steemit/steem/blob/138e32a43a763e964030a97e0cb37ad012265246/libraries/plugins/apis/database_api/database_api.cpp) allows a separate wallet program (or web-based UI) to access the iteration functionality over JSON-RPC.

If you want a more specific example, the [limit order object](https://github.com/steemit/steem/blob/138e32a43a763e964030a97e0cb37ad012265246/libraries/plugins/apis/database_api/database_api.cpp#L1256-L1290) is implemented very similar to how SMT balance objects will be implemented.  It calls a generic [iterate_results function](https://github.com/steemit/steem/blob/138e32a43a763e964030a97e0cb37ad012265246/libraries/plugins/apis/database_api/database_api.cpp#L71-L83) for the actual iteration, the rest is some very simple boilerplate (which might even be auto-generated sometime in the future).

Kiwonik:
@theoreticalbts 
> e) database_fixture::transfer_smt() method looks identical to database_fixture::transfer() method. Let's just have one method, database_fixture::transfer().

Are you aware that these two differ in interface, thus unifying them requires a change to all legacy calls of `database_fixture::transfer`?
For example 
`transfer( STEEM_INIT_MINER_NAME, accounts[i], 10000 );`
will be
`transfer( STEEM_INIT_MINER_NAME, accounts[i], asset( 10000 ) );`
or rather should be
 `transfer( STEEM_INIT_MINER_NAME, accounts[i], asset( 10000, STEEM_SYMBOL ) );`
because with the advent of SMTs the legacy constructor of asset that allows to provide only the amount may easily lead to accidental mistakes. My suggestion is that current
```
asset( share_type a = 0, asset_symbol_type id = STEEM_SYMBOL )
:amount(a),symbol(id){}
```
is replaced with
```
asset( share_type a, asset_symbol_type id )
:amount(a),symbol(id){}
asset()
:amount(0),symbol(STEEM_SYMBOL){}
```
This however seems to be outside the scope of this issue. Let's make it into separate issue and put transfer/transfer_smt unification in it too.

theoreticalbts:
> Are you aware that these two differ in interface, thus unifying them requires a change to all legacy calls of database_fixture::transfer?

No, I wasn't aware of that.

> the legacy constructor of asset that allows to provide only the amount may easily lead to accidental mistakes

Explicit is better than implicit.  I've made an issue #1846.

theoreticalbts:
As far as this issue goes, everything is ready to merge.  (a) is not important enough to block a merge or make another issue for right now, (e) has been moved to #1846.

vogel76:
@theoreticalbts @mvandeberg Regarding to unique/non_unique index definition: iteration order (on **orderded** index shall be always the same and it is determined by less operator defined for given index. Moreover, boost multi_index_container emulates std::multiset and it is explicitly stated in its documentation - please look here [ordered-index doc](http://www.boost.org/doc/libs/1_64_0/libs/multi_index/doc/reference/ord_indices.html#unique_non_unique):

> Except where noted or if the corresponding interface does not exist, ordered indices (both unique and non-unique) satisfy the C++ requirements for associative containers at [associative.reqmts] (supporting unique and equivalent keys, respectively.) Accordingly, validity of iterators and references to elements is preserved. We only provide descriptions of those types and operations that do not exactly conform to or are not mandated by the standard requirements.

Because of that it must satisfy requirements of std::multiset described in STD docs as also in C++ standard, where is explicitly said that:

> The order of the elements that compare equivalent is the order of insertion and does not change. (since C++11)

and from C++ standard:

> For multiset and multimap, insert, emplace, and erase preserve the relative ordering of equivalent elements.

Links below:
[multiset](http://en.cppreference.com/w/cpp/container/multiset) 
[C++ Std 26.2.6 Associative containers, Paragraph 4 ](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf)

Forcing use of unique indices (instead of non_unique) is very bad practice. It leads to artificial key definitions, just to satisfy uniqueness constraint of the index what negatively impacts both performance and memory consumption. Also it is very hard to determine which index really should be unique.
Moreover, lets consider the case when real unique index is needed: if it will be defined together with object-id as you suggest, it can lead to uniqueness violation, for example:
```
struct S {int id; a; b;};
index requirements:
unique(id)
unique(a)
``` 
if 'a' will be defined as [a, id] it will accept 2 objects having the same 'a' value but different 'id' what should be disallowed.

mvandeberg:
Insertion order is not guaranteed the same on all nodes when handling forks.

For example, two limit orders of the same price. Let's assume, for simplicity, we have only a sort order on price. Order A was placed before order B, so it should be iterated over, and thus filled, before B. If a block is sent that is on a minority fork that has A cancel their order and then the block is reversed, A is added back to the non-unique index. But now because it was inserted after B, the fill order is B A.

Now, back on the majority fork a limit order C is filled on the other side of the book that can satisfy A or B, but not both? Whose order is filled. On some nodes A and other B. The network has just forked and the reason why can be quite difficult to diagnose. That is why we always use unique indices.

PR #1849, opened 2017-12-06
Check reputation conditions when undoing a vote

### LINKED ISSUE #946: Unvoting affects reputation score
When an account with 25 rep score showing on steemit.com is upvoting, it adds 0 rep; however, if it unvote later, it will harm the author's rep.

Related code is [here](https://github.com/steemit/steem/blob/v0.16.3/libraries/plugins/follow/follow_plugin.cpp#L96).

Similarly, while a downvote may not decrease an author's rep score, a following unvote will increase the author's rep score.

mvandeberg:
Closes #946

### DIFF
--- a/libraries/plugins/follow/follow_plugin.cpp
+++ b/libraries/plugins/follow/follow_plugin.cpp
@@ -69,15 +69,31 @@ struct pre_operation_visitor
 
          if( cv != cv_idx.end() )
          {
-            const auto& rep_idx = db.get_index< reputation_index >().indices().get< by_account >();
-            auto rep = rep_idx.find( op.author );
+            auto rep_delta = ( cv->rshares >> 6 );
 
-            if( rep != rep_idx.end() )
+            const auto& rep_idx = db.get_index< reputation_index >().indices().get< by_account >();
+            auto voter_rep = rep_idx.find( op.voter );
+            auto author_rep = rep_idx.find( op.author );
+
+            if( author_rep != rep_idx.end() )
             {
-               db.modify( *rep, [&]( reputation_object& r )
+               // Rule #1: Must have non-negative reputation to effect another user's reputation
+               if( voter_rep != rep_idx.end() && voter_rep->reputation < 0 ) return;
+
+               // Rule #2: If you are down voting another user, you must have more reputation than them to impact their reputation
+               if( cv->rshares < 0 && !( voter_rep != rep_idx.end() && voter_rep->reputation > author_rep->reputation - rep_delta ) ) return;
+
+               if( rep_delta == author_rep->reputation )
                {
-                  r.reputation -= ( cv->rshares >> 6 ); // Shift away precision from vests. It is noise
-               });
+                  db.remove( *author_rep );
+               }
+               else
+               {
+                  db.modify( *author_rep, [&]( reputation_object& r )
+                  {
+                     r.reputation -= ( cv->rshares >> 6 ); // Shift away precision from vests. It is noise
+                  });
+               }
             }
          }
       }


theoreticalbts:
I'm not sure this actually solves the problem.  What we basically want to do is:

```
rep_delta = cv->rshares >> 6;

// on initial upvote
if( voter_rep > author_rep )
   author_rep += rep_delta;

// on remove upvote
author_rep -= rep_delta;
```

Then the patch is to basically change the logic to:

```
rep_delta = cv->rshares >> 6;

// on initial upvote
if( voter_rep > author_rep )
   author_rep += rep_delta;

// on remove upvote, undo the effects that were done on initial upvote
```

The problem is that `author_rep` and `voter_rep` could change between the initial upvote and the removal, so you can't just have the removal check `if( voter_rep > author_rep )`.  So I think this patch isn't sufficient, we need to create an object to forward the value that was subtracted to the removal logic:

```
// on initial upvote
if( voter_rep > author_rep )
   rep_delta = cv->rshares >> 6;
else
   rep_delta = 0;

author_rep += rep_delta;

const rep_delta_tracking_object& rdto = db.create( ...
{
   obj.rep_delta = rep_delta;
} );

// on remove upvote, undo the effects that were done on initial upvote
// retrieve the same rep_delta_tracking_object that was created in the upvote we're undoing
const rep_delta_tracking_object& rdto = db.get< ... >( ... );
author_rep -= rdto.rep_delta;
db.remove( rdto );
```

This is very rough code, not all `rdto` will end up in this path so we need to clean them up elsewhere.  Also there will likely be quite a performance / memory usage impact from essentially creating/deleting another object for every upvote.

Isn't it the case that the plugin is deprecated because `hive` re-implements reputation anyway?  If so, let's not spend any more time on the plugin.

mvandeberg:
From the original issue 

>Unless we want to have a comment vote object for every rep change as well (we don't) there will be one case that still can cause problems, but I think it is less likely to occur.

Hive is not handling reputation calculations and has no plans to do so. This part of the plugin will need to continue to be supported.

theoreticalbts:
> Hive is not handling reputation calculations and has no plans to do so. This part of the plugin will need to continue to be supported.

What if we defer updates to reputations until cashout, then read everyone's rshares all at once when cashout occurs?  Since we'd just be reading the final votes once at a single point in time, you'd have correct behavior in all cases without having to actually code explicit handling of changes over time of vote direction (due to undoing / redoing a vote) or changing vote value (due to e.g. VESTS increase / decrease).

We might need a separate signal that allows the plugin to hook an event for "all the votes are in and we're about to cashout this post, but it hasn't been done yet so you can still read the vote objects."  Actually, looking at the code, `comment_payout_update_operation` might occur in the right place for this.

mvandeberg:
We could do that for past votes, but it is important for when newer accounts that begin spamming inappropriate content that their reputation reflects that immediately.

mvandeberg:
> Ideally, I'd suggest creating a plugin from scratch that just tracks reputation, there's no good reason for this to live in the follow plugin.

That is what we will do, but we need to make the switch to hive first. Having to do that refactor, while supporting rep and follows, eats up dev time that gets us nothing.

theoreticalbts:
> we need to make the switch to hive first. Having to do that refactor, while supporting rep and follows, eats up dev time that gets us nothing.

Does that mean no further action on this ticket until Hive is deployed and running smoothly?

mvandeberg:
> Does that mean no further action on this ticket until Hive is deployed and running smoothly?

I think this patch is mergable as is. You made the requirement that we should create a new plugin, to which I responded that it does not make sense to do so until Hive is released. However, I do not think that making a new plugin is a requirement for this branch to be merged.

theoreticalbts:
This PR isn't perfect in all corner cases, but it's less broken than the current code.  So I'm going to go ahead and approve it.

Issue #1824, opened 2017-11-29
SMT Tests Failing on OS X


mvandeberg:
The following tests are failing on OS X to a SIGABRT.

smt_create_apply
setup_emissions_validate
setup_emissions_apply
set_setup_parameters_apply
runtime_parameters_apply

There is a commonality to the stack trace.

```
frame #0: 0x00007fffbe051d42 libsystem_kernel.dylib`__pthread_kill + 10
frame #1: 0x00007fffbe13f457 libsystem_pthread.dylib`pthread_kill + 90
frame #2: 0x00007fffbdfb74bb libsystem_c.dylib`__abort + 140
frame #3: 0x00007fffbdfb7d7e libsystem_c.dylib`__stack_chk_fail + 205
frame #4: 0x00000001007d3a7c chain_test`steem::protocol::asset_symbol_type::to_string(this=0x00007fff5fbf4c60) const at asset_symbol.hpp:0
frame #5: 0x00000001007d3951 chain_test`fc::to_variant(sym=0x00007fff5fbf4c60, v=0x00007fff5fbf48c8) at asset_symbol.hpp:171
frame #6: 0x00000001007d3928 chain_test`fc::variant::variant<steem::protocol::asset_symbol_type>(this=0x00007fff5fbf48c8, val=0x00007fff5fbf4c60) at variant.hpp:542
frame #7: 0x00000001007d38dd chain_test`fc::variant::variant<steem::protocol::asset_symbol_type>(this=0x00007fff5fbf48c8, val=0x00007fff5fbf4c60) at variant.hpp:540
```

It would appear on OS X there is some sort of stack violation in the NAI code.

This happens on release and debug builds, so it is not optimization related.

vogel76:
Didn't debug it but as I see in the code bug there is buffer overrun since STEEM_ASSET_SYMBOL_MAX_LENGTH points to 10 and you are writing to the index 11:
```
         static_assert( STEEM_ASSET_SYMBOL_MAX_LENGTH >= 10, "This code will overflow a short buffer" );
         uint32_t x = to_nai();
         buf[11] = '\0';
         buf[10] = ((x%10)+'0');  x /= 10;

```
I think better would be directly write to resized std::string, currently there is always made a copy...

mvandeberg:
That looks like it could be the problem.

```c++
std::string to_string()const
{
   char buf[ STEEM_ASSET_SYMBOL_MAX_LENGTH+1 ];
   to_string( buf );
[...]

PR #1850, opened 2017-12-07
Issue #1824 - eliminating asset::to_string buffer overruns

vogel76:
Related to issue #1824

### DIFF
--- a/libraries/protocol/asset.cpp
+++ b/libraries/protocol/asset.cpp
@@ -14,38 +14,105 @@ index : field
 
 namespace steem { namespace protocol {
 
-void asset_symbol_type::to_string( char* buf )const
+#define STEEM_ASSET_SYMBOL_MAX_LENGTH  12
+
+namespace
 {
-   uint64_t* p = (uint64_t*) buf;
-   static_assert( STEEM_ASSET_SYMBOL_MAX_LENGTH >= 7, "This code will overflow a short buffer" );
+
+template <typename T>
+class conversion_helper
+   {
+   public:
+      typedef std::array<char, std::numeric_limits<T>::digits10 + 2> buffer_t;
+      typedef typename buffer_t::const_iterator const_iterator;
+
+      /// Returns max. number of potential decimal digits to be produced from value of given type 
+      size_t maxDigits() const
+      {
+         return buffer.size();
+      }
+
+      std::pair<const_iterator, const_iterator> to_decimal_string(T& v, unsigned int forceDecimalDigits)
+      {
+         FC_ASSERT(forceDecimalDigits < maxDigits());
+
+         auto insertionPoint = buffer.rbegin();
+         size_t produced = 0;
+         do
+         {
+            char d = char( v%10 );
+            v /= 10;
+            *insertionPoint = '0' + d;
+
+            if(++produced < forceDecimalDigits)
+               ++insertionPoint;
+            else
+               break;
+         }
+         while(insertionPoint != buffer.rend());
+
+         return std::make_pair(insertionPoint.base() - 1, buffer.end());
+      }
+
+      /// Allows to convert given value to decimal string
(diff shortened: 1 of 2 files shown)


theoreticalbts:
Let's change `to_string(buffer)` override as follows:

- Take `string&` instead of `string*`
- Be `public` not `private`, this is a perfectly well-defined method that we may want to use
- Get rid of `struct asset;` and `friend struct asset;`, it is a little confusing to have `asset_symbol_type` "know" something about `asset`, simpler if the dependency arrow points one way from `asset` to `asset_symbol_type`

theoreticalbts:
- The digit for precision is a hex digit, not a decimal digit.  AFAICT this code doesn't function correctly for precision >= 10.
- Why have two helpers if `decimals()` is always a single digit?  Should just have one helper and pass in the number of decimals.
- Why two `to_decimal_string()` overrides?  Let's only have one instance of this function, if it means we can't avoid a memory allocation somewhere, that's OK.
- There's also a `helper` variable with an anonymous union type, I don't quite understand it.

mvandeberg:
This branch still fails unit tests on OS X. Naively increasing the size of the buffer does solve the problem.

vogel76:
@mvandeberg This code successfully passed valgrind and AddressSanitizer test on Linux where original don't. Could you provide callstack or more info related to crash ?

vogel76:
@theoreticalbts Could you point me why this code fails with precision > 10 ?
Anyway to check it I did test and it fails in both: develop and this branch:
I have put this code at begin of `asset_symbol_validate`
```
   SMT_SYMBOL( alice12, 12);
   SMT_SYMBOL( alice11, 11);
   SMT_SYMBOL( alice10, 10);

   asset alice12a(12, alice12_symbol);
   asset alice11a(11, alice11_symbol);
   asset alice10a(10, alice10_symbol);

   auto ss12 = alice12_symbol.to_string();
   auto ss11 = alice11_symbol.to_string();
   auto ss10 = alice10_symbol.to_string();

   auto as12 = alice12a.to_string();
   auto as11 = alice11a.to_string();
   auto as10 = alice10a.to_string();

   asset a = asset::from_string(as12);
   BOOST_REQUIRE(a == alice12a);

   a = asset::from_string(as11);
   BOOST_REQUIRE(a == alice11a);

   a = asset::from_string(as10);
   BOOST_REQUIRE(a == alice10a);
```

mvandeberg:
Same stack trace as in the issue.

vogel76:
@mvandeberg Could you check that tested chain_test has been correctly rebuilt ? I suppose something was not recompiled and that's cause of same crash.

mvandeberg:
I am 99.9% certain that I rebuilt because I do not normally have my dev env set to build SMTs, but need to to run this test. And I was working on several different branches at the time.

Regardless, I rebuilt and confirmed that the test case still fails on OS X. As stated earlier, the "naive" solution of simply increasing the buffer size solves the issue on OS X.

mvandeberg:
Alright, it turns out it was my error. I went to dig into the code and realized my local branch didn't have any of the changes. Somehow it was not set to track the remote branch. Not sure how that happened. I probably fat fingered a command and broke my branch without realizing it.

It passes.

vogel76:
@youkaicountry @theoreticalbts I have created separate issue #1888 for bug causing bad SMT asset serialization (independent to changes done in this branch). 
Please let me know what is wrong in code processing precision digit as @theoreticalbts commented:

> The digit for precision is a hex digit, not a decimal digit. AFAICT this code doesn't function correctly for precision >= 10.

I will also answer to other remarks too, but please first clarify potential functional issues.

mvandeberg:
Due to the requirement for redoing all asset serialization, I am going to close this PR. It is not going to be a logical step towards the correct code.

PR #1852, opened 2017-12-07
Implement testing scenarios related to smt create operation

Kiwonik:
Implemented testing scenarios of smt creation as described in issue 1653. Modified existing SMT operation tests to use created scenarios. #1653

### DIFF
--- a/libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
+++ b/libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
@@ -25,6 +25,11 @@ public:
       c( *this );
    }
 
+   uint32_t get_nai() const
+   {
+      return symbol.to_nai();
+   }
+
    // id_type is actually oid<smt_token_object>
    id_type           id;
 
@@ -65,6 +70,7 @@ public:
 };
 
 struct by_symbol;
+struct by_nai;
 struct by_control_account;
 
 typedef multi_index_container <
@@ -74,6 +80,8 @@ typedef multi_index_container <
          member< smt_token_object, smt_token_id_type, &smt_token_object::id > >,
       ordered_unique< tag< by_symbol >,
          member< smt_token_object, asset_symbol_type, &smt_token_object::symbol > >,
+      ordered_unique< tag< by_nai >,
+         const_mem_fun< smt_token_object, uint32_t, &smt_token_object::get_nai > >,
       ordered_non_unique< tag< by_control_account >,
          member< smt_token_object, account_name_type, &smt_token_object::control_account > >
    >,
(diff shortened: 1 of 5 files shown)


theoreticalbts:
The `create_smt_3()` function is re-used in several places to create 3 SMT's, but the code is a little strange:  it calls a callback with the symbols of the three created assets.  So you call it like this:

```
create_smt_3( ..., [&](const asset_symbol_type& smt1, const asset_symbol_type& smt2, const asset_symbol_type& smt3)
{
    // do something with smt1
    // do something with smt2
    // do something with smt3
} );
```

It seems easier to change `create_smt_3()` so it returns a `vector< asset_symbol_type >` of length 3, so the above code becomes:

```
vector< asset_symbol_type > sym = create_smt_3( ... );
// do something with sym[0]
// do something with sym[1]
// do something with sym[2]
```

theoreticalbts:
I understand why you're adding the `by_nai` index.  You want to enforce this constraint:

- If some `smt_token_object` `obj` already exists with `obj.symbol.to_nai()` and `op.symbol.to_nai()` identical except for precision, the op should fail.

It turns out we can actually do that with just `by_symbol` index.

For every asset symbol `x` with an embedded precision value, we can define a related *canonical symbol* `CAN(x)` which is the same symbol, except the embedded precision value set to zero:

```
// return asset symbol with decimal places reset to zero
#define CAN(sym) \
   from_asset_num( sym.asset_num & 0xFFFFFFF0 )
```

Then our constraint becomes:

- If some `smt_token_object` `obj` already exists with `CAN(obj.symbol) == CAN(op.symbol)`, the op should fail.

Now `CAN(x)` has the same order as `asset_symbol_type`.  So it suffices to find the next element in the container greater than or equal to `CAN(x)`, which we can conveniently do with the `lower_bound()` function:

```
const auto& idx = _db.get_index< smt_token_index >().indices().get< by_symbol >();
auto it = idx.lower_bound( CAN( o.symbol ) );
```

If there are any objects in the container with `CAN(obj) == CAN(o.symbol)`, all such objects must be in a group together, and must start at `it`.  So we just trigger the exception when `it->symbol.asset_num == CAN( o.symbol )` like this:

```
FC_ASSERT( (it == idx.end()) || ( CAN( it->symbol.asset_num ) != CAN( o.symbol ) ), "SMT has already been created" );
```

theoreticalbts:
I think I'll approve this, let's make separate issues for each of the above comments.

youkaicountry:
I'll create issues for them.

mvandeberg:
I want to see issues before merging this PR so they do not fall through the cracks.

youkaicountry:
@mvandeberg Okay, I was waiting until this was merged to create them so they weren't immediately blocked. I'll make them shortly

mvandeberg:
As soon as they are made, ping me and I'll merge the PR.

youkaicountry:
@mvandeberg #1863 and #1864 have been created, go ahead and merge.

PR #1853, opened 2017-12-07
Remove Authority Challenging

### LINKED ISSUE #1848: Remove challenging authorities
Challenging authorities was a system devised in the summer of 2016 that was never released. The idea was that a user could pay a small fee to challenge the authority of another user, forcing them to sign a transaction with a higher level authority to prove they continued to have control of a key.

This was not released because it could be used for a way to bring a cold storage owner key out to a hot machine to sign a transaction. This was deemed to be undesirable and the feature was scrapped, but the implementation is still in the Steem source code. In order to implement proposed changes in #1847, the challenge authority system needs to be removed. There are a couple of ops with the system that can be repurposed for new ops, but it is imperative that the location of ops in the operation static variant do not change.

Adding this to the HF 20 project because `witness_set_properties_operation` could occupy one of the unused operation tags, helping prevent op bloat.

mvandeberg:
Closes #1848


theoreticalbts:
Looks good.  We should also look at the possibility of removing `last_owner_proved`, `last_active_proved` fields.

mvandeberg:
I will update that in this PR as they should have been removed along with this change.

Will close #1854

theoreticalbts:
Cool, I approve the updates.

Issue #1843, opened 2017-12-04
Add allowed_vote_assets to comment options extensions.


youkaicountry:
In order to determine which assets can be used in voting for a comment, `allowed_vote_assets` struct should be added as a `comment_options_extension`.

As outlined in the whitepaper, `votable_asset_info_v1` will also need to be defined.

vogel76:
@youkaicountry My questions:

1. Who and when can define such votable assets per comment ?
1. Non-existing `allowed_vote_assets` means that all tokens can be used by voters ?
1. How assets are identified in `allowed_vote_assets` - votable_assets map associates acount_name to votable_asset_info_v1, but given account can have defined mutiple SMT tokens

youkaicountry:
@vogel76 
Right now we simply want the structures in place.

1. The user selects which of the assets they are holding can be used for voting.

2. Non-existing allowed_vote_assets should mean that only STEEM can be used for voting. There will be a maximum of three tokens allowed per comment, one of which must always be STEEM. There is an ongoing discussion to decide whether it should actually be listed as one of the vote assets, or assumed to be there #1844. For this ticket, it shouldn't matter, as validation and unit tests will be in seperate tickets. This is only for defining the structures.

3. I see the confusion, this may be a holdover from when there was only one SMT per account. I will consult with @theoreticalbts to clarify this.

theoreticalbts:
Yeah, that's a holdover.  It should be `map< asset_symbol_type, votable_asset_info >`.

vogel76:
@youkaicountry SMT spec says that there can be at most 2 tokens specified in `votable_assets`. Because of that, maybe it is better to rethink `allowed_vote_assets` definition and replace map with pure array. Undefined array entry could be recognized by empty asset_symbol_type object and whole definition could look as follows:
```
struct votable_asset_info
{
  asset_symbol_type asset_symbol; 
  share_type max_accepted_payout    = 0;
   bool allow_curation_rewards = false;
};
struct allowed_vote_assets
{
   std::array<votable_asset_info, 2>  votable_assets;
};
``` 
Such definition could reduce memory usage.

mvandeberg:
I am strongly against a fixed sized data structure in this operation.

[...]

PR #1858, opened 2017-12-07
Issue #1843 Defined data structures and empty visitor methods needed …

vogel76:
…to compile code.


theoreticalbts:
This is failing CI because `FC_REFLECT()` for newly added structs needs to be `#ifdef` protected.

vogel76:
@theoreticalbts Yes it is fixed already, forgot to add #ifdefs to FC_REFLECT too.

theoreticalbts:
Looks good.

PR #1870, opened 2017-12-11
Bump shared memory to 20 GB in default configuration files

jamzed:
As we already crossed 16 GB on consensus nodes the shared-file-size should be updated accordingly.


mvandeberg:
This looks good. Thank you for catching this for us.

To follow our internal development practices, can you please rebase the PR on `develop`?

jamzed:
Hey @mvandeberg, I see contrib/config-for-broadcaster.ini and doc/example_config.ini are missing in develop branch, should I cherry-pick them from master, or you will merge them? Thanks.

mvandeberg:
Yeah, we should update those at the same time. The difference between `master` and `develop` is the greatest it has ever been. We should be resolving that in a week or two.

I am going to go ahead and make a PR myself because I want to bump them all the way to 24G.

Again, thank you for bringing this to my attention.

Issue #1683, opened 2017-10-23
Integrate SMT's with market


theoreticalbts:
The order creation operations `limit_order_create_operation` and `limit_order_create2_operation` should be extended to support SMT's.

We need to check the matching engine works properly when the `multi_index_container` for orders contains orders with multiple assets.  It *should* work but it hasn't been reviewed or tested.

- [ ] Adjust `limit_order_create_operation` and `limit_order_create2_operation` `validate` methods to allow SMTs on  `amount_to_sell`

- [ ] Extend operation tests by SMT based scenarions (`limit_order_create2_operation` specific)

- [ ] Extend operation_tests, operation_time_tests, market_history by SMT based scenarios (`limit_order_create_operation` specific)

- [ ] Create wallet_api tests related to `create_order`method

youkaicountry:
Testing the matching engine has been moved to issue #1814.

vogel76:
@youkaicountry Does limit_order_create2_operation is still effectively used ? limit_order_create_operation object is created in wallet module, but limit_order_create2_operation don't...

youkaicountry:
@vogel76 Good question, actually it seems that both should be maintained. I'll go ahead and modify the issue description to list both.

vogel76:
@youkaicountry Both operations require the `min_to_receive` or `exchange_rate` to be specified in STEEM or SBD. Does it is still true also for SMT ? Maybe these parameters shall be specified also in other SMTs ?

vogel76:
@youkaicountry Although modifying operation_tests probably is not big problem for us (just asset symbol must be changed) I'd like to ask for preparing description of wallet_api scenario using create_order, to be able to correctly verify state after executing the order. Also case when order expires could be valuable.
Maybe let's create separate issue for such API test and link them in this issue description at proper task-item place.

youkaicountry:
@vogel76 Market orders will always be between SMTs and STEEM, never between two types of SMTs.

I'll create issues for the create_order tests with descriptions later today.

vogel76:
[...]

PR #1872, opened 2017-12-11
Issue #1683 - ( 'validate' methods  + unit tests )

mariusztrela:
Issue #1683 - first part


theoreticalbts:
This code permits SMT <-> SBD market, but the spec calls for only SMT <-> STEEM market.

theoreticalbts:
Currently `get_market()` basically returns some value such that `a.get_market() == b.get_market()` if and only if `a, b` involve the same pair of assets.  The current implementation does that fine with SMT's.  Why does it need to be replaced?

theoreticalbts:
It's not written precisely the way I would have done, but it's mergeable.  Maybe I'll do a little refactoring later.

mariusztrela:
a) According to `get_market` - you are right, I removed changes in this method.
b) According `market history plugin` - I made refactoring:
	- variable are embedded in structure `bucket_object_details`. It is more readable.
	- at now is new naming:
		earlier:	`steem <-> sbd`
		now:		`steem <-> sbd_smt`
	- I think, it is worth to add `asset_symbol_type symbol`, because we should know what symbols are put in `bucket_object`

theoreticalbts:
The main concern here is template

We have this code:
```
   bucket_object_details steem;

#ifdef STEEM_ENABLE_SMT
   bucket_object_details sbd_smt;
#else
   bucket_object_details sbd;
#endif
```

Let's refactor to:

```
   bucket_object_details steem;
   bucket_object_details non_steem;
```

Then we can simplify out many of the `#ifdef` that were added in other places.

theoreticalbts:
It is very confusing to have "template variable base class" of the form:

```
class B1 { ... };
class B2 { ... };

/* class C whose base class can be either B1 or B2 */
template< typename B >
class C : public B
{ ... };
```

A less jarring way to achieve this might be:

```
class B { ... };   /* abstract base class defining virtual methods for common functionality of B1, B2 */
class B1 : public B { ... };
class B2 : public B { ... };

/* class C who can use B1 or B2 */
class C
{
    ...
    B b;
};
```

I don't understand some bigger questions:

- (1) What is the difference between `database_fixture` and `clean_database_fixture`?
- (2) Why do you need to subclass both of them?

I suspect the answer to (2) is that "we don't really need to subclass both of them."  In the language of the example above, if `B2` is not actually ever needed, then it suffices to simply do:

```
class C : public B1 { ... };
```

mariusztrela:
1) Refactoring - done. At now we have less `if-def's`.
	
2) I don't agree with you.
	 - Differences between `database_fixture` and `clean_database_fixture` are explained by @vandeberg in #1936. Shortly - constructor in `database_fixture` is empty, but constructor in `clean_database_fixture` does some initializations.

	- Further, the most tests use `clean_database_fixture`, but a few tests use `database_fixture`, because initialization is independent ( inside in given test ).
		
	- I proposed CRTP ( type of design pattern ) : `https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern`.
	  Above pattern solves such problem( inheritance from 2 different classes ) clearly and very readable.

theoreticalbts:
Merge conflicts need to be fixed.

mariusztrela:
Conflicts removed.

youkaicountry:
There are further conflicts to be fixed before it can pass review.

mariusztrela:
Done, no conflicts.

PR #1876, opened 2017-12-11
#1857 Add unit test for unknown key in witness_set_properties

### LINKED ISSUE #1857: Add unit test for unknown key in witness_set_properties_operation.props
As a result of my feedback on the PR #1818, the initial PR for `witness_set_properties_operation`, the code was changed to allow unknown properties to exist, and the unit test which previously checked that unknown properties cause an exception was removed.

We should add a unit test that checks unknown properties don't throw an exception.

NateBrune:
Closes #1857


mvandeberg:
LGTM

PR #1884, opened 2017-12-12
Verify signatures api - Master

### LINKED ISSUE #1674: Verify Signatures API Call

mvandeberg:
Closes #1674
Reflects updates from `develop` PRs #1841 #1726


theoreticalbts:
Looks good.

Issue #1855, opened 2017-12-07
votable_assets validation


youkaicountry:
Validation should be done on `votable_assets` to ensure:
1) There are no repeat asset types
2) There are no more than 2 asset types (Since STEEM is implicit)
3) STEEM is not included as an asset type

- [ ] Update `void comment_options_operation::validate()` by modifying helper visitor processing comment extensions: `comment_options_extension_validate_visitor`
- [ ] Add/extend unit tests to cover implemented functionality in positive and negative cases.

vogel76:
This implementation depends on #1843

PR #1886, opened 2017-12-12
Issue #1855 votable asset validation

vogel76:
Issue #1855:
- added operation validation, also dedicated tests.
Anyway I have a doubts related to `comment_option_operation` validation for case when it has added multiple extensions of same type (ie `allowed_vote_assets`). Visitor(s) used to process such extension do correct work for separate objects, but will ignore another extension objects appearing on the extension list.
This is probably wrong and these visitors shall do checks by doing cumulative checks for given opts, or just to disallow multiple occurrences of given extension type on the same list.
I did try to find how `comment_option_operation` is created but it is not done directly at C++ side 
Same verification problem appears to comment_payout_beneficiaries.


vogel76:
@youkaicountry Any additional info related to above questions ?

youkaicountry:
@vogel76 
`comment_option_operation` is probably created by the front-end only, which is why you couldn't find it. 

The validation visitor should probably be able to check for multiple extensions of the same type, but I am not sure of the consequences of that. @theoreticalbts @mvandeberg would this be okay?

Alternatively, if that is not going to work we used one single extension that contained a vector of asset types?

mvandeberg:
We should be using the same visitor struct to visit all extensions in the vector. We could use a bitmap or similar structure in the visitor and fail if we visit the same entry twice. I think for most operations, this is a reasonable assumption. It would need to be protected by a hardfork check.

libraries/protocol/include/steem/protocol/steem_operations.hpp
```diff
@@ -113,7 +117,59 @@ namespace steem { namespace protocol {
...
+       */
+      void validate() const
+      {
+         FC_ASSERT(votable_assets.size() <= SMT_MAX_VOTABLE_ASSETS, "Too much votable assets specified");
+         FC_ASSERT(is_allowed(STEEM_SYMBOL) == false,
+            "STEEM can not be explicitly specified as one of allowed_vote_assets");
+      }
+
+   private:
       flat_map< asset_symbol_type, votable_asset_info > votable_assets;
```
mvandeberg: For the most part, we make everything public and prefix intended private variables with and underscore.

vogel76:
Member variable renamed as suggested above (even all other members in this file have names without underscore).
There is no need to make the `_votable_assets` public - all needed functionality is covered by class methods and encapsulation is good c++ practice.
@youkaicountry Does you will create another issue related to reusing visitor objects or it shall be done also in this issue ?

youkaicountry:
@vogel76 I'll create an issue for it.

vogel76:
Conflict resolved.

theoreticalbts:
- (a) None of the fields in `protocol` objects should start with an underscore, so should be `votable_assets` not `_votable_assets`
- (b) Don't make reflected fields `private` and use `friend`, instead make all fields in reflected struct `public`
- (c) `add_votable_asset()` and `is_allowed()` methods should be removed.  They are basically duplicating `flat_map` functionality with a slightly different interface for no purpose.  It is shorter and clearer code to simply access `votable_assets` directly.

For (a)-(b), it's for consistency with how other parts of `protocol` are defined.

vogel76:
(a) Maybe define naming conventions. Originally this member was named similar to others in this file, I have changed it according to @mvandeberg request.
(c) `votable_asset_info` construction is inconvenient (needs 2 stage constructor calls to satisfy static_variant - anyway maybe it is not needed here) and that's why I have added this method to simplify usage at caller side.
`is_allowed` method hides internals related to data storage specific to this class (also static_visitor complicates access). When method is removed, most of this code will be needed at caller side in more than 1 place (validate, evaluator, maybe somewhere else too)
(b) This is natural consequence of existing methods covering all functionality and good practice. Atm reader don't need to analyze all code to look how this class is used, where data are just read or modified - everything is encapsulated in the class as it should be.

mvandeberg:
I apologize for the confusion in the naming scheme.

We don't use many private variables. In practice, they don't add much. Many languages don't have private data members at all and do not suffer for it (Python for example).

If a variable would be private, prefix it with an underscore.
However, structs that are meant to be serialized (operations, objects, etc.) by design are public. All of the business logic regarding these objects lives in separate classes with the exception of stateless calls (ex. operation validation).

vogel76:
Conflict resolved, member variable renamed and removed private access specifier.

Issue #1863, opened 2017-12-08
Change create_smt_3 to return a vector


youkaicountry:
Have the `create_smt_3` function return `vector< asset_symbol_type >` of length 3. Discussion in #1852

PR #1887, opened 2017-12-12
Issue #1863 create_smt_3

vogel76:
Issue #1863, built smts are returned to the caller through container as specified in the issue.


vogel76:
Conflict resolved.

Issue #1847, opened 2017-12-06
database::adjust_balance inefficiencies


mvandeberg:
Issue #1682 added the ability to adjust SMT balances through the helper function `adjust_balance( account_object&, asset )`

If the asset is an SMT it then calls `adjust_smt_balance( account_name_type, asset )`. In the transfer evaluator, accounts are looked up, but only to be send to `adjust_balance`. If we are transferring an SMT we now lookup the account, to get the name in `adjust_balance, to pass to `adjust_smt_balance`. And the only thing we need the name for is verifying existence of the account.

The `from` account is guaranteed to exist as part of the authority check in the operation. And the balance has to be taken from somewhere. Either the balance object exists and the transfer can continue, or it does not. This means the account either has a 0 balance or the account does not exist (in the unlikely scenario that the authority system is broken and an can be signed for a non-existent account). We're good.

The challenge is then verifying the `to` account as efficiently as possible. In the case of STEEM, the account is needed to adjust the balance. For an SMT, only the name is needed, but if the balance object does not exist, the account needs to be looked up to verify existence.

I propose we change `adjust_balance` to also accept an account name. If the asset type is STEEM or SBD, then the account will be looked up so it can be modified. If it is an SMT, the balance object is non-existent, and the delta is positive (a deposit) then the account object needs to be looked up to verify the existence of the account. However, in cases where balance objects already exist, we can skip this lookup altogether.

This would need to happen after HF 20 when the sufficient balance checks are officially moved to `adjust_balance`.

Additionally, the challenge authorities need to be removed. Filing another ticket for this.

PR #1905, opened 2017-12-14
Issue #1847 - improvement of database::adjust_balance

mariusztrela:
Issue #1847


theoreticalbts:
Looks good.

mvandeberg:
I am not the biggest fan of having overloaded `adjust_balance` with different time complexities. It would be easy to write code that calls adjust balance via name when the account object is known and vis versa which result in 2x the lookups. But I am not thinking of a better way to do it right now. Ultimately, I think the right decision is going to be to move STEEM and SBD into balance objects as well. However, let's get SMTs working first. And then we can worry about that sort of refactor.

PR #1906, opened 2017-12-14
Automatically Scale Shared Memory

### LINKED ISSUE #1891: Automatically Scale Shared Memory File
We should add parameters to define the behavior for automatically scaling the shared memory file.

I propose the following:

We define two config variables, 

`shared-file-fill-threshold`: A uint16_t percentage (0-10000) that defines how full the file can get before scaling it.
`shared-file-scale-rate`: A uint16_t percentage (0-10000) that define how quickly the file can scale up when the fill threshold is met.

We should add a method to chainbase called, `resize( uint64_t new_file_size )` that will close the database, resize it, and open it, restoring internal references in the process.

The tricky part here is that all undo states have pointers to mapped pages which can potentially become invalidated after the file is resized. Because of this, no undo sessions can exist. Undo states, that live in the shared memory file can exist and will persist because they live within the mapped segment. Chainbase will need to track active undo sessions and fast fail resizing if an undo session exists. It is responsibility of Steem to ensure it has no active undo sessions when resizing.

The end of `_push_block` is a safe place to perform this check. Pending transactions are not a part of state and the `_pending_tx_session` is invalid because `_push_block` is within a `without_pending_transactions` block. Resizing here should be safe and will still restore pending transactions successfully on return from `push_block`.

We also need to check for resizing while reindexing. This can be performed after each `apply_block`. 

While should move `check_free_memory` to these locations as well and combine it with the threshold check. The result of the function is resizing, or displaying a log message.

Boost Reference:

http://www.boost.org/doc/libs/1_63_0/doc/html/interprocess/managed_memory_segments.html#interprocess.managed_memory_segments.managed_memory_segment_advanced_features.growing_managed_memory

mvandeberg:
Closes #1891

This branch is currently a WIP. A PR is being created to begin review of changes to a critical component of the blockchain stack, the Chainbase database.

### DIFF
--- a/contrib/steemd.run
+++ b/contrib/steemd.run
@@ -52,6 +52,10 @@ if [[ ! "$DISABLE_BLOCK_API" ]]; then
    ARGS+=" --plugin=block_api"
 fi
 
+if [[ ! "$DISABLE_SCALE_MEM" ]]; then
+   ARGS+=" --shared-file-full-threshold=9500 --shared-file-scale-rate=1000"
+fi
+
 NOW=`date +%s`
 STEEMD_FEED_START_TIME=`expr $NOW - 1209600`
 
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -211,11 +211,8 @@ uint32_t database::reindex( const open_args& args )
                "   (" << (get_free_memory() / (1024*1024)) << "M free)\n";
             apply_block( itr.first, skip_flags );
 
-            if(    (args.benchmark.first > 0)
-                && (cur_block_num % args.benchmark.first == 0) )
-            {
-               args.benchmark.second( cur_block_num, false /*is_initial_call*/ );
-            }
+            if( (args.benchmark.first > 0) && (cur_block_num % args.benchmark.first == 0) )
+               args.benchmark.second( cur_block_num, get_abstract_index_cntr() );
             itr = _block_log.read_block( itr.second );
          }
 
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -149,6 +149,9 @@ void database::open( const open_args& args )
          auto last_block_num = _block_log.head()->block_num();
          args.benchmark.second(last_block_num, get_abstract_index_cntr());
       }
+
+      _shared_file_full_threshold = args.shared_file_full_threshold;
+      _shared_file_scale_rate = args.shared_file_scale_rate;
    }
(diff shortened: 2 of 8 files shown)


theoreticalbts:
The `int_incrementer` stands on its own.  If we merge it separately as its own PR, it would make the rest of the changes a little easier to review.

theoreticalbts:
You're really checking `free_mem < max_mem * (1 - full_threshold)`.  This is confusing.

Let's check `(max_mem - free_mem) / max_mem >= full_threshold / STEEM_100_PERCENT`.  Add a comment that this is what we're checking.

In the actual code you want to clear denominators in this expression, so the actual (non-commented) check would be something like `(max_mem - free_mem) * STEEM_100_PERCENT >= max_mem * full_threshold`.

theoreticalbts:
Refactoring `add_index()` into `add_index_helper()` could also be its own PR.

mvandeberg:
Yes, they could easily be and are separate commits for that reason.

mvandeberg:
I needed to get a proof of concept testing as quickly as possible yesterday, but tried to at least think enough in advance when structuring my commits.

theoreticalbts:
So I was thinking `resize()` would go like this in pseudocode:

```
resize()
{
// Caller is responsible for making sure no undo sessions are active

new_index_file = open( new_index_filename, "w", size=new_file_size )
for each index:
    for each object in index:
        write( new_index_file, object )
close( new_index_file )

os.rename( new_index_file, index_filename )
}
```

But I guess `managed_mapped_file::grow()` takes care of this?

mvandeberg:
I am not 100% on the specifics of `managed_mapped_file::grow()`, but about the only thing that makes sense to me is that it appends new pages to the current file. This would seem right as we need to unmap the file, grow it, and then remap it. Theoretically, I suppose it could be possible to grow the file, open it with a different fd and map the added pages and then make the allocator aware of that. But that seems like the difficult approach.

We don't get any benefits from defragmenting the file with this approach, but we also don't have to pay the price for copying objects. If you run this on a witness node, you risk missing blocks without some significant changes as you already mentioned.

> Also, copying the file will not be a quick process, if a witness does it, he/she will probably miss blocks. So maybe the witness plugin should suppress scaling if it's recently produced a block. If you're a witness, you could still take advantage of the functionality if you just run two nodes with different block production keys, and update your current block production key to switch between them once a week.

mvandeberg:
@theoreticalbts Rebased to resolve merge conflicts

theoreticalbts:
Looks good to me.

Issue #1885, opened 2017-12-12
Market order matching doesn't always start at top of book


theoreticalbts:
While writing a comment for #1683, I looked carefully at `apply_order()` and I think there may be a bug in it.  If Alice has an existing order and Bob is creating a new order, there's a finite price that's the worst price Bob will accept, and an infinite price that represents the direction of "more favorable to Bob."  The code *starts* at the finite price, then iterates *toward* the infinite price.  Consequence is, if Alice, Amy and Anna all have existing orders that overlap with Bob's new order, then Bob should choose the one that's giving the *best price* (where best/worst price are defined from Bob's point of view).  The code actually gives Bob the *worst price orders* of the matching orders.

What *should actually happen* is starting at the infinite price and moving toward the finite price.

This is a hardforking change that needs to be confirmed by a unit test.

mvandeberg:
My reading of the code indicates that what you are describing _should_ happen _is_ what happens.

``` c++
auto max_price = ~new_order_object.sell_price;
auto limit_itr = limit_price_idx.lower_bound(max_price.max());
auto limit_end = limit_price_idx.upper_bound(max_price);
```

`limit_itr` starts at the infinite price. `limit_end` points to the end of the list, or the first order with a worse price than Bob is willing to accept. This iteration starts at the infinite price and moves towards the finite price.

I find it unlikely that this is wrong because it is the same limit order matching algorithm that Bitshares uses.

https://github.com/bitshares/bitshares-core/blob/master/libraries/chain/db_market.cpp#L260-L262

There are already some extensive limit order tests, but with a quick reading of the tests, I am not sure if any of them test this scenario exactly. It has been almost two years since I looked at the tests.

theoreticalbts:
I'm a bit confused.

I think the code's actually right, and I made a couple mistakes when I decided this might be an issue, first of all I didn't notice [this comparator specification](https://github.com/steemit/steem/blob/da8f6210f4143a79cec7829811cccf42125cb074/libraries/chain/include/steem/chain/steem_objects.hpp#L289) which says the container should be sorted in *reverse order* of `price`.

Second, I wrote some debugging code to help try to understand how this code works, the debugging code prints out `max_price` followed by `max_price.max()`, then prints out `limit_itr` followed by `limit_end`.  So it is very easy to assume, looking at the debugging code output, that `limit_itr` is initialized from `max_price`, and `limit_end` is initialized from `max_price.max()`, but in the actual code they are in reverse order.

PR #1908, opened 2017-12-14
Add test to check order matching

NateBrune:
Closes: #1885


mvandeberg:
Please use `limit_order_create2_operation` to make the test clearer with regard to price and amount to fill.

Also extend the code so that the top of the book are orders A, B, and C. Fill with order D such that A is filled and B is partially filled. Then verify the orders in the order book are the partial B and full C.

@theoreticalbts That should fulfill testing of the scenario described in #1885?

PR #1913, opened 2017-12-15
Crash removed.

mariusztrela:
Crash in destructor `database::~database` because of `out of range` error.  Take a look at
_hardfork_versions[ STEEM_HARDFORK_0_20 ] = STEEM_HARDFORK_0_20_VERSION;

in method: `void database::init_hardforks()`


mvandeberg:
The better thing to do would be to move the HF 20 initialization inside the IS_TEST_NET

mvandeberg:
This PR seems to be a duplicate of #1916 we should chose one of them to close.

vogel76:
Let's choose https://github.com/steemit/steem/pull/1916 as it contains fix with very limited number of changes (just moved #ifdef).

Issue #1515, opened 2017-09-06
Supply checks should support SMT's


theoreticalbts:
The supply accounting should support SMT's.

In more detail:

- There is a method called `database::validate_invariants()` which ensures the sum of tokens in all places where they can exist [1] is equal to the total supply [2].

- Implement `validate_smt_invariants(db)` which does similar checks for SMT's.

The reason for this functionality is to help ensure that we have no bugs that can create / destroy SMT's in the wrong places.  Any code path that intentionally creates / destroys SMT's should be routed through `adjust_supply()` which modifies the supply field of the SMT object.  Unit tests using the `database_fixture` should always run `validate_invariants()` and `validate_smt_invariants()` after the test.

[1] "Places where tokens can exist" includes user balances, market orders, `total_vesting`, escrow, savings, reward balances, etc.

[2] The total supply is tracked in GPO fields `current_supply` (for STEEM) and `current_sbd_supply` (for SBD).  For SMT's, each SMT should have its own supply field on the SMT object.

theoreticalbts:
Added some detail to OP.  @youkaicountry this is not strictly necessary for MVP, but we should do it sooner rather than later so bugs are caught ASAP (this is mostly useful during the initial development phase, if bugs of this kind are going to happen, they will probably happen near the beginning of the project).

youkaicountry:
@theoreticalbts This appears to be a duplicate of #1648

theoreticalbts:
Unassigned myself since I'm not actively working on this at the moment.

Kiwonik:
Partially implemented. Further implementation awaits merging of pending PRs (e.g. PR 1872).

youkaicountry:
#1872 is now merged.

Kiwonik:
Expanded SMT supply control over market orders. Further expansions require that SMT vesting and rewards are implemented.

Kiwonik:
Reopening due to not being complete, see my previous comment above.

youkaicountry:
Closing as outdated.

PR #1914, opened 2017-12-15
SMT support in supply checks

Kiwonik:
Partial implementation of validate_smt_invariants. Extended smt tests with supply checks. Removed a piece of no longer necessary code #1515


theoreticalbts:
Looks good.

Issue #1573, opened 2017-09-15
Rationalizing Market Orders (was: Tick Pricing Rules)


theoreticalbts:
Currently `price` accepts any value for base / quote amounts.  The result is that you can have prices which don't have any decimal representation, like 3 STEEM / 7 SBD.

- Each 1 STEEM is worth `$2.33333333333...` in SBD
- Each 1 SBD is worth `0.42857142857142857...` in STEEM

Additionally you can express base / quote in either order, i.e. you can "flip" the prices at the blockchain level.

Expressing prices like this leads to various counterintuitive UI behaviors, for example, two apparently-overlapping orders do not execute.  This is because the user expectation is that (a) UI's will use decimal notation to express prices, and (b) the UI's decimal value is an exact representation of the effective price on the blockchain.

Let's introduce some rules which I'll call Tick Pricing Rules:

- For prices involving Steem Dollars (SBD), the base asset must be SBD.
- For prices involving SMT assets, the base asset must be STEEM.
- The quote must be a power of 10.
- The base value can have at most 4 nonzero decimal digits.

Basically these rules together mean that trading takes place on "ticks" of between 0.001% and 0.01%.  Let's say STEEM is trading at `$1.253`.  According to the TPR, an order to buy or sell STEEM at `$1.253` will trade at *exactly* that price.  The next larger price would be `$1.254`, and the next smaller price would be `$1.252`.  The "tick" is `$0.001`, which works out to 0.00079% of `$1.252`.  If someday STEEM starts trading at SBD at a larger value, say the `$10.00-99.99` range, then the "tick" will become `$0.01`.  Likewise, once STEEM breaks `$100` barrier, the "tick" will become `$0.10`.

Likewise, the "tick" can also become smaller if necessary, for example if STEEM is `$0.3281` could be expressed as a quote of `0.100 STEEM` and a quote of `3.281 SBD`.

In "flipped" terms, these prices do not have round values (i.e. short, terminating decimal representation), but approximately correspond to `0.79872`, `0.79808` and `0.79745`.

This ticket is to implement a hardfork with the following semantics:

- `feed_publish_operation` and `limit_order_create2_operation` must use prices that obey TPR.
- `limit_order_create_operation` is disabled

theoreticalbts:
[...]

PR #1924, opened 2017-12-18
Add real_price helper to market_history

### LINKED ISSUE #1912: Add real_price helper to market_history
In #1517, `to_real` was removed from the `asset` class because floating point numbers have no place in a blockchain. It appears in testing that Condenser is making use of the `real_price` field in the market.

Instead of adding the back to `asset`, let's create a helper in the API to return data as has been expected (the suggestion in #1517).

mvandeberg:
Closes #1912

### DIFF
--- a/libraries/plugins/apis/market_history_api/market_history_api.cpp
+++ b/libraries/plugins/apis/market_history_api/market_history_api.cpp
@@ -3,6 +3,8 @@
 
 #include <steem/chain/steem_objects.hpp>
 
+#define ASSET_TO_REAL( asset ) (double)( asset.amount.value )
+
 namespace steem { namespace plugins { namespace market_history {
 
 namespace detail {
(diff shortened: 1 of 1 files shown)


theoreticalbts:
It may be possible for orders to have identical `real_price` (especially if truncated to decimal places), but both remain on the books because they don't actually overlap.  By only providing rational values to clients, it should be clear to clients that they could potentially have this case, and it is therefore the client's responsibility if they don't handle it and counterintuitive behavior results.

Additionally, buy / sell orders in the same market have "flipped" units.  This is also counterintuitive.

We should remove the `real_price` field, since there is really no good way to deal with the issue of overlapping prices, unless we implement [tick pricing rules](https://github.com/steemit/steem/issues/1573).

mvandeberg:
I do not disagree with the rationale to remove `real_prices`, but the Steemit.com market relies heavily on the field (perhaps incorrectly). The goal of the APIs in Appbase is to replicate the existing behavior as closely as possible as to not break existing applications. As this is something that apps currently rely on, we need to continue to support it as is.

We can discuss changing/removing the field in the future but it will also require notifying developers and a proper end of life support.

theoreticalbts:
OK, works for me.  I compared it against what's in `master`.  It looks like this should return the same "flippedness" as the old call did.

PR #1928, opened 2017-12-18
Restore `get_ticker` functionality

### LINKED ISSUE #1926: Restore `get_ticker` functionality
In #1571, `get_ticker` was changed so it could compile with changes to `asset`. It needed to be refactor to use a local version of `to_real` as documented in the issue.

An issue to do this refactor was not made and the commented code was subsequently removed in https://github.com/steemit/steem/commit/24a9c155161566945612854ffe15be3fd1e7f76e#diff-ed4afb98aca7b45c93aab1ab6a54ac10

It needs to be restored to its original functionality.

mvandeberg:
Closes #1926

### DIFF
--- a/libraries/plugins/apis/market_history_api/market_history_api.cpp
+++ b/libraries/plugins/apis/market_history_api/market_history_api.cpp
@@ -33,8 +33,16 @@ DEFINE_API_IMPL( market_history_api_impl, get_ticker )
 {
    get_ticker_return result;
 
-   result.latest = 0;
-   result.percent_change = 0;
+   const auto& bucket_idx = _db.get_index< bucket_index, by_bucket >();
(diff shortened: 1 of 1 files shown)


theoreticalbts:
This line raises a couple issues:

```
result.percent_change = ( (result.latest - open ) / open ) * 100;
```

- Should be `STEEM_100_PERCENT` instead of `100`, allowing us to give two decimal places rather than be limited to integer precision
- The result of the division will be between 0 and 1, since it's integer math the result will always be zero, so we have to do the multiplication first.

To summarize, the calculation should be

```
(STEEM_100_PERCENT * (result.latest - open) ) / open
```

mvandeberg:
Because this is a non-consensus API, both `latest` and `percent_change` are doubles.

mvandeberg:
@theoreticalbts Is this PR mergeable?

theoreticalbts:
I guess if they're both `double` we can go ahead and merge it.

PR #1933, opened 2017-12-19
Temp Account Creation Recovery Account

### LINKED ISSUE #1782: Temp Account Creation Recovery Account
Accounts created by the temp account should not have a recovery partner. In the worst case, an account created in this manner is permanently recoverable by the attacker.

mvandeberg:
Closes #1782


theoreticalbts:
All control paths should set `acc.recovery_account`.  Explicit is better than implicit.

We should set it to `STEEM_NULL_ACCOUNT` instead of `account_name_type()`.

mvandeberg:
Force pushed to resolve conflict

theoreticalbts:
This is a little confusing with the conditionals.  We are changing:

```
if( !hf_11 )
    // pre HF 11 behavior
else
   // post HF 11 behavior
```

to

```
if( !hf11 )
    // case (1), pre-HF11 behavior
else if( !temp || !hf20 )
    // case (2)
// else case (3) is a no-op
```

So what exactly do case (1), (2) and (3) represent?  It is not obvious, I couldn't figure it out in my head. 
 I had to go through and write a propositional truth table:

| hf11   | hf20 | temp | Case |
| ------ | ---- | ---- | ---- |
|      F |    F |    F | (1)  |
|      F |    F |    T | (1)  |
|      F |    T |    F | N/A  |
|      F |    T |    T | N/A  |
|      T |    F |    F | (2)  |
|      T |    F |    T | (2)  |
|      T |    T |    F | (2)  |
|      T |    T |    T | (3)  |

The code becomes much clearer if we change the structure a little bit:

```
if( hf11 )
{
   if( hf20 && temp )
      // case 3
   else
      // case 2
}
else
{
   // case 1
}
```

Similarly the below check should be changed from

```
if( !temp || !hf20 )
{
   // case (4)
}
// else case (5) is a no-op
```

to

```
if( hf20 && temp )
   // case (5)
else
   // case (4)
```

And instead of having no-op, you should explicitly set `acc.recovery_account = account_name_type()` and maybe also add a comment reminding that you're setting to `account_name_type()` instead of `STEEM_NULL_ACCOUNT` because of this rule:

> There is a corner case rule that if the recovery account is account_name_type() then the top witness has the ability to recover the account.

theoreticalbts:
This PR needs to be rebased to deal with merge conflicts.

mvandeberg:
@theoreticalbts Resolved

theoreticalbts:
I'm not sure if this message is correct:

> FC_TODO( "If after HF 20, there are no temp account recoveries, the HF check can be removed." )

Suppose temp creates Alice.  The old way of setting Alice's recovery account is to set it to temp, the new way of setting Alice's recovery account is to set it to `account_name_type()`.  If temp's creation of Alice occurs before HF20, then the old way allows temp to recover Alice at any time, *even after HF20*.

So suppose we have two nodes M and N, and the following sequence of events:

```
temp creates Alice
M, N upgrade to HF20
FC_TODO is fixed
N upgrades to fixed version
temp recovers Alice
```

At the end of this sequence of events, `M` and `N` will stop syncing.  From `M` point of view, the recovery transaction is valid since (according to `M`) `temp` is Alice's recovery account.  From `N` point of view, the recovery transaction is invalid since `account_name_type()` is Alice's recovery account.

We really need to have one of two things:

- (a) Hardfork processing which iterates over accounts and fixes any account which converts recovery account from `temp` to `account_name_type()`.
- (b) Change message of `FC_TODO()` to clarify that the code can only be simplified if no one creates an account with the temp account as the recovery account before the hardfork

theoreticalbts:
Looks reasonable

PR #1935, opened 2017-12-19
Deprecate account_create_with_delegation

### LINKED ISSUE #1760: Deprecate account_create_with_delegation_operation
The operation should no longer be valid after HF 20

mvandeberg:
Closes #1760


mvandeberg:
Merge conflicts, will fix.

theoreticalbts:
BitShares had code to have one file per HF issue, and then the build scripts concatenated the files.  It was coded that way specifically to prevent this kind of merge conflict.  Maybe we should consider porting that build hackery to Steem.

PR #1939, opened 2017-12-21
Add documentation - rewards in hf19

joticajulian:
The reward system is a part of Steem that is changed constantly every hardfork. If you find a post explaining the details be careful because maybe the post is from last year and therefore obsolete. How to deal with that? we need more documentation on the github repository explaining the details of every change. This work consist in a pull request explaining the reward system for authors and curators in HF19.

### DIFF
--- a/doc/rewards-hf19.md
+++ b/doc/rewards-hf19.md
@@ -1,0 +1,121 @@
+# Introduction
+
+In this document we show the calculations used in the reward system for authors and curators in Hard Fork 0.19.
+
+# Voting power
+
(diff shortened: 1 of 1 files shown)


mvandeberg:
Thank you for putting in the work to document the rewards mechanism present in Steem.

We actually have a public repo for the [Steem Whitepaper](https://github.com/steemit/whitepaper) and want to keep it up to date. If you find inaccuracies there or see topics that need further explanation, you are more than welcome to make PR against that repo directly.

PR #1942, opened 2017-12-22
Add test to check order matching

### LINKED ISSUE #1885: Market order matching doesn't always start at top of book

mvandeberg:
Closes #1885 
Internal PR for #1908


theoreticalbts:
Looks good to me, it tests the condition I was worried about

PR #1955, opened 2017-12-28
BIP-0062 Signature Canonization

### LINKED ISSUE #1944: "Canonical" transaction signatures
To address the [well-known signature-based transaction malleability problem](https://en.bitcoin.it/wiki/Transaction_malleability), steemd enforces the use of "canonical" ECDSA signatures. But the algorithm steemd uses is unexplained, is not the same as the one described in BIP-0062, and does not to my knowledge correspond to any generally accepted principle of elliptic curve cryptography.

**The desired outcome of this issue is that the algorithm steemd uses should be explained in such a way that anyone with a good working knowledge of ECDSA can understand it.**

To recap some basic principles:

"Transaction malleability" refers generally to the ability to alter the semantic content of a transaction without altering its signature.

ECDSA signatures are intrinsically malleable in the sense that for any valid elliptic curve signature `(r, s)` of a message hash `m`, the signature `(r, -s (mod N))` is also valid, where N is the order of the base point `G` of the curve. "Valid" here means that the ECDSA signature validation algorithm will confirm that both `(r, s)` and `(r, -s (mod N))` are valid signatures of the hash `m` for the public key that produced `s`.

This property can be exploited for replay attacks, in which one transaction, identified by the hash of its contents, is submitted twice with the two different signatures.

[BIP-0062 proposed](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures) to address the intrinsic malleability of signatures by enforcing that only signatures for which the invariant `s < (N/2)` holds are valid signatures. This prevents the attack described above.

There is additionally the concern that that for any valid signature `(r, s)` of a message hash `m`, there are four separate public keys (i.e., points on the elliptic curve) that could have generated S. Although it is mathematically intractable for an attacker to generate a private key corresponding to the three other public keys, it is desirable for authentication purposes to deterministically identify which public key was, in fact, used to generate a signature.

Deterministic public key recovery is addressed by addition of a recovery parameter, which by the Bitcoin standard is a single byte in which the two least significant bits are used to identify the one of the four public keys that actually signed the message. Therefore the length of the "recoverable" signature in its most compact representation is 65 bytes: the recovery parameter (1 byte), `r` expressed as a 256-bit number (32 bytes), and `s` expressed as a 256-bit number (32 bytes).

Now to turn to steemd: the way steemd defines signature canonicity is via [this code in fc](https://github.com/steemit/fc/blob/c009d2d8332a3ce9936c00deb6f4629555f26fad/src/crypto/elliptic_common.cpp#L171):

```c++
    bool public_key::is_canonical( const compact_signature& c ) {
        return !(c.data[1] & 0x80)
               && !(c.data[1] == 0 && !(c.data[2] & 0x80))
               && !(c.data[33] & 0x80)
               && !(c.data[33] == 0 && !(c.data[34] & 0x80));
    }
```

`compact_signature` [is defined here](https://github.com/steemit/fc/blob/c009d2d8332a3ce9936c00deb6f4629555f26fad/include/fc/crypto/elliptic.hpp#L25):
```c++
typedef fc::array<unsigned char,65> compact_signature;
```

There is no further documentation of `compact_signature` but it seems clear from its usage that this is a Bitcoin-type compressed 65-byte signature of the format I described above. Therefore the conditional expression in this code could be described by the following pseudocode:
```
procedure is_signature_canonical(signature):
  if signature.r is less than 2^255 and
    signature.r is less than 2^246 and
    signature.s is less than 2^255 and
    signature.s is less than 2^246:
    return true
  else:
    return false
```

The following flaws become immediately apparent:
1. The first and third conditionals are redundant. If `c.data[1]` and `c.data[33]` have bit 0x80 set, they obviously cannot be equal to zero.
2. The value of `r` is checked. This to my knowledge has no basis in the literature.
3. The check is incorrect from the perspective of BIP-0062, which says `The value S in signatures must be between 0x1 and 0x7FFFFFFF FFFFFFFF FFFFFFFF FFFFFFFF 5D576E73 57A4501D DFE92F46 681B20A0 (inclusive).` (the long value is simply the hexadecimal value of `N/2`)

**I would like to get answers to the following three questions:**
1. What is the meaning of `canonical` in the context of this function? Is it intended to refer to BIP-0062?
2. Why is the value of `r` checked?
3. How was the value 2^246 selected, and what does it signify?

I further propose that in the interest of simplicity the function `is_canonical` be rewritten as follows:
```c++
    bool public_key::is_canonical( const compact_signature& c ) {
        return c.data[1] == 0 && 
              c.data[2] & 0x80))
               && !(c.data[33] & 0x80)
               && !(c.data[33] == 0 && !(c.data[34] & 0x80));
    }
```

mvandeberg:
Closes #1944

steemit/fc#34 needs to be merged first.

Merge conflict is due to `fc` and will be fixed when the above PR is merged.


sneak:
Please get external review before merge.

mvandeberg:
Replaced by #2500

PR #1970, opened 2018-01-08
Add tag stats update skip flag

### LINKED ISSUE #1965: Tags "Optimization" is causing Steemd launch timeout
In #1612, I implemented two optimizations to the tags plugin.

One of the optimizations was based on the observations that the current state of tags is not dependent on any transient state, but only the current state. Because comments are frozen after payout, the tags plugin only has to record such statistics on payout. Then, on completion of reindexing, the stats for all active comments is calculated.

The time of this calculation would appear to have been severely underestimated. I believe this is causing steemd nodes to not start in time in the dev environment. I have an EC2 node starting from the most recent state and have observed the process stuck in this processing for several minutes. I believe it is progressing because steemit-dev-steemdsync is continuing to generate logs. However, looking at logs, it appears that it can take close to 20 minutes to do this processing. 

This morning I started a deployment at 7:24. At 7:55 steemd was starting. By my estimate, it would start syncing at 8:15. Deployment timeout is 45minutes and we are already over that by 10 minutes (no sync time accounted for).

We should add a runtime flag to skip this check if you know your tag stats are up to date (in the case of starting steemd from an existing state file). This should save us the entire 20 minutes spent doing this calculation.

mvandeberg:
Closes #1965

### DIFF
--- a/contrib/steemd.run
+++ b/contrib/steemd.run
@@ -75,6 +75,8 @@ if [[ ! -d $HOME/blockchain ]]; then
         tar xvjpf /var/cache/steemd/blocks.tbz2
         chown -R steemd:steemd $HOME/blockchain
     fi
+else
+   ARGS+=" --tags-skip-startup-update=true"
 fi
 
 # without --data-dir it uses cwd as datadir(!)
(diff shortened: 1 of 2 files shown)


theoreticalbts:
LGTM

Issue #1751, opened 2017-11-10
Add support for chain_id to sign_transaction


theoreticalbts:
The `sign_transaction` command line tool needs to support a custom `chain_id`.

PR #1972, opened 2018-01-08
1751 sign transaction tool

vogel76:
Issue #1751 sign_transaction tool extended by parameter allowing to specify chain-id to be used at transaction signing.


theoreticalbts:
Command line argument handling should use `boost::program_options` instead of C-style string processing.

Since this code solves an immediate problem, I've approved it as-is, since it functionally accomplishes what we need.

Let's file a separate ticket to convert this code to use `boost::program_options`.

vogel76:
Life of this tool is limited due to actual signing code to be done in Python, so I decided to use direct C style arg processing. Also tool is very simple so probably boost::program_options is too big cannon for that.

mvandeberg:
The point of the tool is not for test net transaction signing. Having a signing tool in C++ and one in Python are not mutually exclusive.

I agree with @theoreticalbts that we should use Boost program options if for no other reason that to be consistent with our command line args across our C++ binaries.

PR #1980, opened 2018-01-10
Incorrect index in 'list_votes' method.

mariusztrela:
Two small mistakes in `list_votes` method ( database_api ).


mvandeberg:
I am approving, but all PRs, including small ones like this, should have an accompanying issue.

Issue #1986, opened 2018-01-12
Implement json_rpc plugin option to dump all API request and responses


vogel76:
Option should dump data in a form matching scheme of API automatic tests suite (see #1567) to be able automatically generate test inputs and initial patterns ie by using a condenser connected to given steemd instance.

All changes are done in state_model_opts2 branch, must be separated to dedicated branch and rebased according to current develop status.

mvandeberg:
The plugin used to do an idump of all accepted requests. Let's add that back with a custom logger (see p2p code) so that we can specify a log location in the config file.

PR #1989, opened 2018-01-12
1986 dump api call request and response

mkochanowicz:
Issue #1986


mkochanowicz:
@mvandeberg Maybe option --log-json-rpc has wrong name - it is not logging but dumping (name --dump-json-rpc will be more adequate). We dump each pair of request and response into separate json files and finally generate yaml config file to be ready to run with `pyresttest`. Maybe logging is good idea, but it should be a new issue. This part is only for generate inputs for automatic tests with `pyresttest`.

mvandeberg:
I understand what this PR is doing and acknowledge that it is technically correct. My feedback is stylistic in nature. Maybe I didn't explain myself well enough.

The FC logging library already handles file output as well as console output. We already have the config parsing to specify arbitrary loggers and outputs. See the default config and the p2p logger destination. If you set up the jsonrpc plugin to use another logger, then dumping to a file is as simple as `idump( (request) );` There is no need to have a log option specific for this. Let's use existing constructs in our code base. In the past it has been useful to log rpc requests to console as well. If this issue is implemented in this way, then we can config nodes to dump to a file or log to console without needing to change code, just config. All of the heavy lifting for this issue is already done in our libraries. The rest of our code base already uses that convention, let's continue using that convention rather than implementing a custom solution for this particular issue.

mkochanowicz:
Can I dump each request/response to new files with custom names with this approach? What with final yaml config file listing all the tests (pairing request+response and approriate validator)? Now it is generated in dumper dtor (when we are sure no more rpc).

mvandeberg:
Yes, you can specify custom file. Please look at our example configs.

```
# File appender definition json:  {"appender", "file"}
log-file-appender = {"appender":"p2p","file":"logs/p2p/p2p.log"}
```

mvandeberg:
I'm sorry, I read your last message too quick. You want each request in a new file?

Can't you log using the existing logger and then parse the file with sed or some other tool?

mkochanowicz:
@mvandeberg This is done as you have suggested.

libraries/plugins/json_rpc/json_rpc_plugin.cpp
```diff
@@ -227,6 +232,12 @@ namespace detail
       {
          response.error = json_rpc_error( JSON_RPC_INVALID_REQUEST, "jsonrpc value is not \"2.0\"" );
       }
+
+      idump( (request) );
```
mvandeberg: Let's make these debug level logging (`ddump`). Our nodes are set up for info level logging, and this is a bit much for our logging infrastructure to save.

python_scripts/parse_json_rpc_log.py
```diff
@@ -0,0 +1,103 @@
+# python
+"""
+Module parse json_rpc log file and create request/response.result pairs for each call.
+Additionally yaml file is generated to execute these calls with pyresttest.
+
+json_rpc_plugin will generate log messages if you add lines like these into config.ini:
+log-file-appender = {"appender": "json_rpc", "file": "logs/json_rpc/json_rpc.log"}
+log-logger = {"name": "json_rpc", "level": "info", "appender": "json_rpc"}
```
mvandeberg: Change this to debug as well

mkochanowicz: Done

libraries/plugins/json_rpc/json_rpc_plugin.cpp
mkochanowicz: Done.

mvandeberg:
I just doubled checked, and there is already a ddump in `json_rpc_plugin_impl::rpc`. With this patch, all successfully parsed messages will be dumped twice. I am trying to reconcile the requirements of this branch with the requirements of our production environment. Logging the requests has been very successful in the past to help diagnose problems. Logging of the responses too, can be helpful, but is extremely spammy.

As I understand this issue, we want to dump API requests and responses to generate a test suite of expected results. The problem with this in practice is that it needs to be on a specific block. Testing of this with steemd is pretty straight forward. Testing other services such as SBDS and Hivemind are more difficult. I am not sure if they are designed to sync up to a point and stop. This could be forced by having them sync to a steemd node stalled at a specific block. The work to get a version of a stack deployed in this particular state could be worth it, but it could be costly in terms of developer hours. @jredbeard @goldibex Thoughts?

I hate to send this back again, but after thinking through these requirements I think the following is the best course of action.

1. Choose either `rpc_jsonrpc` or `rpc` as the method to log requests and responses. Remove logging from the other.

2. Make logging responses a runtime flag `jsonrpc-log-responses` so that we can easily log responses for test suite generation and not spam our production logging.

I really like the idea of testing our APIs in this manner, but I have my doubts of the long-term efficacy given the unique challenges posed by the blockchain.

mvandeberg:
With `--jsonrpc-log-reponses` I wanted just the responses to be logged via this flag. Requests should always be logged. This is a hacky way of creating a second debug log level.

mkochanowicz:
@mvandeberg To be sure - you mean sth like this:
```
      // dump request only to main logger
      ddump( (request) );

#ifdef DEFAULT_LOGGER
# undef DEFAULT_LOGGER
#endif
#define DEFAULT_LOGGER "json_rpc"

      // dump request/response on demand to specified in config.ini logger
      if (_jsonrpc_log_responses)
      {
         ddump( (request) );
         if ( response.result.valid() )
            ddump( (response.result) );
         else
            ddump( (response.error) );
      }
```

mvandeberg:
@mkochanowicz I was more thinking

```c++
ddump( (request) );

if( _jsonrpc_log_responses )
{
   if( response.result.valid() )
      ddump( (response.result) );
   else
      ddump( (response.error) );
}
```

mkochanowicz:
@mvandeberg Are you know in such case (if DEFAULT_LOGGER is defined locally), request will never appear in main log as before?

mvandeberg:
Continue to define `DEFAULT_LOGGER` at the top of the file.

mvandeberg:
I believe if the specific logger is never configured, it will default to main logger. We should verify this behavior. If it does not we can fix it in a different issue.

vogel76:
@mkochanowicz @mvandeberg 
As it was directly discussed/clarified main goal of this branch was to implement an option for **internal use**. The simplicity of initial implementation has been significantly violated by requests to use common solutions which are not applicable for this.
I propose to get back to the last commit (remove all further changes):
https://github.com/steemit/steem/pull/1989/commits/1cfb75993715ede601f0fe9bdbf0d1e109c9285d
and merge this PR when it will be ready.

mkochanowicz:
> I believe if the specific logger is never configured, it will default to main logger. We should verify this behavior. If it does not we can fix it in a different issue.

@mvandeberg It works as follows:
1. If `DEFAULT_LOGGER` not defined all messages go to main logger.
1. If defined, then:
- if properly configured in `config.ini` messages go there
- if not configured messages are disappeared.

So mine piece of code will do what we want (`--jsonrpc-log-reponses` looks redundant IMHO).

mvandeberg:
Let's revert the commit as @vogel76 suggested and we can figure out a better way to handle the logging later if we want to integrate it into the rest of the logging system. Long term, I think that is the best way to handle it, but I want to unblock your work on the PR first.

mkochanowicz:
New PR created: #2071.

Issue #1985, opened 2018-01-12
Extend reindex-process benchmark data collector by report containing info about built datastructures


vogel76:
All changes are already done inside state_model_opts2 branch. We need to rebase it against current develop and put changes into dedicated branch to make clear PR for it.

PR #1991, opened 2018-01-15
1985 extend benchmark data collector

mkochanowicz:
Issue #1985


libraries/chainbase/include/chainbase/chainbase.hpp
```diff
@@ -467,6 +505,13 @@ namespace chainbase {
             _revision = revision;
          }
 
+         void remove_object( int64_t id )
```
mvandeberg: Why is this being added back in?

mvandeberg:
We might be able to utilize the FC_REFLECT visitors to make dynamic size benchmarking templated. If this is a direction we continue to go for performance monitoring, I would suggest we look into that to make the benchmark code less tedious. However, than can be done later.

libraries/chainbase/include/chainbase/chainbase.hpp
mkochanowicz: @mvandeberg Accidentally added back. Removed in last commit 217e200993a70858e9f10358c173ace10cf3f256

PR #1993, opened 2018-01-15
Legacy Asset Serialization

### LINKED ISSUE #1937: Change Asset String Serialization
Assets are currently serialized as `1.000 STEEM`. Parsing this as an integer value by apps interacting over APIs requires custom parsing of the string. The biggest reason we use this serialization is that it is easy to use via the cli wallet.

We are going to change to a different serialization format to make handling assets easier via API.

The new format will be a tuple of `[satoshis, precision, nai]`. We will also assign special NAIs to represent STEEM, SBD, and VESTS.

The previous example would then serialize as `[1000, 3, NNNN]`. Where `NNNN` is whatever we end up using as the reserved STEEM nai.

All APIs returned via condenser API will return the existing format. Other APIs will return the new format. This will release with Appbase so the new APIs use the new format at launch.

mvandeberg:
Closes #1937

Critical sections of this PR are serialization logic and tests. Any objects that are not successfully converted to `legacy_asset` for `condenser_api` will be caught quickly in testing of our stack.


theoreticalbts:
Has it been checked that this can replay the blockchain?

In particular, the modifications to `legacy_asset` seem suspect.  I created `legacy_steem_asset` in particular because `chain_properties::account_creation_fee` at one time did not have its symbol checked, and some operations with an incorrect symbol got into the blockchain.  It does not seem that the new code correctly treats all the possibilities where`ser` is one of these non-canonical serializations.

We really have two pieces of functionality here:

- (a) An asset type whose instances accept any of the non-standard binary serializations of `STEEM` which were included in the blockchain before `chain_properties::account_creation_fee` was properly checked
- (b) An asset type whose instances serialize to strings like `"1.234 STEEM"`

This patch attempts to combine both of these pieces of functionality into a single type.  However, they apply to different fields for different purposes.

So let's move `legacy_asset` to `condenser_api`, and restore the old version of `legacy_steem_asset` / `legacy_steem_asset_symbol_type` in `protocol`.

mvandeberg:
It shouldn't have problems syncing. The reason I added it to the existing `legacy_steem_asset` was because that asset could represent the entirety of the old serialization (raw and variant) and the new asset could represent the entirety of the new serialization for SMTs. By this logic all overwritten to/from variant serialization doesn't need to go in protocol because it does not impact the serialization of blocks and signatures.

I can move it to condenser API to make review easier, but is not quite as simple because the existing legacy types need to be converted as well to this new, third, type of asset.

It shouldn't have problems reindexing because the branch should not modify any of the raw serialization. If it does then the implementation is wrong. The only thing that was changed is allowing any asset to be converted into a legacy asset. However, for consensus, we only do the conversion in the opposite direction. Legacy assets in ops become smt compliant assets in data structures.

mvandeberg:
I will double check it reindexes before merging to be safe.

theoreticalbts:
OK, so apparently I made a mistake in the above review -- I mistook `from_asset()` for `unpack()`.  It would still be better to have two separate types, but because it is functionally correct and there is a lot of code here, I'm not going to block the merge on that.  Approved.

mvandeberg:
This branch has been shown to successfully reindex.

Issue #1896, opened 2017-12-13
Create claim_reward_balance2_operation


youkaicountry:
The original `claim_reward_balance_operation` had no extensions field, and does not fit our specification, so we will create a new operation. Discussion in #1859

`claim_reward_balance2_operation` should contain an extensions field, and it should accept a vector of tokens that will be rewarded to an account.

- It should be able to store STEEM/SBDs/VESTS as well as SMTs.
- Multiple assets of the same symbol should be disallowed

vogel76:
1. Does the new operation shall store only SMT related assets or also STEEM and SBDs ?
1. Does it is allowed to store multiple assets of the same symbol ?

youkaicountry:
(1) It should be able to store STEEM and SBDs as well. 
(2) I don't see any reason it should be able to store assets of the same symbol, so that is something we should disallow

mvandeberg:
(1) And VESTS

Duplicate assets would not make sense in the operation, so it should be disallowed.

mvandeberg:
I would suggest using a flat_set to sort assets, but comparing assets with different symbols is not allowed. This is generally a good thing because we do not want to succeed transferring 10 CATCOINS from Alice's 5 DOGCOIN balance because the CATCOIN symbol ends up being greater than 5 DOGCOINS.

We could create a `comparable_asset` class with different operators and an explicit conversion to and from asset to be careful how we use it.

This way assets will be ordered by asset symbol and can be guaranteed unique.

PR #1997, opened 2018-01-16
Created claim_reward_balance2_operation.

Kiwonik:
Also added validation test and implemented evaluation. #1896


youkaicountry:
Until it passes CI it can't be properly reviewed. Any idea what is causing it to fail?

mvandeberg:
CI box was full. It has been cleaned up. Started build.

theoreticalbts:
It's close, but see #2067

mvandeberg:
Do we want to merge this or wait for #2067, @youkaicountry

youkaicountry:
Let's not merge it yet, a structural change need to be made, so I'd rather not have it in the repo in this form yet.

The best way to proceed rather than modifying fc to allow use of a flat_set, is to use a vector and validate it. 

From #2067:
Use vector< asset > instead, and have validate() enforce reward_tokens[i].symbol > reward_tokens[i-1].symbol for all i > 0

Issue #1907, opened 2017-12-14
Use visitor to visit all operation extensions


youkaicountry:
Visitors ignore multiple extensions of the same type, so we should make it able to visit them.

From #1886:
> We should be using the same visitor struct to visit all extensions in the vector. We could use a bitmap or similar structure in the visitor and fail if we visit the same entry twice. I think for most operations, this is a reasonable assumption. It would need to be protected by a hardfork check.

Kiwonik:
Turned out to be unnecessary, see comments to its PR https://github.com/steemit/steem/pull/2005

PR #2005, opened 2018-01-17
Equipped comment extensions visitors with duplication checks.

Kiwonik:
I used STEEM_HARDFORK_0_20__1815 as HF guard (my best guess). #1907

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -505,7 +505,8 @@ struct comment_options_extension_visitor
 #ifdef STEEM_ENABLE_SMT
    void operator()( const allowed_vote_assets& va)
    {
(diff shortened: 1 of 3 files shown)


mvandeberg:
I don't think this is needed. I'll take the blame for not updating the issue.

Equality for a static_variant is not determined by the contained value, but simply by the encapsulating type.

``` c++
friend bool operator == ( const static_variant& a, const static_variant& b )
{
   return a.which() == b.which();
}
friend bool operator < ( const static_variant& a, const static_variant& b )
{
   return a.which() < b.which();
}
```

Because a flat set does not allow duplicate values, a flat set of static variants does not allow for duplicate types. Only one of each extension is allowed in the set, making the duplicate check redundant.

Kiwonik:
Indeed, a good catch of what would be a mere code bloat.

Issue #2029, opened 2018-01-24
Create smt_market_maker_state struct


youkaicountry:
Market maker state `smt_market_maker_state` should be be a struct embedded in `smt_token_object`, that represents the current state of an SMT's market.

Create this struct with the following conditions:
- It should be embedded in the `smt_token_object` with which it is associated.
- Each SMT should have one and only one market struct.
- Additionally, it should contain the following member fields to track its state:
  - `asset steem_balance;` - s in the paper
  - `asset token_balance;` - t in the paper
  - `uint32_t reserve_ratio = 0;` - r in the paper

PR #2035, opened 2018-01-25
Embedded market maker struct in smt_token_object

Kiwonik:
The new member replaced `current_supply` member too. #2029


youkaicountry:
There seems to be a miscommunication, the `token_balance` member represents the balance of the token held by the market (t in the paper), not the total supply. The supply should still be held in the `current_supply` member of the object. 

I removed the member from the original spec that represented total supply (T in the paper) from this struct.

Kiwonik:
Yes, that exactly what happened, i removed a member that has already been removed from (now) embedded structure. A commit restoring the member I removed has already passed the tests. Ready to merge again.

PR #2046, opened 2018-01-26
Fix legacy_operations serialization to match non-legacy serialization

### LINKED ISSUE #2045: Remove "legacy_" prefix on legacy operations
Condenser appears to use the `type` field of the operation static variant to dispatch formatting. #2036 was returning valid JSON (not JSON within a string). The reason it did not work was because the `type` field was returned as an integer (Described in #2044). However, for the legacy operations to work in Condenser (and perhaps other interfaces) the legacy operations need to return identically to how the would have normally. This includes removing the `legacy_` prefix of such operations.

Thankfully, this is quite simple. The operation serialization uses a method called, `name_from_type` that transforms `steem::protocol::transfer_operation` to `transfer`. We can extend this to transform `steem::protocol::legacy_transfer_operation` to `transfer` instead of `legacy_transfer`.

mvandeberg:
Closes #2045

As stated in the issue #2038 was not sufficient. This was tested against Condenser and works.

### DIFF
--- a/libraries/plugins/apis/condenser_api/CMakeLists.txt
+++ b/libraries/plugins/apis/condenser_api/CMakeLists.txt
@@ -3,6 +3,7 @@ add_library( condenser_api_plugin
              condenser_api.cpp
              condenser_api_plugin.cpp
              condenser_api_legacy_asset.cpp
+             condenser_api_legacy_operations.cpp
              ${HEADERS} )
 
 target_link_libraries( condenser_api_plugin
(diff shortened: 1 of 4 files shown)


theoreticalbts:
This may not be the most elegant way to do it, but it looks functionally correct.  So I'll let it go.  Hopefully all code in this plugin will eventually die out, and this code won't be in there long-term anyway.

mvandeberg:
It will be deleted along with condenser API.

PR #2051, opened 2018-01-29
Update default webserver threads to 32

### LINKED ISSUE #2013: Reduce Webserver Thread Default
The current default is set at 256.
Our config files are all 1024.

Let's reduce this to something more reasonable, say 32.

mvandeberg:
Closes #2013


theoreticalbts:
Why was it so high to begin with!?

PR #2061, opened 2018-01-30
exchangequickstart.md - Update RAM requirements

TimCliff:
The RAM requirements for an exchange node have increased. Exchange nodes should have at least 24GB of RAM, recommended 32GB.


mvandeberg:
The shared memory requirements for consensus and exchange nodes will soon increase. RAM requirements remain unchanged.

PR #2070, opened 2018-02-01
Updated mem settings and info to reflect current state of Steem

Gandalf-the-Grey:
Changed default shared memory size to 40GB in config files and updates some docs about mem requirements.
Of course automatic shared memory file resize would be preferred, but this PR could be used as a temporary fix for default docker installations to run out of the box before auto-resize will be included in stable.


mvandeberg:
@jredbeard Can you double check this?

PR #2077, opened 2018-02-02
Don't use lazy boost::promise 

### LINKED ISSUE #2076: Segfault on Appbase Network Broadcast API
I've encountered a segfault in Appbase's version of `network_broadccast_api::broadcast_transaction_synchronous`.

The long-term solution is the deprecation of this API call and providing a transaction query API (first proposed #486). However, for the time being we need to support this call as it exists for Appbase compatibility.

```
fc::segfault_handler(int, siginfo_t*, void*)+0x1c
boost::promise<steem::plugins::network_broadcast_api::broadcast_transaction_synchronous_return>::lazy_init()+0xa1
steem::plugins::network_broadcast_api::detail::network_broadcast_api_impl::on_applied_block(steem::protocol::signed_block const&)+0x390
boost::signals2::detail::signal_impl<void (steem::protocol::signed_block const&), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (steem::protocol::signed_block const&)>, boost::function<void (boost::signals2::connection const&, steem::protocol::signed_block const&)>, boost::signals2::mutex>::operator()(steem::protocol::signed_block const&)+0x150
steem::chain::database::notify_applied_block(steem::protocol::signed_block const&)+0x30
steem::chain::database::_apply_block(steem::protocol::signed_block const&)+0x2ad
steem::chain::database::apply_block(steem::protocol::signed_block const&, unsigned int)+0x5f
steem::chain::database::_push_block(steem::protocol::signed_block const&)+0x101
steem::chain::database::push_block(steem::protocol::signed_block const&, unsigned int)+0x97b
steem::plugins::chain::chain_plugin::accept_block(steem::protocol::signed_block const&, bool, unsigned int)+0x4d
steem::plugins::p2p::detail::p2p_plugin_impl::handle_block(graphene::net::block_message const&, bool, std::vector<fc::ripemd160, std::allocator<fc::ripemd160> >&)+0x1af
fc::task_base::run_impl()+0x44
fc::thread_d::process_tasks()+0x50b
fc::thread::exec()+0x3b
```

It appears I had seen this stack trace and thought it had been fixed #1781

Both errors occur in `boost::promise::lazy_init`. The default behavior was changed to longer lazily init the future in version 3 of the Boost Thread library (By default, steemd is compiled with version 2 of the library). I would like to consider upgrading to version 3 of the library (possibly 4), but I suspect we may introduce some unintended side effects with FC cooperative multi-tasking's already liberal/abusive use of the Boost threading library.

For now, I propose defining `BOOST_THREAD_DONT_PROVIDE_PROMISE_LAZY` during compilation to not use this behavior. From a cursory inspection of `fc::promise` it does not look like that FC is using a lazy implementation. The current behavior of `broadcast_transaction_synchronous` is based heavily on the old, fc::promise, version. I think there is a decent change that eliminating the lazy initialization will stop the crash even if our implementation has a race condition. At the very least, we will get a different stack trace that might help us diagnose the problem.

mvandeberg:
Closes #2076

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,8 +237,8 @@ if(ENABLE_COVERAGE_TESTING)
 endif()
 
 # fc/src/compress/miniz.c breaks strict aliasing. The Linux kernel builds with no strict aliasing
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing -Werror" )
-SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-strict-aliasing -Werror" )
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing -Werror -DBOOST_THREAD_DONT_PROVIDE_PROMISE_LAZY" )
+SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-strict-aliasing -Werror -DBOOST_THREAD_DONT_PROVIDE_PROMISE_LAZY" )
 
 # external_plugins needs to be compiled first because libraries/app depends on STEEM_EXTERNAL_PLUGINS being fully populated
 add_subdirectory( external_plugins )


theoreticalbts:
> The long-term solution is the deprecation of this API call

Yes, yes, please yes.  But for now, hopefully this will make it slightly less unstable.

Issue #2043, opened 2018-01-26
Add support for onReindexStart, onReindexDone signals in chain::database


vogel76:
Plugins which hold data in own storage will need to know when it shall be cleared/rebuild.

mvandeberg:
This is a task that needs to be handled in a more general manner for the modular blockchain fabric, but for the purposes of the RocksDB AH Plugin, simply adding these signals is probably the simplest implementation.

We'll refactor later with a better solution.

PR #2081, opened 2018-02-06
add signals for on-reindex-start and on-reindex-done events

mkochanowicz:
Issue #2043

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -30,6 +30,8 @@
 
 #include <fc/io/fstream.hpp>
 
+#include <boost/scope_exit.hpp>
+
 #include <cstdint>
 #include <deque>
 #include <fstream>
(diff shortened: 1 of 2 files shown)


libraries/chain/database.cpp
```diff
@@ -153,9 +155,17 @@ void database::open( const open_args& args )
 
 uint32_t database::reindex( const open_args& args )
 {
+   bool reindex_success = false;
+   uint32_t last_block_number = 0; // result
+
+   BOOST_SCOPE_EXIT(this_,&reindex_success,&last_block_number) {
```
mvandeberg: Is there a use case, today, for a plugin doing something different on reindex failure? (Specifically the AH RocksDB plugin)

If not, I think this is being clever for the sake of being clever and the reindex success signal should be used when the reindex succeeds.

vogel76: Reindex-done signal shall be always sent to avoid reconfiguration of a plugin after finished reindex (RocksDB-AH plugin needs it to change data flushing strategy).
It can be also helpful for generating a report of time spent in plugin during reindex process.

Issue #2067, opened 2018-01-31
claim_reward_balance2_operation reward_tokens member needs changed


theoreticalbts:
This ticket is about PR #1997.

The `claim_reward_balance2_operation` has a field `flat_map< asset_symbol_type, asset > reward_tokens;`

The code assumes for any `item` in `reward_tokens`, `item.first == item.second.symbol`, but this is not enforced anywhere.

Additionally, the binary serialization is inefficient, since each element's symbol must be serialized twice.  A couple ways to fix this:

- Use `flat_map< asset_symbol_type, share_type >` instead
- Use `vector< asset >` instead, and have `validate()` enforce `reward_tokens[i].symbol > reward_tokens[i-1].symbol` for all `i > 0`

mvandeberg:
Let's use a `flat_set` with a custom comparator that only sorts by asset symbol type.

Put the comparator in `asset.hpp` so we can use it elsewhere.

vogel76:
We did try to use `flat_set` with custom comparison op, what is natural solution for this, but it is impossible because of bad implementation of `pack`/`unpack` methods operating on flat_set in fc. The problem is, that they can match only containers (**this problem is specific to all containers having implemented serialization there**) when other args than held type(s) are default ones. If we will specify flat_set<asset, asset_symbol_less> it won't match, because template functions `pack`/`unpack` have missing other template arguments.
To avoid such big change in fc, we chosen less complicated way and used a map.

mvandeberg:
These challenges should have been discussed before choosing a suboptimal implementation. We have our own fork of `fc`. We can make whatever changes we need to support `steemd`. `steemd` should never suffer because don't want to modify `fc`.

theoreticalbts:
> Let's use a flat_set with a custom comparator that only sorts by asset symbol type

No, no, please no.

> flat_set with custom comparison op, what is natural solution for this

Maybe in C++, but this construction will make porting to other languages more difficult.

Essentially you're saying "This is a container of `asset` where each asset's symbol is unique."  Which breaks the abstraction layer between the container and the contained object.  In other words, you want to have a special container that "knows" how to dig into the `asset` data structure and extract the `symbol`.  You're implementing business logic constraints in the container type.

The conceptually simpler way is to simply use a `vector` and *in the validator* enforce that the container elements are sorted and distinct.

The code to enforce uniqueness belongs in the validator.  It feels very wrong architecturally to put it in a comparator.

youkaicountry:
[...]

PR #2082, opened 2018-02-06
2nd take on claim_reward_balance2_operation creation

Kiwonik:
As requested `reward_tokens` container was changed to `vector` and a separate comparator was added to `asset.hpp` #1896 #2067

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -2213,7 +2213,7 @@ void claim_reward_balance2_evaluator::do_apply( const claim_reward_balance2_oper
             FC_ASSERT( a != nullptr, "Could NOT find account ${a}", ("a", op.account) );
          }
 
-         if( is_asset_type( token, VESTS_SYMBOL) ) // VESTS here
+         if( token.symbol == VESTS_SYMBOL)
          {
             FC_ASSERT( token <= a->reward_vesting_balance, "Cannot claim that much VESTS. Claim: ${c} Actual: ${a}",
                ("c", token)("a", a->reward_vesting_balance) );   
@@ -2243,7 +2243,7 @@ void claim_reward_balance2_evaluator::do_apply( const claim_reward_balance2_oper
 
             _db.adjust_proxied_witness_votes( *a, token.amount );
          }
-         else // STEEM & SBD here
+         else if( token.symbol == STEEM_SYMBOL || token.symbol == SBD_SYMBOL )
          {
             FC_ASSERT( is_asset_type( token, STEEM_SYMBOL ) == false || token <= a->reward_steem_balance,
                        "Cannot claim that much STEEM. Claim: ${c} Actual: ${a}", ("c", token)("a", a->reward_steem_balance) );
@@ -2252,6 +2252,8 @@ void claim_reward_balance2_evaluator::do_apply( const claim_reward_balance2_oper
             _db.adjust_reward_balance( *a, -token );
             _db.adjust_balance( *a, token );
          }
+         else
+            FC_ASSERT( false, "Unknown asset symbol" );
       } // non-SMT token
    } // for( const auto& token : op.reward_tokens )
 }
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -2197,9 +2197,8 @@ void claim_reward_balance2_evaluator::do_apply( const claim_reward_balance2_oper
 {
    const account_object* a = nullptr; // Lazily initialized below because it may turn out unnecessary.
 
-   for( const auto& token_entry : op.reward_tokens )
+   for( const asset& token : op.reward_tokens )
    {
-      const asset& token = token_entry.second;
       if( token.symbol.space() == asset_symbol_type::smt_nai_space )
       {
          _db.adjust_reward_balance( op.account, -token );
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -2199,6 +2199,9 @@ void claim_reward_balance2_evaluator::do_apply( const claim_reward_balance2_oper
 
    for( const asset& token : op.reward_tokens )
    {
+      if( token.amount == 0 )
+         continue;
+         
       if( token.symbol.space() == asset_symbol_type::smt_nai_space )
       {
          _db.adjust_reward_balance( op.account, -token );
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -2192,6 +2192,72 @@ void claim_reward_balance_evaluator::do_apply( const claim_reward_balance_operat
    _db.adjust_proxied_witness_votes( acnt, op.reward_vests.amount );
 }
 
+#ifdef STEEM_ENABLE_SMT
+void claim_reward_balance2_evaluator::do_apply( const claim_reward_balance2_operation& op )
+{
+   const account_object* a = nullptr; // Lazily initialized below because it may turn out unnecessary.
+
+   for( const auto& token_entry : op.reward_tokens )
+   {
+      const asset& token = token_entry.second;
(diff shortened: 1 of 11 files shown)


libraries/protocol/steem_operations.cpp
```diff
@@ -631,6 +631,39 @@ namespace steem { namespace protocol {
...
+   }
+
+   void claim_reward_balance2_operation::validate()const
+   {
+      validate_account_name( account );
+      FC_ASSERT( reward_tokens.empty() == false, "Must claim something." );
+      FC_ASSERT( reward_tokens.begin()->amount >= 0, "Cannot claim a negative amount" );
+      for( auto itl = reward_tokens.begin(), itr = ++itl; itr != reward_tokens.end(); ++itl, ++itr )
+      {
+         FC_ASSERT( itl->symbol.to_nai() <= itr->symbol.to_nai(),
```
mvandeberg: We should make it standard practice that whenever an NAI is used in an assertion that the nai strings be returned in the error message. @youkaicountry We should create an issue to audit all SMT related error messages prior to release to enforce this standard.

libraries/protocol/steem_operations.cpp
```diff
@@ -631,6 +631,39 @@ namespace steem { namespace protocol {
       FC_ASSERT( reward_steem.amount > 0 || reward_sbd.amount > 0 || reward_vests.amount > 0, "Must claim something." );
    }
 
+#ifdef STEEM_ENABLE_SMT
+   void claim_reward_balance2_operation::add_reward_token( const asset& reward_token )
```
mvandeberg: I suppose this is useful for testing, but it doesn't belong on the operation. As far as I can tell, it isn't being used anywhere outside of the tests and will not be called in the normal validation/evaluation of the operation.

In fact, this is very bad. The function is only used in testing and is tested in testing. All operation testing should be done using `validate` and by pushing the transaction to the blockchain.

libraries/protocol/steem_operations.cpp
```diff
@@ -631,6 +631,39 @@ namespace steem { namespace protocol {
...
+                 "Duplicate symbol ${s} inserted into claim reward operation container.", ("s", reward_token.symbol) );
+      reward_tokens.insert( insert_it, reward_token );
+   }
+
+   void claim_reward_balance2_operation::validate()const
+   {
+      validate_account_name( account );
+      FC_ASSERT( reward_tokens.empty() == false, "Must claim something." );
+      FC_ASSERT( reward_tokens.begin()->amount >= 0, "Cannot claim a negative amount" );
+      for( auto itl = reward_tokens.begin(), itr = ++itl; itr != reward_tokens.end(); ++itl, ++itr )
```
mvandeberg: I believe `itr = ++itl` will result in `itr == itl` evaluating to true to start the loop, which should instantly fail `FC_ASSERT( itl->symbol.to_nai() <= itr->symbol.to_nai() )`.

libraries/protocol/include/steem/protocol/asset.hpp
```diff
@@ -85,6 +85,22 @@ namespace steem { namespace protocol {
       }
    };
 
+   /** Helper comparator facilitating asset (or asset_symbol_type)
+    *  storage sorted by symbol's nai.
+    */
+   struct nai_less
```
mvandeberg: This comparator is not needed since `add_reward_token` is not needed.

If we want to abstract anything to be re-usable, it would be the validation of `vector< asset >` done in `claim_reward_balance2_operation`.

tests/tests/smt_operation_tests.cpp
```diff
@@ -924,5 +924,110 @@ BOOST_AUTO_TEST_CASE( smt_limit_order_create2_apply )
...
+      op.get_required_active_authorities( auths );
+      BOOST_REQUIRE( auths == expected );
+
+      expected.insert( "alice" );
+      op.get_required_posting_authorities( auths );
+      BOOST_REQUIRE( auths == expected );
+   }
+   FC_LOG_AND_RETHROW()
+}
+
```
mvandeberg: There is no `claim_reward_balance2_apply` test. At the very least, it should contain the identical test for STEEM, SBD, and VESTS as `claim_reward_balance_apply` to verify existing behavior is not broken.

libraries/chain/steem_evaluator.cpp
```diff
@@ -2192,6 +2192,71 @@ void claim_reward_balance_evaluator::do_apply( const claim_reward_balance_operat
...
+
+               gpo.pending_rewarded_vesting_shares -= token;
+               gpo.pending_rewarded_vesting_steem -= reward_vesting_steem_to_move;
+            });
+
+            _db.adjust_proxied_witness_votes( *a, token.amount );
+         }
+         else // STEEM & SBD here
+         {
+            FC_ASSERT( is_asset_type( token, STEEM_SYMBOL ) == false || token <= a->reward_steem_balance,
```
mvandeberg: This block assumes that the token must be STEEM or SBD. However, there are asset nums not in the SMT space that are not core assets. They are not defined asset types, but could be input in the op as a non-existent asset. The validation of the op cannot check the existence of assets in the database like the old op could, so we need to be even more careful when writing our evaluators to not make such assumptions about the input. I think this block is doing premature optimization and resulting in an implementation that is both confusing and wrong.

libraries/protocol/steem_operations.cpp
youkaicountry: I like that idea, I'll enshrine this in an issue we can execute later.

libraries/chain/steem_evaluator.cpp
mvandeberg: I do think this would fail when `adjust_reward_balance` is called. However, I think it is best not to rely on exceptions being thrown to do input verification.

Kiwonik: Fixed. A separate assertion is triggered when unknown asset type is found.

libraries/protocol/include/steem/protocol/asset.hpp
Kiwonik: Removed.

libraries/protocol/steem_operations.cpp
Kiwonik: Removed.

tests/tests/smt_operation_tests.cpp
Kiwonik: No, there is no evaluation tests because as far as I know both evaluation and its testing were outside the scope of the original issue (which was https://github.com/steemit/steem/issues/1896 ). That was brought to my attention after I had already had the validator implemented so I decided to leave it as is and created original PR (which was https://github.com/steemit/steem/pull/1997 ). Then @theoreticalbts decided to create another issue (for reasons unknown to me) and so we're here. Shall we expand this issue/PR to create the evaluation test too, I will.

libraries/protocol/steem_operations.cpp
Kiwonik: That's a nasty bug, that slipped through due to insufficient coverage of validation test. I added appropriate cases and fixed the bug.

tests/tests/smt_operation_tests.cpp
mvandeberg: @youkaicountry Your call.

mvandeberg: However, because there is an implemented evaluator, I expect it to be tested. If the evaluator is out of scope of this issue, then it should be removed from this PR and added to another.

It is too easy to see the implemented evaluator and assume the other issue is complete when it is not because of a lack of a test.

mvandeberg: I don't want to delay merging of this PR, but I don't want to merge blatantly untested code.

Kiwonik: @mvandeberg @youkaicountry I committed `claim_reward_balance2_apply` test that contains both legacy symbols and SMT testing copying in scope what original `claim_reward_balance_apply` does. Now that the `claim_reward_balance2_evaluator` is being tested, this PR is ready to be merged, pending CI pass.

PR #2089, opened 2018-02-08
legacy network broadcast

### LINKED ISSUE #2084: Appbase Legacy Network Broadcast
On transfer broadcast in appbase: `RPCError: Bad Cast:Invalid cast from string_type to Array`

This is because it is using the new operation deserializer, not the legacy. Another visitor needs to be added to convert from legacy to normal operations.

Along with this, we should move the implementation of `broadcast_transaction_synchronous` to `condenser_api` only and plan on implementing the transaction confirmation API (#486) as part of the new API changes. This will force us to move away from synchronous broadcasts.

mvandeberg:
Closes #2084

### DIFF
--- a/libraries/plugins/apis/condenser_api/condenser_api.cpp
+++ b/libraries/plugins/apis/condenser_api/condenser_api.cpp
@@ -1588,7 +1588,7 @@ namespace detail
       CHECK_ARG_SIZE( 1 )
       FC_ASSERT( _network_broadcast_api, "network_broadcast_api_plugin not enabled." );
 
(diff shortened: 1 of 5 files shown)


theoreticalbts:
Looks functionally correct but a little verbose.  Suggestion to make it shorter #2094.

mvandeberg:
Any quick to implement suggestions are welcome. I will remind you once again that this API is intended to be deprecated and removed as soon as we can reasonably do so. I will fully admit that maintainability was not a priority when writing this code.

PR #2092, opened 2018-02-08
Train chain_plugin writes using single thread io service

### LINKED ISSUE #2090: Chain Plugin write race condition
Pre-appbase all writes were executed in a single thread using cooperative multi-tasking. The read-write lock implementation used by chainbase uses this assumption. However, in Appbase writes can come in from the `network_broadcast_api`. Under heavy load this can lead to race conditions surrounding the write lock. If a block takes longer than 1 second to apply, the write lock could be moved forward by the pending transaction and two writers could enter critical sections at the same time. To avoid this the chain_plugin should explicitly train writes using an io_service with only one thread.

mvandeberg:
Closes #2090

Running locally without problems.

### DIFF
--- a/libraries/plugins/chain/chain_plugin.cpp
+++ b/libraries/plugins/chain/chain_plugin.cpp
@@ -6,6 +6,12 @@
 
 #include <fc/string.hpp>
 
+#include <boost/asio.hpp>
+#include <boost/optional.hpp>
+#include <boost/bind.hpp>
+#include <boost/preprocessor/stringize.hpp>
+#include <boost/thread/future.hpp>
+
 #include <iostream>
 
 namespace steem { namespace plugins { namespace chain {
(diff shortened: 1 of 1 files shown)


theoreticalbts:
In `chain_plugin::accept_block()`, you call `get()` on a `boost::future` which is blocking.  But `chain_plugin::accept_block()` is in turn called by `p2p_plugin_impl::handle_block()`, which is a `node_delegate` method called by the p2p code.  I *think* the multithreading stuff that happens as a result is all okay (basically it blocks anything else fc async'ed in the p2p code, but that's all right because all the future does `push_block()`, which itself shouldn't block.  So the handler should always make progress as fast as the CPU can go and there's no potential for deadlock, and while lock contention can happen and blocks all p2p activity, it should never take long enough for this to be a serious problem.

But I'm not 100% sure about all that.

The above text should probably be a comment near the `get()` call.

theoreticalbts:
Also, `allow_future_time` -> `allow_promure_time` -- what is up with that?  Typo or wayward find/replace?

mvandeberg:
Wayward find/replace. I'll fix

mvandeberg:
Fixed and squashed

mvandeberg:
I think the way the p2p code is working is that `handle_block` always happens separate from p2p maintenance. I do know the p2p code operates in a separate thread and that pre-appbase there was no race conditions because fc async handled the cooperative multi-tasking. If we end up having performance issues, this interaction will be one of the first things we look at.

PR #2093, opened 2018-02-08
Move check_max_block_age short-circuit check out of lock

### LINKED ISSUE #2091: Broadcast_transaction_api check_max_block_age critical section
The critical section of code for this function is too big and is leading to unnecessary locking.

``` c++
if( max_block_age < 0 )
   return false;
```

Does not read from the database and does not need to be locked.

mvandeberg:
Closes #2091


theoreticalbts:
Short and simple, LGTM

Issue #2056, opened 2018-01-29
Create smt_refund_operation


theoreticalbts:
Basically this is just `smt_claim_operation` with an added `smt` field to specify which SMT (as the `amount` must be an amount of STEEM):

```
struct smt_refund_operation
{
   account_name_type     executor;
   account_name_type     contributor;
   contribution_id       contribution_id;
   asset_symbol_type     smt;
   asset                 amount;
   extensions_type       extensions;
};
```

See #2054 for description of `smt_claim_operation`.

Kiwonik:
@theoreticalbts @youkaicountry Today I discovered a possibility of malicious attack on some SMT ICOs enabled by introduction of `executor` idea in this operation.
When the ICO is in `post_launch_phase` and enough contributions were declared but SMT creator still delays some cap reveal, the contributors may wait until `launch_expiration_time` or may issue refund operation, that may in turn result in ICO failure (when total contributions drop below min cap).
Now a malicious person interested in ICO failure may issue these `smt_refund_operation` against the will of the contributors but in their stand using the `executor` field.
I hope that the whole idea of executor is obsolete per my recent comments in https://github.com/steemit/steem/issues/2054 but if it turns out otherwise, this issue needs implementation that blocks the attack possibility.

youkaicountry:
Closing as outdated.

PR #2109, opened 2018-02-12
Added smt_refund_operation with validation.

Kiwonik:
Also added validation/authorities tests. #2056


mvandeberg:
Rebased on `develop` and pushed.

Issue #2100, opened 2018-02-09
Update Docker for RocksDB


mvandeberg:
#2066 will need to update our Dockerfile to build with RocksDB.

@vogel76 will document/link me to build instructions for RocksDB integration and I will update the Dockerfile.

vogel76:
All needed dependencies has been added to Dockerfile, but in the branch specific to the issue #2066 to make possible merging it into develop. 
This issue is done.

Issue #2099, opened 2018-02-09
support exfiltration of virtual ops via json rpc without current AH implementation


sneak:

sneak:
clarity: exfiltration via json-rpc response - so a new api call or modified get_ops_in_block api call.

vogel76:
Implemented account_history_api.enum_virtual_ops method taking following parameters:
- block_range_begin - number of block (inclusive), starting the block range, the virtual ops should be iterated from
- block_range_end - number of block (exclusive), ending the block range, the virtual ops should be iterated from
So the call:
```
{"jsonrpc": "2.0", "method": "call", "params": ["account_history_api","enum_virtual_ops",{"block_range_begin": 1, "block_range_end": 100}], "id": 2}
```
will retrieve all virt. ops existing in blocks [1..100).
Method returns pair: collection of operations (as `ops` member) and number of next block starting block range containing any virt. op as `next_block_range_begin`. If there is no more virt. ops `next_block_range_begin` will be set to 0.
Example result set: 
```
{
	"jsonrpc": "2.0",
	"result": {
		"ops": [
			{
				"trx_id": "0000000000000000000000000000000000000000",
				"block": 1,
				"trx_in_block": 0,
				"op_in_trx": 0,
				"virtual_op": 0,
				"timestamp": "2016-03-24T16:05:00",
				"op": [
					"producer_reward",
					{
						"producer": "initminer",
						"vesting_shares": [
							"1000000",
							6,
							"@@000000037"
						]
					}
				]
			}
		],
		"next_block_range_begin": 1093
	},
	"id": 2
}
```

vogel76:
@youkaicountry This probably can be closed as this code has been merged by @mvandeberg into develop.

Issue #2066, opened 2018-01-31
Enhance AH rocksdb-prototype to cover all exiting functionality of account-history plugin


vogel76:
Add support for:
- account white-listing, black-listing
- operation filtering 
- data pruning to cover all exiting functionality of account-history plugin

vogel76:
Added previously missing functionality:
- account filtering
- operation filtering
- AH data pruning
Needed changes did force redesign of data structures and it lead to small slowdown according to golden results initially taken (up to 19M blocks):
- reindex process is a little longer (reindex slowdown is near to 10%) and takes now 6678s (6065s originally, and 13328 in develop) what is caused by directly reading data while writing them into db.
- reindex process is significantly longer when prune is enabled and takes  10411s. Comparison will be added when reindex + prune will complete on develop version (it is longer than no-prune).
- memory allocation is stable, depends on cache configuration.
To make PR, docker build configuration must be fixed, what will be next step to be done.

vogel76:
As a side note, prune option by default shall be *disabled* now. It damages historical data, makes data collection longer. So it makes no sense to have it enabled by default

PR #2129, opened 2018-02-16
2066 supplement ah features, virtual ops enumeration

vogel76:
Issues #2066 #2099 #2100


libraries/plugins/account_history/account_history_plugin.cpp
```diff
@@ -206,13 +206,13 @@ void account_history_plugin::set_program_options(
...
-         ("account-history-whitelist-ops", boost::program_options::value< vector< string > >()->composing(), "Defines a list of operations which will be explicitly logged.")
-         ("history-whitelist-ops", boost::program_options::value< vector< string > >()->composing(), "Defines a list of operations which will be explicitly logged. Deprecated in favor of account-history-whitelist-ops.")
-         ("account-history-blacklist-ops", boost::program_options::value< vector< string > >()->composing(), "Defines a list of operations which will be explicitly ignored.")
-         ("history-blacklist-ops", boost::program_options::value< vector< string > >()->composing(), "Defines a list of operations which will be explicitly ignored. Deprecated in favor of account-history-blacklist-ops.")
-         ("history-disable-pruning", boost::program_options::value< bool >()->default_value( false ), "Disables automatic account history trimming" )
+         ("ah-account-history-track-account-range", boost::program_options::value< vector< string > >()->composing()->multitoken(), "Defines a range of accounts to track as a json pair [\"from\",\"to\"] [from,to] Can be specified multiple times.")
```
mvandeberg: The pattern we are trying to use for all options is `plugin-name-option`. Prefixing this breaks the pattern because the plugin name is `account-history`, not `ah`. I also think that for the time being it is best to make the rocksdb options different so that the decision to use rocksdb is purposeful and not accidental.

vogel76: I have renamed such options (in legacy account-history plugin) because they conflicted to these defined in rocksdb plugin. Since we want to provide replacement for original AH plugin, I still think that is better to keep existing names of options for backward compatibility. That will make switch very easy at customer installations.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "libraries/fc"]
 	path = libraries/fc
 	url = https://github.com/steemit/fc.git
+[submodule "libraries/plugins/rocksdb/rocksdb"]
+	path = libraries/plugins/rocksdb/rocksdb
+	url = https://github.com/vogel76/rocksdb.git
```
sneak: even if we were going to use a submodule (we're not using submodules or anything of the sort), pointing at a private fork in a personal account is a hard no.

vogel76: The fork has been created because I had to fix one bug in rocksdb/CMakeLists.txt which disallow using this project as part of other project. And this is not private repo, but public one. Made changes were accepted and merged to the rocksdb master branch, but still didn't appear in any release.
We can of course make dedicated steem repo for it (even I don't see good reason for it because it is rather temporary state - until change will be published in regular rocksdb release), but we still need a way to integrate it to the main steem project (so use rejected by you git subtree or git subrepo).

mvandeberg:
I cannot merge this until rocksDB is vendored.

.gitmodules
sneak: I think you misunderstood me.

No submodules.

No subtrees.

**Please vendor it in.  ASAP. **

(This also solves the other problem you describe (needing a local edit that's not in an upstream release.))

sneak:
I think perhaps that the new rocksdb-based AH plugin should be given a new name, so that people can enable the new-style rocksdb-based plugin or the old-style AH plugin with a small list of whitelisted ops or account names and continue using steemd like they did before.

mvandeberg:
There is some cleanup that needs to happen in the plugin. We can rename and deal with config type changes in another issue. But let's not postpone merging for such changes. @sneak As it stands, this branch now has RocksDB vendored in. I believe it can be merged to develop in the current state.

mvandeberg:
@theoreticalbts @youkaicountry I did some work on this branch. Can one of you review it please?

Issue #1839, opened 2017-12-03
Improve SBD to USD price parity


bitcoiners:
There are currently only about 3.5 million of SBD, so it does not take much money to pump price and break USD parity, as has been happening pretty regularly lately.
In order to improve SBD to USD price parity we need to be able to borrow (create) SBD at will, by putting up Steem as collateral, similar to how it is done in Bitshares when BitUSD is created. 
This will help SBD keep price parity to USD as more will be created and sold if SBD gets more expensive that USD.

chenlocus:
This might be a short term surge of SBD,  is it necessary to increase the supply of SBD at once?  Why not wait for a period, e.g. if the SBD keeps on above 10% of USD for 1 month, then take actions?

joticajulian:
I think this would be good. because the SBD must be pegged to USD. In this moment there is a lot of volately.
In this post I talk about the same, a proposal for the SBD, we need that!!
https://steemit.com/utopian-io/@jga/steem-proposal-possibility-to-transform-from-steem-to-sbd-in-order-to-reduce-speculation

bitcoiners:
Here is another proposal on how to improve SBD to USD peg.
https://steemit.com/sbd/@timcliff/saving-sbd-proposal-to-restore-the-usd1-usd-usd1-sbd-peg

iamsmooth:
Some of these proposals are controversial and there have been claims these proposed new mechanisms could be vulnerable to manipulation.

In addition we might consider changes to the passive elements of SBD creation that might help.

The basic idea being that an SBD shortage (SBD surplus is easily handled by the existing conversion mechansim) is what causes the peg to fail, there are some more incremental ideas which can be used to improve the peg by increasing the supply of SBD in various ways:

1. An an option for post payouts in 100% SBD (or 90% SBD, which could be favored because of the desire to retain a vested STEEM incentive component). When SBD is overvalued and in short supply authors will tend to choose this, increasing total SBD payouts.
2. Pay out curation rewards in the same mix as the post (so when authors choose 50/50 or the hypothetical 100%/90% SBD from #1 above, curators would receive the same mix, increasing SBD supply when needed.
3. Pay out beneficiaries in the same mix as the post. Currently beneficiaries are always paid in 100% SP. As the beneficiaries feature has become more used, this has reduced the supply of SBD as an undesirable side effect. With increasing usage of the feature, this could get worse.
[...]

PR #2135, opened 2018-02-19
Remove reduced SBD print rate. Set print rate to constant 100% (50% SBD / 50% STEEM)

netuoso:
Explanation by @TimCliff 

> After the SBD passes 2% of the STEEM marketcap, the blockchain starts producing STEEM for a portion of the author rewards instead of SBD. It does more STEEM and less SBD linearly up until 5% debt, at which point it is doing 100% STEEM instead of SBD.
> 
> At 3.5% debt a post would pay out:
> 50% SP
> 25% SBD
> 25% STEEM
> 
> At 5% debt a post would pay out:
> 50% SP
> 50% STEEM (edited)

This issue will contain the necessary changes to keep SBD print rate at max regardless of the debt ratio


mvandeberg:
This is a pretty major economic change to SBD and should be discussed in an issue _before_ creating a PR.

mvandeberg:
This PR is also not mergeable as is because it change SBD payouts on a HF19 check which would change payouts in the past. Even if this patch managed to reindex (I think that is highly unlikely) it has the potential to fork the blockchain.

roadscape:
There's been a fair amount of discussion around this in chat but it does need to be publicly documented/evaluated. @netuoso did you already file an issue? If so please link to it from the PR.

Also https://github.com/steemit/steem/issues/1839 touches on this topic.

mvandeberg:
I think removing the debit cap is the wrong approach to the SBD print rate.

A few thoughts that might constitute a more detailed proposal:

1. The debt cap has served a useful purpose. 

2. Right now, the debt cap is not doing anything. Removing it would have no immediate impact on the creation of SBD.

3. It is relatively easy to deflate SBD through burning of rewards/converting to STEEM and difficult to print. This leads to a naturally deflationary policy.

4. Steem's current inflation rate is 8.98%. Of that, there is an effective 2.5% going to printing SBD. If the value of STEEM increases faster than 2.5% annually (we are all hoping that it does) then the debt percentage drops.

5. Instead of removing the debt cap to allow for the increase in SBD supply (because this is likely to do nothing in the short term), I would propose increasing the SBD print rate below a certain debt percent.

6. Create a new witness parameter, `sbd_desired_debt_ratio`. We would leave the current debt cap in at some cap of the ratio with the current behavior. When the debt ratio falls below the the voted ratio, we begin printing SBD faster.


Based on the price of SBD, there seems to be a debt ratio that the market is comfortable sustaining. With a valuation of $5 per SBD, that ratio would seem to be around 5%. This would indicate that our hard cap of 5% is too low and should be increased, but the logic should not be thrown out. Based on the inflation rates about, we would only expect ~25m SBD to be printed in the next year. This only increases the debt percentage to 3.5% (assuming STEEM is price stable for an entire year).

If such a system were implemented, I would recommend all witnesses begin reporting the actual price feed (not an adjusted price feed) and slowly change the desired debt ratio until SBD is price stable again.

netuoso:
I have been participating in the discussions that have been taking place. Notably with @iamsmooth, @timcliff, and others. I agree having the main of the chat taking place now on a Github issue makes sense. (But I had to see if there was a desire for this work before taking even that next step)

Making the issue and documenting the quote explanation from @TimCliff took about just as long as making the actual changes here in the PR. I have been researching and studying the Steem source code for a few weeks now and am starting to get a grasp on the way the project is organized.

To your point though, I believe having some code provided with the issue is far more useful than the issue being entirely theoretical in discussion. With this format, you were able to make more pointed comments about my intended code changes that might help improve my process for this issue and others in the future.

I want to make one final point ... I am not an economist or a financial modeler. I have no experience in such a field. I can only take the theoretical discussion here and help turn it into actual code. Please do not look to me to suggest what numbers to change the print rate to unless you also provide a ton of information that can help me as a developer understand the field in question.

I already linked this issue/PR in the dev slack but I will do it again and ask for more discussion.

@roadscape This PR is also the Issue. I utilize `hub` by Github to convert an Issue into a PR in an effort to retain the discussion from the Issue on the PR.

@mvandeberg Is it proper on this repo to convert issue to PR or does your process prefer an issue and a separate PR. The benefit to having the separate is if someone wanted to make another PR at the same time as this one ... but in that case, I would highly advice the interested dev to reach out to me so I can grant them editing access to this particular PR (assuming it isnt just way off in left field)

mvandeberg:
Our process is to discuss in an issue (which will result in a defined spec) and then only discuss the implementation in the PR.

netuoso:
@mvandeberg These chats have been taking place for months in the dev slack.

Ill make a new issue so they can be duplicated here.

Please dont act like I did this on my own interest without consensus from others. The process I specifically asked you about was if you wanted me to:
1. Create an Issue
2. Discuss it
3. Create a NEW PR that links to issue in step 1

OR

1. Create an issue
2. Discuss it
3. Use hub to convert the Issue to a PR as i have seen many other PRs from the steemit org follow.
 (and is the exact same process I have used in my previous jobs)

---

This page for all intents and purposes is an issue, and now discussion is taking place on it. I most definitely did not assume the PR was going to be thumbed up and merged without further chats.

If SteemIt wants more community devs on the actual source code, the learning has to start somewhere. By providing the code example relating to this PR, I figured that learning had already started.

If a PR has to be closed to better address it in another, that is not time wasted, that is time learned.

Please remember SteemIt, INC doesn't have any kind of class or anything to teach community devs what process is expected. As long as a community dev follows the guidelines as best as possible, it might be wise to work with them to fit your desired format.

TimCliff:
Two comments:
1. I support moving the discussion to an issue until there is consensus on what we are going to code.
2. The original change that was discussed in slack was much smaller scope than what we are discussing here, and what got coded. 

The intended change was to remove the gradual reduction in SBD printing, and just have a hard switch between SBD or STEEM. This could either be at the current 5% level, or moved to 10%.

No changes to the 10% debt level were within the scope of what we discussed, and there is agreement that with that 10% level in place - continuing to print SBD after it reaches that point would not be advised.

netuoso:
Recreating this issue to contain further discussion.

PR #2145, opened 2018-02-21
Vendor fc

### LINKED ISSUE #1301: Vendor FC
Should we vendor in FC to steem?

We will be making major changes to it that will differentiate it from cryptonomex/fc soon. Doing so protects it with the same branch protections that we have now and allows commits that change both steemd and fc code simultaneously rather than a steemd commit followed by an fc submodule bump and requiring PRs for both.

Thoughts?

mvandeberg:
Closes #1301


mvandeberg:
I checked out fc separately and ran diff between Steem's vendored fc and standalon fc. Here is my output:

```
Files ./fc/.DS_Store and ../../fc/.DS_Store differ
Files ./fc/vendor/.DS_Store and ../../fc/vendor/.DS_Store differ
Files ./fc/vendor/diff-match-patch-cpp-stl/.git and ../../fc/vendor/diff-match-patch-cpp-stl/.git differ
Files ./fc/vendor/secp256k1-zkp/.git and ../../fc/vendor/secp256k1-zkp/.git differ
Files ./fc/vendor/websocketpp/.git and ../../fc/vendor/websocketpp/.git differ
```

`.DS_Store` is an OS X specific setting file and we would expect the `.git` files to be different. All of the source code is identical.

Issue #1996, opened 2018-01-16
Extend reindex-process benchmark by collection statistical and execution time informations related to operations processed for each visited block


vogel76:
Statistic and time data should be collected for each processed operations. Report should be dumped in json format similary to reindex benchmark collector.
Best collect operation processing times in distribution to plugins which processed them, to generate report contatining time distribution for each plugin.

mkochanowicz:
Blocked until accepted https://github.com/steemit/fc/pull/40

mvandeberg:
fc is vendored to steemd. fc changes are now a part of the steemd PR. This is no longer blocked.

PR #2149, opened 2018-02-22
1996 bench

mkochanowicz:
Issue #1996


libraries/legacy_plugins/account_statistics/account_statistics_plugin.cpp
```diff
@@ -75,7 +75,7 @@ void account_statistics_plugin::plugin_initialize( const boost::program_options:
    {
       ilog( "account_stats plugin: plugin_initialize() begin" );
 
-      database().post_apply_operation.connect( [&]( const operation_notification& o ){ _my->on_operation( o ); } );
+      database().post_apply_operation_proxy( [&]( const operation_notification& o ){ _my->on_operation( o ); } );
```
mvandeberg: No, we should not be changing interfaces. It is ugly to require changing interfaces to support this. If we need a different interface, we should discuss changing that as another task.

vogel76: @mvandeberg The biggest problem is that most of member functions and variables defined in your code is public what is troublesome from interface maintenance point of view. Then I think we shall define public methods to hide internals. In this case, it was needed to be sure that every plugin has been correctly connected to the database signals. This allows us to centralize time-measurement when callbacks are executed by each plugin code, next to print time-distribution report gathered during reindex process.  
Also I think it is good time to change plugin interfaces because of AppBase release, which by design changed plugin structure and interfaces.

mvandeberg:
@mkochanowicz Please rebase with vendored fc

libraries/legacy_plugins/account_statistics/account_statistics_plugin.cpp
mvandeberg: I'll make an issue to clean up these interfaces. Each plugin has an abstract name() method. We should be using that instead of each plugin specifying manually. Perhaps in a macro.

mkochanowicz:
@mvandeberg It's done.

libraries/legacy_plugins/account_statistics/account_statistics_plugin.cpp
vogel76: Maybe best way would be to pass into proxy methods (i.e. pre_apply_operation_proxy) plugin instance (and change `const std::string& plugin_name` argument to  `const abstract_plugin& plugin`. Then inside proxy method plugin name can be retrieved. This is also more flexible from future point of view, since it could be simpler to use other abstract_plugin data during registration without changing the interface.

mkochanowicz:
Change plugin_name into plugin_instance as @vogel76 suggested.

libraries/chain/database.cpp
```diff
@@ -3010,10 +3012,160 @@ void database::apply_operation(const operation& op)
...
+      _func(std::forward<TArgs>(args)...);
+
+      if (_benchmark_dumper.is_enabled())
+         _benchmark_dumper.end(_name);
+   }
+
+private:
+   TNotification                    _func;
+   util::advanced_benchmark_dumper& _benchmark_dumper;
+   const abstract_plugin&           _plugin;
```
mvandeberg: From what I can tell, this isn't needed anywhere. Is there a reason we are saving the reference?

mkochanowicz: The reason is: for future uses - but ofkoz you are right here - I removed this member.

PR #2164, opened 2018-02-27
Fix network broadcast api to use accept_block/transaction

### LINKED ISSUE #2163: Network Broadcast API does not use thread safe accept_block/transaction
Chain Plugin provides thread safe ways to push blocks and transactions. Network Broadcast API needs to use them.

mvandeberg:
Closes #2163


theoreticalbts:
LGTM

PR #2168, opened 2018-02-28
Bump websocket thread pool size to 256

mvandeberg:


mahdiyari:
What was the problem with 1024?

Issue #2075, opened 2018-02-01
Remove JSON Tuples


mvandeberg:
There are a number of places in the JSON serialization where tuples are used. This is not idiomatic JSON and should be replaced with objects.

Two examples:

The new asset serialization `["1000", 3, "0x00000021"]` should be changed to `{"amount":"1000", "precision":3, "nai":"0x00000021"}` (#2028)

Static Variant `["TAG", {...}]` to `{"tag":"TAG", "object":{...}}` (#2044)

In each place where tuples are used, their usage should remain the same in `condenser_api` and new default serializers written.

PR #2187, opened 2018-03-02
2075 json tuples

mariusztrela:
#2075


mvandeberg:
This is the wrong approach to fixing the problem. It needs to be fixed for all static variants in fc, not just operations.

PR #2195, opened 2018-03-02
Batch Write Queue

### LINKED ISSUE #2194: Batch Write Requests
Currently each write to database acquires the locks, performs the write, and releases it.

When we are live, this is not the worst (can be improved upon), but it is horribly inefficient when syncing the blockchain.

We should move to a producer/consumer model where all writes are placed in a queue and consumed by a dedicated thread that can hold the write lock for a specified amount of time and batch writes so that we minimize the amount of locking that needs to be done.

To prevent starvation of writers the write lock should be willingly released every so often only when live. When syncing, the write process can (and should) be greedy with the lock so long as there are pending writes.

mvandeberg:
Closes #2194

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -545,16 +545,13 @@ bool database::push_block(const signed_block& new_block, uint32_t skip)
    bool result;
    detail::with_skip_flags( *this, skip, [&]()
    {
-      with_write_lock( [&]()
+      detail::without_pending_transactions( *this, std::move(_pending_tx), [&]()
       {
-         detail::without_pending_transactions( *this, std::move(_pending_tx), [&]()
+         try
          {
-            try
-            {
-               result = _push_block(new_block);
-            }
-            FC_CAPTURE_AND_RETHROW( (new_block) )
-         });
+            result = _push_block(new_block);
+         }
+         FC_CAPTURE_AND_RETHROW( (new_block) )
       });
    });
 
@@ -689,10 +686,7 @@ void database::push_transaction( const signed_transaction& trx, uint32_t skip )
          detail::with_skip_flags( *this, skip,
             [&]()
             {
-               with_write_lock( [&]()
-               {
-                  _push_transaction( trx );
-               });
+               _push_transaction( trx );
             });
          set_producing( false );
       }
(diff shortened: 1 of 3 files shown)


theoreticalbts:
It appears functionally correct.  There are a few rough edges:

Using the system time to determine whether `is_syncing` doesn't quite work.  There are use cases for an idle node not caught up to the present -- for example, creating a testnet, the unit tests, or replaying to a specific block because you want issue RPC queries against a specific snapshot in time, instead of a moving target.

The lazy coder's heuristic-based solution for this problem is to busy-wait for a second [1] or so, then start sleeping in 10ms increment if you didn't see anything during the busy-wait.

A less lazy and more "correct" solution is to use a `wait()` / `notify()` threading primitive like C++11 `std::condition_variable` or its Boost equivalent.  This lets you have the semantics of "sleep for 10ms, but wake me up immediately if someone calls `notify()`," and then you always call `notify()` after loading a new element in the queue.

I personally like the lazy approach better, and it might be slightly better performance as well, since using a busy-wait means you don't to worry about a thread wakeup which goes through the kernel's thread scheduler.  You probably have a whole core devoted to the busy-wait, if you don't have enough other load on the system to peg all the cores.

You'd probably get the highest performance from re-writing the callers to do pipelining.  That is, rather than always waiting for a value to be returned before proceeding, the caller should continue about its business and put more blocks in the queue.  That way, when fully loaded, the queue will hopefully always have multiple elements and you'll hide the latency of the data handoff between threads.  The busy-wait or `wait() / notify()` code becomes irrelevant to performance because the queue will almost always have an element immediately available when you `pop()`.   But this is a pretty big architectural modification.

[1] The `1s` wait time can be arrived at a couple different ways:  It's definitely enough for some processing, inter-thread communication delays, and a couple network round trips, which means it should be an upper bound on what it takes to receive subsequent blocks over the p2p when syncing under normal conditions.  It's also 100 times the 10ms sleep time, meaning if you're waiting more than 1s for a queue element to arrive anyway, an additional worst-case 10ms latency is only at most a 1% increase in latency.

mvandeberg:
> That is, rather than always waiting for a value to be returned before proceeding, the caller should continue about its business and put more blocks in the queue.

I agree that pipeling is a better design. The point of this refactor is to improve performance without changing semantics. Blocking on the result was the existing semantic, so it is the current semantic. Changing this may have unintended side effects in the p2p code as it is not designed for pipelining writes.

This code already has some impressive performance gains, returning appbase to pre-appbase sync times if not better under some circumstances (this behavior has been observed but not easily replicated). If we have other ideas on how to improve this particular code block, let's file them as new issues. I don't expect disagreement to this proposal due to the two approvals of the PR.

PR #2200, opened 2018-03-06
Fix typos in chainbase README.md

nafest:


sneak:
@nafest Thank you so much for sending in this PR! :)

Issue #1508, opened 2017-09-06
Implement smt_setup_operation


theoreticalbts:
All `smt_setup_operation` semantics needs to be implemented.

- Fields should be copied from operation to corresponding fields `smt_token_object`
- Per-block processing should be added to check for arrival of timed events `generation_begin_time`, `generation_end_time`, `announced_launch_time`
- `smt_creation_fee` field should be removed, since it is now in `smt_elevate_account_operation`
- Implement unit tests

vogel76:
Waits for answers to questions inside:
https://github.com/steemit/steem/issues/1554

Kiwonik:
Is any backend reaction supposed to happen when SMT creator abandons its token i.e. no further smt operation follows successful `smt_setup_operation`? In other words all three dates (contribution start/end & launch) pass by with no operation triggered by creator (no contributors either).
If the answer is 'none' (I assume so) then `smt_setup_operation` is not the place to start per-block processing described in this issue - the next smt operation (including contribution attempt) is.

theoreticalbts:
From the spec:

> ICOs may specify a minimum number of STEEM units `min_steem_units`. If the ICO does not reach `min_steem_units` before `generation_end_time`, then it does not occur, and contributors become eligible for refunds.

The situation where `min_steem_units` is not reached is called a *failed launch*.  We should require `min_steem_units > 0`, so in case no contributions occur, it will be a failed launch.

The asset ID remains reserved to the SMT, at least until the end date occurs, because `smt_contribute_operation` needs a way to uniquely refer to the asset.

vogel76:
I think the asset ID being used at executed smt_create_operation shall stay reserved forever, since the control-account owner already paid for this reservation.

Kiwonik:
@theoreticalbts I didn't find any mention about _failed launch_ term in whitepaper (including the state diagram). I assume it's a kind of SMT creation phase - which would for example 
- Allow contributors for refund.
- Allow for possible re-launch attempt with another `smt_create_operation`.
Can you confirm?

@vogel76 Yes, the asset ID stays reserved with the control account per whitepaper:
> At launch_expiration_time, if the ICO has not yet launched, all contributors will be automatically refunded (with virtual operations) and the ICO will be cancelled. The symbol will remain reserved to the specified control_account. However, in order to launch the token, an smt_create_operation must be issued and the smt_creation_fee must be paid again.

mvandeberg:
[...]

PR #2208, opened 2018-03-07
1508 smt setup operation2

mariusztrela:
#1508


theoreticalbts:
Looks good, except:

- `max_supply` is not initialized (all integer fields should be default initialized with `= 0`)
- `smt_phase` inherits `unsigned char`, should be `uint8_t`

I will submit a two-line PR to correct these problems.

mvandeberg:
@mariusztrela merge conflict

mariusztrela:
It was made 2 small fixes pointed by @theoreticalbts + 1 fix resulting from `copy-paste` effect.

PR #2214, opened 2018-03-09
Renamed macOS X to macOS according to Apple's recent renaming of the OS.

frogg:
See https://developer.apple.com/app-store/marketing/guidelines/mac/


mvandeberg:
@frogg Thank you for this PR

Issue #1952, opened 2017-12-27
Review Unique Index Usage


mvandeberg:
> Review unique index usage - there shall be left only these unique indices where such uniqueness is defined by business logic. Correct item ordering shall be still still preserved by superfluous storing 'id' of the object together with actual key. Removing unneeded uniqueness can allow reduction of key definition which now must match to enforced uniqueness.

Mentioned in #1923

We need to be careful with this optimization to preserve iteration ordering on some blockchain functions. Changing the ordering should result in failed reindexing if the unit tests do not catch it (Some of the changes are subtle enough to not show up in unit tests).

vogel76:
@youkaicountry Most of work has been done. First measures shows significant reduction of mem. consumption:
- on fullnode without AH there was ~9GB less memory used
- on consensus node 300MB less
We have also ideas how to improve other cases.

abitmore:
Not bad. Although I don't think 300MB of 30GB is "significant reduction".

vogel76:
Significant performance changes will appear after (finally) made decision to move from shared memory file based storage to rocksdb based one, which was not yet enough prioritized, even quite successfully done AH replacement in #2066.

theoreticalbts:
This ticket should be closed and never implemented.  It is the policy of this project that all consensus indexes must be `ordered_unique`.

> We need to be careful with this optimization to preserve iteration ordering on some blockchain functions.

Yes, we do, and that's the problem:  From a practical standpoint, that's an impossible level of "careful."

> Changing the ordering should result in failed reindexing if the unit tests do not catch it (Some of the changes are subtle enough to not show up in unit tests).

You can't assume that bugs will always be caught by reindexing.  If the business logic iterates on non-unique indexes, it is possible to get nodes with undefined state synced to the live blockchain.  I personally observed this, in production, in an early version of BitShares 2.0.  In the worst case, block producers are affected by the undefined state, leading to forks, missed blocks.  In the absolute worst case, you have a completely broken blockchain which included blocks and transactions after a certain time, but is not reindexable past that time.

> there shall be left only these unique indices where such uniqueness is defined by business logic

One problem here is that "business logic" is already very complicated, it would require a lot of review to get a comprehensive list of the indexes that are iterated on.

[...]

PR #2228, opened 2018-03-14
1952 unique index2

mariusztrela:
#1952 

**short description**

It has been made 2 changes:

- **simplification of index `by_location` in `operation_object`**

before: `ordered_unique( block, trx_in_block, op_in_trx, virtual_op, id )`
after: `ordered_non_unique( block )`

This index doesn't attend in consensus, therefore there is't any problem with potential incorrect working. This index is used only during API call.

- **removing two excess indexes in `comment_vote_object`**
before: `by_comment_voter`, `by_voter_comment`, `by_voter_last_update`, `by_comment_weight_voter`
after: `by_comment_voter`, `by_voter_comment`

Missing indexed are simulated by algorithm.

**correctness tests**
During 1 week were made hundred of thousands API calls to builds: `develop`, `1952-unique-index2`( using `pyresttest` tool ).
Comparision showed, that result from above servers are identical.

**performance tests**
- reindex( `develop` branch vs `1952-unique-index2` branch ): branch `1952-unique-index2` faster 10 minutes

- Time of 5000 different calls[ s ]( local machine, 1 thread, used tool` pyresttest` )
`list_votes;by_voter_last_update`	( `12min. 23 sec` develop ) ( `12min. 20 sec.` 1952-unique-index2 )

- Time of 4000 different calls[ s ]( local machine, 1 thread, used tool `pyresttest` )
`get_ops_in_block`	( `435s` develop ) ( `435s` 1952-unique-index2 )

**Memory consumption**
- on fullnode without AH there was ~9GB less memory used
- on consensus node 300MB less

### DIFF
--- a/libraries/plugins/apis/account_history_api/account_history_api.cpp
+++ b/libraries/plugins/apis/account_history_api/account_history_api.cpp
@@ -5,8 +5,6 @@
 
 namespace steem { namespace plugins { namespace account_history {
 
-using boost::container::flat_set;
-
 namespace detail {
 
 class abstract_account_history_api_impl
--- a/libraries/plugins/apis/account_history_api/account_history_api.cpp
+++ b/libraries/plugins/apis/account_history_api/account_history_api.cpp
@@ -37,30 +37,20 @@ DEFINE_API_IMPL( account_history_api_chainbase_impl, get_ops_in_block )
 {
    return _db.with_read_lock( [&]()
    {
-      std::multiset< api_operation_object > tmp_result;
-
       const auto& idx = _db.get_index< chain::operation_index, chain::by_location >();
       auto itr = idx.lower_bound( args.block_num );
+
       get_ops_in_block_return result;
+
       while( itr != idx.end() && itr->block == args.block_num )
       {
          api_operation_object temp = *itr;
          if( !args.only_virtual || is_virtual_operation( temp.op ) )
-            tmp_result.emplace( std::move( temp ) );
+            result.ops.emplace( std::move( temp ) );
          ++itr;
       }
 
-      if( !tmp_result.empty() )
-      {
-         get_ops_in_block_return result;
-
-         result.ops.resize( tmp_result.size() );
-         std::copy( tmp_result.begin(), tmp_result.end(), result.ops.begin() );
-
-         return result;
-      }
-
-      return get_ops_in_block_return();
+      return result;
    });
 }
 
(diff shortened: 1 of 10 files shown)


mvandeberg:
@theoreticalbts The changes to the indices seems reasonable to me. I want your opinion on them as well. Currently, `database_api` is WIP, so I am not worried about the dense implementation or it even being correct right now.

mvandeberg:
I am fine removing unused indices from `database_api` code, but we need to keep them in for the time being if they are used by `condenser_api`. To be clear, `appbase` is not breaking APIs. We are deprecating old, inefficient, and inconsistent APIs with a properly designed API. If, as part of that redesign, we want to change how certain data is returned, that is fine. We do need to support the old APIs in `condenser_api` for the time being however. Any change that breaks `condenser_api` needs to have a compatibility fix. If it is something we cannot change now, but want to in the future, comment it in source with an `FC_TODO`.

mariusztrela:
- `comment_vote_object` instead of `proxy_item`
I agree with you - fix is done.

- `vector` instead of `set`
I don't agree, because for `vector` double scanning is needed( inserting + sorting ). For `set` is only inserting.

- list_votes re-implementation is problematic

In fact if we want to retrieve N records, now we have to retrieve N + K additional information to close actual subset.
Despite that additional records, there is no problem with performance. It was done many tests regarding to performance and no any slowdown was detected.

mvandeberg:
I agree that using a `set` instead of a `vector` for comment vote sorting is best. `vector` is alright, but as @mariusztrela points out, we need to insert and then sort, whereas we can insert into a sorted data structure with `set`. If we absolutely wanted to use a flat structure, a `flat_set` would be the correct choice.

libraries/plugins/apis/account_history_api/account_history_api.cpp
```diff
@@ -37,17 +37,30 @@ DEFINE_API_IMPL( account_history_api_chainbase_impl, get_ops_in_block )
 {
    return _db.with_read_lock( [&]()
    {
+      std::multiset< api_operation_object > tmp_result;
```
mvandeberg: FC has multiset serialization. Let's change the returned type in `get_ops_in_block_return` to the multiset type to eliminate copying to a vector. When we implement this for rocksdb account history plugin, we can override the comparator to always return true so that it behaves as a queue.

PR #2232, opened 2018-03-15
2075 json tuples3

mariusztrela:
#2075 
@mvandeberg I had couple problems with removing conflicts, so I decided to make new PR.

### DIFF
--- a/libraries/fc/include/fc/static_variant.hpp
+++ b/libraries/fc/include/fc/static_variant.hpp
@@ -350,7 +350,7 @@ struct visitor {
       template<typename T> void operator()( const T& v )const
       {
          auto name = trim_typename_namespace( fc::get_typename< T >::name() );
-         var = variant( std::make_pair( name, v ) );
+         var = mutable_variant_object( "type", name )( "value", v );
       }
    };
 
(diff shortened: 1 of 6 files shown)


libraries/fc/include/fc/static_variant.hpp
```diff
@@ -365,22 +366,54 @@ struct visitor {
...
+            fc::static_variant<T...> tmp;
+            tmp.set_which(i);
+            string n;
+            tmp.visit( get_static_variant_name( n ) );
+            name_map[n] = i;
+         }
+         return name_map;
+      }();
+
+      FC_ASSERT( v.is_object(), "Operation has to treated as object." );
```
mvandeberg: Error messages should be generic to static variant, not operations. Operations don't exist in fc.

theoreticalbts:
@mvandeberg I'm leaving review of this PR to you, since you seem to have a plan for dealing with backward compatibility.

I would personally suggest adding a `static_variant` wrapper to `condenser_api`, and letting the other API's break.

mvandeberg:
@theoreticalbts That is exactly what we I had done. It just needs to be updated on this branch. There was a miscommunication. I had done nearly all of the work for this in another issue. There was significant overlap that was not documented and lead to duplicate work being done. To make reviewing my changes easier, I am going to merge this and then make a new PR for `condenser_api` compatibility.

Issue #2160, opened 2018-02-26
Modify transfer_to_vesting_operation to support SMT vesting


youkaicountry:
The `transfer_to_vesting_operation` should be modified to support vesting SMTs. This will include modification of `database::create_vesting` to support vesting SMTs.

vogel76:
@youkaicountry @theoreticalbts Given questions appeared to `database::create_vesting`:
1) does  the same algorithm shall be used to calculate new_vesting for SMTs ? If where is specified actual one.
2) if in 1) the same algorithm shall be used, how vesting share prices shall be retrieved ? Does smt_token_object shall have additionally defined such parameters ?
Does the SMT whitepaper specify such rules ?

theoreticalbts:
So here's some information on how Steem vesting works.  This information should probably go in the wiki, but I can't find the wiki, so I will put it here.

# Steem, vesting, and Steem Power

The core idea of vesting is that *holding STEEM creates value for everyone*.  So the blockchain should have an operation allowing users to say "This lump of STEEM, I won't transfer it for a while."  And the blockchain enforces that commitment.

Such Steem is called "Steem Power."  New commitments to holding are called "Powering Up."  Changing held Steem to liquid Steem is called "Powering Down."

Originally, the terminology used was "vesting."  But at some point, it was deemed that "Powering Up/Down" is better terminology.

People won't voluntarily choose to hold their STEEM unless there is some incentive.  The incentives are as follows:

- Only vesting STEEM counts for voting
- Vesting STEEM is paid a portion of STEEM emissions

So then the implementation problem becomes, how do you implement the idea of "OK, a bunch of users all have powered-up Steem.  Every block, we get a little lump of STEEM from emissions, and we want to divide that new STEEM among all users with powered-up Steem according to their balances.  How do
we distribute that little lump without having to iterate over everyone's balance every block?"

The answer is, you combine all the powered-up Steem in a single "pot," and then give everyone claims against that pot.  With this data structure, the rest of the implementation becomes simple and efficient:

- If you want to "pay everyone Steem proportionally to their powered up balance," you just add Steem to the pot without printing any new claims.
- If I have `K` claims, and in total `N` claims exist, how much are my `K` claims worth?  Well, the value of all `N` claims is the pot, and since claims should be "fungible" or "homogenous", the value of `K` claims should be the Steem in the pot, times `K/N`.
[...]

PR #2233, opened 2018-03-16
Implemented SMT transfer to vesting

Kiwonik:
Modified database::create_vesting and transfer_to_vesting_operation, reorganized balance objects, added tests #2160
Notes to reviewers:
1. Changes that may not be obvious at first sight:
    - In spirit of `database::adjust_balance` name change I added appropriate assertions in both of these methods.
    - I had to split common SMT `account_balance_object` into separate rewards and regular classes due to different number of fields they need to hold - specifically due to pending vesting balance needing to be stored in two fields. Fortunately I managed to keep the template code that manipulates SMT balances unified.
    - SMT versions of `get_vesting_share_price` & `get_reward_vesting_share_price` diverged from originals placed in `dynamic_global_property_object` in that both of them are zero-vesting-safe. Why only one of them has an appropriate check in dgpo? And why original `get_vesting_share_price` returns STEEM_SYMBOL/VESTS_SYMBOL price when it should be the other way round?
    - modified `database::adjust_balance` and `database::adjust_reward_balance` to support vesting too.
1. ~There are several additional modifications that I if agreed would like to add to this PR:~ Additional change made to keep the code tidy:
    - removing unused database::pay_fee

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -1058,35 +1058,85 @@ std::pair< asset, asset > database::create_sbd( const account_object& to_account
 
 /**
  * @param to_account - the account to receive the new vesting shares
- * @param STEEM - STEEM to be converted to vesting shares
+ * @param liquid     - STEEM or liquid SMT to be converted to vesting shares
  */
-asset database::create_vesting( const account_object& to_account, asset steem, bool to_reward_balance )
+asset database::create_vesting( const account_object& to_account, asset liquid, bool to_reward_balance )
 {
    try
    {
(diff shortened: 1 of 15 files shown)


vogel76:
I doubt if creating separate set of balance related methods for vesting assets is good decision. If any specific behavior is needed inside, better code it inside under condition determining that given asset points to vesting one. Otherwise we pollute caller side (where such checks will be moved to call appropriate function).

theoreticalbts:
It is functionally correct, but there are a few minor things that need fixed.  I will file as a separate issue.

Issue #2235, opened 2018-03-19
Create smt_base_operation


youkaicountry:
To avoid redundant code, create a base operation `smt_base_operation` from which which smt operations will descend. 

It should contain members representing control_account and token, and should have a validate method which validates these members.

theoreticalbts:
I am not sure if FC reflection or some of our other template automagic (for example `static_variant.visit()` dispatch) can handle this.

mvandeberg:
`FC_REFLECT_DERIVED` handles this.

PR #2245, opened 2018-03-21
Added common base classes for all smt operations

Kiwonik:
Issue #2235
Notes for reviewers:
1. Although `smt_base_operation` has been created and used as described in the issue, the pinged/executor operation required separate `smt_executor_base_operation` base class as it needs `executor` account name instead of `control_account` name. However the base classes' validate methods share common code.
2. Due to decision to have two base classes not only `validate` but also `get_required_active_authorities` could be moved there, allowing further unification and uncluttering of the code.

### DIFF
--- a/libraries/protocol/include/steem/protocol/smt_operations.hpp
+++ b/libraries/protocol/include/steem/protocol/smt_operations.hpp
@@ -8,27 +8,50 @@
 namespace steem { namespace protocol {
 
 /**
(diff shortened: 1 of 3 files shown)


theoreticalbts:
Accepted, but see #2251.

Kiwonik:
Actually #2251 has nothing to do with this issue/PR. #2056 does.

PR #2273, opened 2018-03-26
Account History RocksDB API Refactor

### LINKED ISSUE #2246: Refactor rocksdb_api
`rocksdb_api` registers itself as `account_history_api`. What should be done is both versions of the APIs should exist in the `account_history_api` and the appropriate version loaded depending on which backing database is running.

There are calls in `condenser_api` that make calls to `account_history_api`. This approach means `condenser_api` can work seamlessly with either backing db.

mvandeberg:
Closes #2246


libraries/plugins/apis/account_history_api/account_history_api.cpp
```diff
@@ -1,92 +1,212 @@
...
 
-account_history_api::account_history_api(): my( new detail::account_history_api_impl() )
+account_history_api::account_history_api()
 {
+   auto ah_cb = appbase::app().find_plugin< steem::plugins::account_history::account_history_plugin >();
+   auto ah_rocks = appbase::app().find_plugin< steem::plugins::rocksdb::rocksdb_plugin >();
+
+   if( ah_rocks != nullptr )
+   {
+      if( ah_cb != nullptr )
```
vogel76: I think better could be add dedicated option `account-history-legacy-storage` which by default is false and allows to enable old AH implementation. It could be more readable to customers/admins than implicit decision caused by adding old account_history_plugin to the list.

mvandeberg: The point of this is that it is automatic. The option `account-history-legacy-storage` accomplishes the same thing they already have to do with `plugin=account_history` or `plugin=account_history_rocksdb`

PR #2274, opened 2018-03-26
Cleanup Account History RocksDB

### LINKED ISSUE #2247: Cleanup rocksdb configuration
As @sneak noted in #2129 

> I think perhaps that the new rocksdb-based AH plugin should be given a new name, so that people can enable the new-style rocksdb-based plugin or the old-style AH plugin with a small list of whitelisted ops or account names and continue using steemd like they did before.

I suggest renaming `rocksdb_plugin` to `account_history_rocksdb`. The existing account history plugin and rocksdb account history should be made mutually exclusive and should both accept the same config options so that upgrading to rocksdb is as simple as changing the plugin list and reindexing.

mvandeberg:
Closes #2247 

PR #2273 should be reviewed and merged first


mvandeberg:
We have a standard that we are trying to follow with all plugin options that they should be prefixed by the plugin name. Over time we have added many options to the config file and it is not clear what options need to be set and when. Your implementation violates that standard. I understand your intentions, and I think they are good, that is not how we will be implementing the options for the rocksdb plugin. If we prefix the old account history then if someone wants to continue using their node exactly as is, then nothing needs to change. If we change their config options then their node will not work the same way when they update.

Let's assume that we are going to follow the option naming standard. Then we have two options.

1. The RocksDB backed account history has a different option name and the existing account history will remain fully backwards compatible.

2. We change the legacy account history plugin name and rename the RocksDB backed account history plugin to `account_history`. All existing configs work with new plugin and all old account history setups will be running RocksDB when they update. This might seems like a good idea, but my counter argument to this is Windows 10. Updating to the RocksDB backed account history plugin should be a choice for the node operator. All of the options work the same, so the upgrade path is minimal, but we should not force them to upgrade immediately.

PR #2313, opened 2018-04-05
piped lz4 and retries

jredbeard:
This is a better way to handle this - the decompression doesn't take longer than the file transfer so it's not blocking, and it will retry on failure up to five times.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN \
         libreadline-dev \
         libssl-dev \
         libtool \
+        liblz4-tool \
         ncurses-dev \
-        pbzip2 \
         pkg-config \
         python3 \
         python3-dev \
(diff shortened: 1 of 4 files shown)


mvandeberg:
There is still the issue that if the sync service fails to pull down the state file then it will reindex. On `master` that is not a problem because there is a bug preventing the sync service from reindexing at all, so the node dies and is restarted.

However, that is fixed in `develop`, which means if the sync service fails to pull down the state file then it _will_ start a reindex, which can lead to the latest state file becoming stale and not being able to bring up new steemds if they go down.

mvandeberg:
When the file does exist, `aws s3` prints a nice message.

```
aws s3 cp s3://steemit-dev-blockchainstate/foobar ./foobar
fatal error: An error occurred (404) when calling the HeadObject operation: Key "foobar" does not exist
```

But it still has a return code, 1.

What we can do is sed the stdout/err for this message and return a custom defined error code on no-match.

Then we can differentiate between the state file failing to download vs not existing.

jredbeard:
The idea here is that it still tries five times to pull the state file on steemdsync - if it can’t get it on the fifth try, it reindexes... if it’s a failure, it probably won’t fail five times in a row unless the file really doesn’t exist

jredbeard:
I like the idea of parsing the output if even via sed but as a quick hack this should work 99% of the time or more

PR #2316, opened 2018-04-06
Wait for steemdsync to exit

### LINKED ISSUE #2305: Sync Service Wait Time
Sync services need to wait for the process to exit completely instead of sleeping before uploading state files to S3.

https://github.com/steemit/steem/blob/master/contrib/sync-sv-run.sh#L42-L45

``` bash
kill -SIGINT $STEEMD_PID
echo steemdsync: waiting for steemd to exit cleanly
while [ -e /proc/$STEEMD_PID ]; do sleep 0.1; done
echo steemdsync: starting a new blockchainstate upload operation
```

mvandeberg:
Closes #2305


mvandeberg:
I tested this change locally with great success.

mvandeberg:
Currently we upload regardless, so I do not see this as any worse than what we currently have. steemd should never take 60 seconds to shut down, so I think this is acceptable. We could terminate the instance without uploading if we hit the timeout.

Issue #2230, opened 2018-03-14
Clean Up Database Signal Interface


mvandeberg:
To implement #1996, #2149 changed the interface for plugins registering with a signal to save the name for statistics collecting.

Each plugin has a `name()` method already, that should be used here. Instead of creating a proxy method, which isn't well documented and confusing when the signal is still exposed, I suggest creating a macro `PLUGIN_REGISTER_SIGNAL( signal, method )` that calls the appropriate method in the database, creates the lamba properly, etc. This will also allow use of `name()`, so the correct name is always registered and is a much cleaner interface.

The group can also be exposed by the macro. It should still return the connection object to be stored in the plugin itself.

PR #2325, opened 2018-04-11
Clean up signal handling

theoreticalbts:
This PR is a fairly self explanatory refactoring.  Here are the main features:

- Rename signals and related methods to be descriptive and consistent
- Always use an `XXX_notification` struct as the signal argument (simplifies code changes needed to send more data to a signal handler)
- Add pre-apply signal for blocks
- Always use groups (allows control of processing order when needed, so some plugins' handlers can depend on the results of other handlers)
- Remove unused handler (its functionality is now available by checking whether or not you're currently processing a block)


mvandeberg:
Closes #2230

Issue #2244, opened 2018-03-21
Routing per-block data


theoreticalbts:
Several initiatives require a common data path:  Data generated by core or plugins while processing operations needs to be attached to the block where those operations occurred.

Examples of problems that would benefit from this data path:

- Virtual ops in external account history implementation
- Enodes
- Modular blockchain

Here is the strategy I will use to implement this data path:

- Allow querying whether the database is currently processing a block
- Change method signature from `applied_block( const signed_block& )` to  `applied_block( const block_notification& )` where `block_notification` is a newly defined `struct` that encapsulates a `signed_block` reference and any other data that needs to follow the same path (matching the `operation` signals)
- Rename `applied_block` signal to `post_apply_block` to match the naming of operations
- Implement `pre_apply_block` signal so plugins know a new block is starting
- Implement `exc_apply_block` signal so plugins know block processing is stopping due to an exception
- Implement `struct block_output` which will be a container any output produced by core and plugins
- Allow access to current block's `block_output` for core / plugins
- Implement `block_output` cache, which is `map< block_id_type, block_output >` and is not persistent over node shutdown
- Implement configuration constants describing when to discard `block_output` (irreversibility plus height), or allowing `block_output` generation to be disabled entirely
- Allow `block_api` to return `block_output`
- When switching forks, disable generation of `block_output` and use cached values
- Implement `block_stream` plugin which streams block output to file (or UNIX FIFO)
- Implement a couple small example Python scripts to consume `block_stream` output

PR #2326, opened 2018-04-11
Export stats

theoreticalbts:
This code implements two plugins:

- `block_data_export_plugin` allows export of per-block data; it effectively implements #2244.
- `stats_export_plugin` uses the interface provided by `block_data_export_plugin` to dump GPO and per-user bytes per block.  The intent is that this data will be consumed by a Python script for further processing.

The data path implemented by `block_data_export_plugin` will be extremely useful going forward, especially when combined with a UNIX FIFO connected to some external script.

I also implemented a simple thread pool to allow all available processor cores to be utilized for the JSON conversion of exported data objects.  Unfortunately Boost's thread pools aren't suitable, because (1) they're marked experimental and may be subject to change in future Boost versions, and (2) there's no way to set the thread stack size, which we need because FC JSON processing uses stack very heavily.

When processing the main blockchain, approximately 40 GB of stats are produced.  To minimize disk usage, you can do streaming compression of the output JSON by simply using a UNIX FIFO connected to `pbzip2` or similar.

### DIFF
--- a/libraries/plugins/stats_export/stats_export_plugin.cpp
+++ b/libraries/plugins/stats_export/stats_export_plugin.cpp
@@ -1,0 +1,156 @@
+
+#include <steem/plugins/block_data_export/block_data_export_plugin.hpp>
+#include <steem/plugins/block_data_export/exportable_block_data.hpp>
+
+#include <steem/plugins/stats_export/stats_export_plugin.hpp>
+
+#include <steem/chain/account_object.hpp>
+#include <steem/chain/database.hpp>
+#include <steem/chain/global_property_object.hpp>
+#include <steem/chain/index.hpp>
+#include <steem/chain/operation_notification.hpp>
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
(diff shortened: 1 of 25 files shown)


libraries/plugins/stats_export/stats_export_plugin.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+namespace steem { namespace plugins { namespace stats_export { namespace detail {
+
+class stats_export_plugin_impl
+{
+   public:
+      stats_export_plugin_impl( stats_export_plugin& _plugin ) :
+         _db( appbase::app().get_plugin< chain_plugin >().db() ),
+         _self( _plugin ),
+         _export_plugin( appbase::app().get_plugin< block_data_export_plugin >() )
```
mvandeberg: Because you are using the `block_data_export_plugin` as a dependency, you should specify it in `APPBASE_PLUGIN_REQUIRES`

mvandeberg:
I only reviewed commits 3548eae and 9354983 because the others are included in #2325

Issue #2302, opened 2018-04-03
Fix appbase shutdown logic


mvandeberg:
As discussed in #2301

There is a problem with the async writes. The chain plugin accepts write references from the p2p plugin. However, the p2p plugin shuts down before the chain plugin. This is by design as the p2p plugin depends on the chain plugin to operate. This is also true with the `network_broadcast_api_plugin`.

There are two solutions that we could implement. 

1. The chain plugin gets its own copy of all writes.
2. The p2p plugin (and other plugins that write) need to track their writes and only shutdown after they can guarantee there are not outstanding writes.

There is a more subtle problem that exists with other plugins that falls into this category. Any plugin that accepts data via a db signal depends on the chain plugin and thus shuts down prior to the chain plugin. This means a plugin that is technically shut down may still receive data from the chain plugin. These plugins should reject such writes if they are shutting down and clean up the signal in the shutdown code. I do not believe any references are invalidated during shut down, only during deletion of the object which happens implicitly during process clean up.

Issue #2221, opened 2018-03-12
P2P Statistics Segfault


mvandeberg:
There have been a number of race condition related segfaults. Many have been fixed.

While investigation #2205 I ran into another one.

I found that when running the witness plugin my node was dying within the first 1m sync blocks. When I ran it in the debugger it crashed reliably. Part of this is that building in debug does not optimize offset pointers used in the memory mapped file, slowing steemd by ~10x. This seems to indicate the bug is somewhat load dependent.

```
* thread #2, stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x0000000100b3a80c steemd`boost::cb_details::debug_iterator_base::next(this=0x48e5894855000044) const at debug.hpp:228
    frame #1: 0x0000000100b3a7db steemd`boost::cb_details::debug_iterator_registry::unregister_iterator(this=0x000000010a014db0, it=0x0000700001bfe490) const at debug.hpp:145
    frame #2: 0x0000000100b3a77d steemd`boost::cb_details::debug_iterator_base::unregister_self(this=0x0000700001bfe490) at debug.hpp:239
    frame #3: 0x0000000100b3a735 steemd`boost::cb_details::debug_iterator_base::~debug_iterator_base(this=0x0000700001bfe490) at debug.hpp:211
    frame #4: 0x0000000100b3a715 steemd`boost::cb_details::iterator<boost::circular_buffer<fc::ripemd160, std::__1::allocator<fc::ripemd160> >, boost::cb_details::nonconst_traits<boost::container::allocator_traits<std::__1::allocator<fc::ripemd160> > > >::~iterator(this=0x0000700001bfe490) at details.hpp:198
    frame #5: 0x0000000100afcc85 steemd`boost::cb_details::iterator<boost::circular_buffer<fc::ripemd160, std::__1::allocator<fc::ripemd160> >, boost::cb_details::nonconst_traits<boost::container::allocator_traits<std::__1::allocator<fc::ripemd160> > > >::~iterator(this=0x0000700001bfe490) at details.hpp:198
[...]

PR #2327, opened 2018-04-11
2302 rocksdb shutdown

vogel76:
Issue #2302 Fixed synchronization issues during shutting down p2p pluggin, what could lead to application crash.


vogel76:
I think it is worth to consider re-enabling statistics disabled in task:
https://github.com/steemit/steem/issues/2221
Such problem IMO also was caused by missing sync and accessing already deleted node object.

PR #2336, opened 2018-04-16
2302 rocksdb shutdown supplement

vogel76:
Issue #2302
1. Solved synchronization problems during graphene::node shutdown (all async tasks are forcing a node to wait until their finish). To simplify code dedicated methods `async_task`, `schedule_task` have been added, which encapsulate synchronization management.
2. Eliminated superfluous statistic_collector object declaration from `INVOKE_AND_COLLECT_STATISTICS` macro, which caused crash during regular sync (looks like there was uncontrolled write to the same accumulator_set object done by 2 threads).
3. SIGPIPE is ignored by steemd right now to correctly handle SIGINT in case steemd was spawned in pipe ie: `steemd | tee steemd.log`. Previously it was killed (default action for broken pipe) when Ctrl+C was pressed on the tee input.

### DIFF
--- a/libraries/net/node.cpp
+++ b/libraries/net/node.cpp
@@ -801,7 +801,8 @@ namespace graphene { namespace net {
       _average_network_write_speed_hours(72),
       _average_network_usage_second_counter(0),
       _average_network_usage_minute_counter(0),
-      _node_is_shutting_down(false)
+      _node_is_shutting_down(false),
+      _activeCalls(0)
     {
       _rate_limiter.set_actual_rate_time_constant(fc::seconds(2));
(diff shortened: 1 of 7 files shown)


libraries/net/node.cpp
```diff
@@ -1003,7 +1041,6 @@ namespace graphene { namespace net {
 
     void node_impl::fetch_sync_items_loop()
     {
-      VERIFY_CORRECT_THREAD();
```
mvandeberg: What is the rationale for removing this check? It seems like removing it to work properly would indicate a potential problem in the existing code.

mvandeberg: Ok, looks like this was moved to `activity_tracer`

PR #2339, opened 2018-04-16
Add additional assertion to prevent loop

### LINKED ISSUE #2337: account_history_api.get_account_history infinite loop
Change in commit:
https://github.com/steemit/steem/commit/068d90a36cf946cee82143db038d5268d6c6a4d8
caused a bug leading to infinite iteration over retrieved AH entries inside `account_history_api_chainbase_impl, get_account_history.
Steps to reproduce:
1) enable old account_history plugin and API in the config file ie:
`plugin = account_history account_history_api database_api`
2) Replay blocks up to 100000
3) Make a query:
```
{"jsonrpc": "2.0", "method": "call", "params": ["account_history_api","get_account_history",{"account": "aaa", "start": 0, "limit": 0}], "id": 2}
```
Execution hangs inside loop `while(itr != end)`.
As a sidenote, application will crash in case the container is empty since there is unconditional reference to itr during upper_bound call.

mvandeberg:
Closes #2337

### DIFF
--- a/libraries/plugins/apis/account_history_api/account_history_api.cpp
+++ b/libraries/plugins/apis/account_history_api/account_history_api.cpp
@@ -63,6 +63,7 @@ DEFINE_API_IMPL( account_history_api_impl, get_account_history )
    const auto& idx = _db.get_index< chain::account_history_index, chain::by_account >();
    auto itr = idx.lower_bound( boost::make_tuple( args.account, args.start ) );
    auto end = idx.upper_bound( boost::make_tuple( args.account, std::max( int64_t(0), int64_t(itr->sequence) - args.limit ) ) );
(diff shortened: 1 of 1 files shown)


theoreticalbts:
I'm not sure this is the best way to deal with this.  I suggest something like the following:

```
while( true )
{
   if( itr == idx.end() )
      break;
   if( itr->account != args.account )
      break;
   result.history[ itr->sequence ] = ...;
   if( result.history.size() >= args.limit )
      break;
   ++itr;
}
```

If you try to construct a sentinel `end` value, you have to get it exactly right, or you could potentially iterate past the end.  It's much easier to be confident the past-the-end check is correct when it's written this way.

mvandeberg:
Closing in favor of #2342

PR #2343, opened 2018-04-16
Fix iteration in account_history_api.cpp #2337 (master branch)

theoreticalbts:
This is PR #2342, rebased against `master` branch.


mvandeberg:
This appears to be working.

Issue #2291, opened 2018-03-29
Implement StorageConfigurationManager class


vogel76:
Functionality of this class covers:
- gathering and distributing directory options for main RocksDB store and eachone specific to plugin(s)
- providing functionality to pass RocksDB configuration file to any held RocksDB storage and allow this way broad reconfiguration for each embedded database.

Details in [architecture document](https://docs.google.com/document/d/1LPScSMURj5x4hQumkKSLkap__BDnyLpOGMKqWVnBo3c/edit?usp=sharing)

Issue #2289, opened 2018-03-29
Design and implement custom STL-like iterators to be used in bmic replacement


vogel76:
Iterator classes must provide functionally equivalent to bmic iterators:
- iterator
- const_iterator
- reverse_iterator
- const_reverse_iterator

There are needed implementations being able to filter out source data and to concatenate data gathered from multiple data sources.

Iterator features:
- [x] dereference operator
- [x] forward iteration (increment operator)
- [x] backward iteration (decrement operator)
- [x] comparison operator (==, !=)
- [x] copy constructible, assignable
- [x] std::iterator_traits support

vogel76:
Work is done inside:
https://github.com/steemit/steem/blob/2041-bmic-replacement

Issue #2288, opened 2018-03-29
Create an unit test covering bmi::multi_index_container functionality 


vogel76:
Test is needed to verify correctness of work done inside #2041

vogel76:
Completed on 19.03.2018. Work done in branch:
https://github.com/steemit/steem/blob/2041-bmic-replacement

Issue #2041, opened 2018-01-26
Design and implement boost::multi_index replacement basing on RocksDB storage 


vogel76:
Designed class must emulate boost::multi_index container interface to allow simple and iterative replacement of each defined index, without a need to do massive changes in other code using these indices.

Work can be split into following phases:

- [x] Create an unit test covering the container use cases #2288 
- [x] Implement custom iterators be responsible for providing data from one or more concatenated sources #2289
- [ ] Implement pure `boost::multi_index::multi_index_container` adapter class, which can have specified different (potentially multiple) data sources #2290
- [x] Implement StorageConfigurationManager class providing functionality described in the [architecture document](https://docs.google.com/document/d/1LPScSMURj5x4hQumkKSLkap__BDnyLpOGMKqWVnBo3c/edit?usp=sharing) #2291
- [ ] Implement PersistentStorage class providing functionality described in in the [architecture document](https://docs.google.com/document/d/1LPScSMURj5x4hQumkKSLkap__BDnyLpOGMKqWVnBo3c/edit?usp=sharing) #2292
- [ ] Implement a data source providing objects stored in the RocksDB database #2293

vogel76:
This task is blocked until #1984 will be merged to develop to have codebase for further work.

youkaicountry:
Blocked by #2158

youkaicountry:
An additional requirement while implementing this:
For any index that is non-unique, an iterator should not be allowed to be returned.

PR #2355, opened 2018-04-18
2041 bmic replacement

mariusztrela:
**Design and implement boost::multi_index replacement basing on RocksDB storage #2041** 

- Create an unit test covering the container use cases #2288
- Implement custom iterators be responsible for providing data from one or more concatenated sources #2289
- Implement StorageConfigurationManager class providing functionality described in the architecture 'High level design of RocksDB based persistent layer' created by @vogel76  #2291

### DIFF
--- a/libraries/chainbase/include/chainbase/chainbase.hpp
+++ b/libraries/chainbase/include/chainbase/chainbase.hpp
@@ -21,6 +21,7 @@
 #include <boost/throw_exception.hpp>
 
 #include <chainbase/allocators.hpp>
+#include <chainbase/util/object_id.hpp>
 
 #include <array>
 #include <atomic>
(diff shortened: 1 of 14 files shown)


mvandeberg:
Is it possible to split this into 3 PRs for each of the issues?

Even if some of them are dependent upon others, this would make reviewing the changes significantly easier. If there are dependencies, please document them in the PRs.

vogel76:
I don't think so it is worth to split code into separate parts. This code does not affect application runtime right now since it is not yet live (not effectively used), We would like to merge it into develop to make code base current against develop changes and continue work on strict container replacement etc.

mvandeberg:
I am asking to split it to make it easier to review each specific task. I understand why you want it merged. I have no problems with it. But I am not able to review this code as effectively as I could if it were split.

mariusztrela:
I split this PR into 3 parts.

mvandeberg:
Replaced by #2362, #2363, and #2364

Issue #2253, opened 2018-03-21
PR #2233 SMT transfer_to_vesting problems


theoreticalbts:
These are some minor issues with PR #2233 for SMT `transfer_to_vesting`.

First, a word about `validate()` methods:

- `account_regular_balance_object::validate()` and `account_rewards_balance_object::validate()` return a `bool` value.  Elsewhere, `validate()` methods indicate failure by throwing an exception.  It would be very easy to write code that calls these `validate()` methods that looks correct, but is actually a no-op since it doesn't check return value.
- `xxx_object` usually doesn't have `validate()`, only operations have `validate()`
- If these `validate()` methods are unused, let's delete them.
- If these `validate()` methods can be easily rewritten to `FC_ASSERT()`, let's do that.
- If it is hard to rewrite these methods, let's rename them something other than `validate()` (maybe `check_xxx()` or `is_xxx()`)

The line `FC_ASSERT( smt.allow_voting == to_reward_balance )` is problematic.  I think maybe it should be `if( to_reward_balance ) { FC_ASSERT( smt.allow_voting ); }` which is not logically equivalent.

The `delta` value is passed into `smt_balance_operator` and also passed as a parameter.  This seems unnecessary, let's get rid of the parameter and read the `smt_balance_operator` field when needed.

Why are we adding the VESTS case to `modify_balance()`?  Shouldn't vesting adjustment always go through vesting methods to avoid problems?

Look at the `is_vesting` initializer in `smt_reward_balance_operator`.  This may be incorrect.

youkaicountry:
Referenced in #2746 
Closing as outdated.

PR #2356, opened 2018-04-18
Post-merge fixes to SMT transfer to vesting

Kiwonik:
#2253 
Done, to the extent described in issues last comment https://github.com/steemit/steem/issues/2253#issuecomment-382403447

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -3747,18 +3747,18 @@ void database::clear_expired_delegations()
 }
 #ifdef STEEM_ENABLE_SMT
 template< typename smt_balance_object_type, class balance_operator_type >
-void database::adjust_smt_balance( const account_name_type& name, const asset_symbol_type& delta_symbol, bool positive_delta, bool zero_delta,
-                                   bool check_account, balance_operator_type balance_operator )
+void database::adjust_smt_balance( const account_name_type& name, bool check_account, const balance_operator_type& balance_operator )
 {
-   asset_symbol_type liquid_symbol = delta_symbol.is_vesting() ? delta_symbol.get_paired_symbol() : delta_symbol;
+   asset_symbol_type balance_symbol = balance_operator.get_balance_symbol();
+   asset_symbol_type liquid_symbol = balance_symbol.is_vesting() ? balance_symbol.get_paired_symbol() : balance_symbol;
    const smt_balance_object_type* bo = find< smt_balance_object_type, by_owner_liquid_symbol >( boost::make_tuple( name, liquid_symbol ) );
    // Note that SMT related code, being post-20-hf needs no hf-guard to do balance checks.
    if( bo == nullptr )
    {
       // No balance object related to the SMT means '0' balance. Check delta to avoid creation of negative balance.
-      FC_ASSERT( positive_delta, "Insufficient SMT ${smt} funds", ("smt", delta_symbol) );
+      FC_ASSERT( balance_operator.is_positive_delta(), "Insufficient SMT ${smt} funds", ("smt", balance_symbol) );
       // No need to create object with '0' balance (see comment above).
-      if( zero_delta )
+      if( balance_operator.is_zero_delta() )
          return;
 
       if( check_account )
@@ -3777,7 +3777,7 @@ void database::adjust_smt_balance( const account_name_type& name, const asset_sy
       bool is_all_zero = false;
       int64_t result = balance_operator.get_combined_balance( bo, &is_all_zero );
       // Check result to avoid negative balance storing.
-      FC_ASSERT( result >= 0, "Insufficient SMT ${smt} funds", ( "smt", delta_symbol ) );
+      FC_ASSERT( result >= 0, "Insufficient SMT ${smt} funds", ( "smt", balance_symbol ) );
 
       // Exit if whole balance becomes zero.
       if( is_all_zero )
@@ -3900,19 +3900,23 @@ struct smt_regular_balance_operator
 {
    smt_regular_balance_operator( const asset& delta ) : delta(delta), is_vesting(delta.symbol.is_vesting()) {}
 
-   void add_to_balance( account_regular_balance_object& smt_balance )
+   void add_to_balance( account_regular_balance_object& smt_balance ) const
    {
       if( is_vesting )
          smt_balance.vesting += delta;
       else
          smt_balance.liquid += delta;
    }
-   int64_t get_combined_balance( const account_regular_balance_object* bo, bool* is_all_zero )
+   int64_t get_combined_balance( const account_regular_balance_object* bo, bool* is_all_zero ) const
    {
       asset result = is_vesting ? bo->vesting + delta : bo->liquid + delta;
       *is_all_zero = result.amount.value == 0 && (is_vesting ? bo->liquid.amount.value : bo->vesting.amount.value) == 0;
       return result.amount.value;
    }
+
+   const asset_symbol_type&   get_balance_symbol() const { return delta.symbol; }
+   bool                       is_zero_delta() const      { return delta.amount == 0; }
+   bool                       is_positive_delta() const  { return delta.amount.value >= 0; }
 
    asset delta;
    bool  is_vesting;
@@ -3926,7 +3930,7 @@ struct smt_reward_balance_operator
        FC_ASSERT( value_delta.symbol.is_vesting() == false && share_delta.symbol.is_vesting() );
    }
 
-   void add_to_balance( account_rewards_balance_object& smt_balance )
+   void add_to_balance( account_rewards_balance_object& smt_balance ) const
    {
       if( is_vesting )
       {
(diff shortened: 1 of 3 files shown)


libraries/chain/database.cpp
```diff
@@ -1132,7 +1132,7 @@ asset database::create_vesting( const account_object& to_account, asset liquid,
          FC_ASSERT( liquid.symbol.is_vesting() == false );
          // Get share price.
          const auto& smt = get< smt_token_object, by_symbol >( liquid.symbol );
-         FC_ASSERT( smt.allow_voting == to_reward_balance, "No voting - no rewards" );
+         FC_ASSERT( smt.allow_voting || to_reward_balance == false, "No voting - no rewards" );
```
mvandeberg: I think this assert is still problematic, but for a different reason than stated in #2253

The vesting payout code isn't done yet, but it appears that this is designed for the payout code to blindly call `create_vesting` and fail if there shouldn't be a payout. This is bad way to implement this when the state of allowed voting can be determined before `create_vesting` is called. As a general rule, we only throw exceptions in consensus code to indicate a consensus rule is broken and the operation/transaction/block in question should be rejected. In this case it appears that you are using the exception to indicate a control path, which is not a good use of exceptions.

libraries/chain/database.cpp
```diff
@@ -3746,18 +3746,18 @@ void database::clear_expired_delegations()
 }
 #ifdef STEEM_ENABLE_SMT
 template< typename smt_balance_object_type, class balance_operator_type >
-void database::adjust_smt_balance( const account_name_type& name, const asset& delta, bool check_account,
-   balance_operator_type balance_operator )
+void database::adjust_smt_balance( const account_name_type& name, const asset_symbol_type& delta_symbol, bool positive_delta, bool zero_delta,
```
mvandeberg: If we are going to use a templated generic method with specific implementation in a class type, let's also encode such fields in the class. Both `smt_regular_balance_operator` and `smt_reward_balance_operator` contains assets, which contain the asset symbols. And as far as I can tell, `positive_delta` and `zero_delta` can either be calculated from the `balance_operator_type` or can be encoded if needed.

Also, the `balance_operator` should be a reference, not a value type.

libraries/chain/database.cpp
Kiwonik: OK, I replaced the assertion with zero amount return. See commit https://github.com/steemit/steem/pull/2356/commits/ca85d02645d2477217f58702e1ade07b66e24bdc
My reasoning is that though putting that assertion in `create_vesting` was a bad idea for reasons you described, but still it is a good place to check SMT's `allow_voting` property and react appropriately, because a) we've got the smt token object in hand and b) it allows to keep the STEEM vs SMT ifology hidden inside the method.

libraries/chain/database.cpp
Kiwonik: Done. See commit https://github.com/steemit/steem/pull/2356/commits/a897d4a2fa11cc33a6d79040f934574a7125af0d

Kiwonik:
@mvandeberg 

1.
> Another thing that should change that was not previous, in the reward balance operators, `get_combined_balance` should take references instead of pointers. Especially for `account_rewards_balance_opbject`. A const point can be dereferenced and the contained object can be modified without breaking the const promised. Using a const reference, the value of the object itself is consted, which is the intended meaning of const in this context. As a general rule, we use references by default and pointers only when necessary (such as a vector of pointers).

OK, I modified `get_combined_balance` parameters and return type so that no pointer (const or not) is used there. See commit https://github.com/steemit/steem/pull/2356/commits/45983b2d18e4a70818fd7eab70ea15d869a57012
Note however that I used `const account_..._object* bo` syntax before which does **not** allow modification of dereferenced object. You must have mistaken it for `account_..._object* const bo` syntax that I never use (and I know nobody who does).

2.
> I do not see any changes in reference to validate methods on objects.

They are still here:
```
-   bool validate() const
-   { return liquid.symbol == vesting.symbol.get_paired_symbol(); }
+   void validate() const
+   { FC_ASSERT( liquid.symbol == vesting.symbol.get_paired_symbol() ); }
```
```
-   bool validate() const
    {
-      return pending_liquid.symbol == pending_vesting_shares.symbol.get_paired_symbol() &&
-             pending_liquid.symbol == pending_vesting_value.symbol;
+   void validate() const
    {
+      FC_ASSERT( pending_liquid.symbol == pending_vesting_shares.symbol.get_paired_symbol() &&
+                 pending_liquid.symbol == pending_vesting_value.symbol );
```

3.
> is_vesting initialization in smt_reward_balance_operator. If this is not wrong, please add comments explaining why it is correct because it certainly looks wrong.

It looked like this at the time @theoreticalbts wrote #2253 issue
```
smt_reward_balance_operator( const asset& value_delta, const asset& share_delta )
      : value_delta(value_delta), share_delta(share_delta), is_vesting(is_vesting == (share_delta.amount.value != 0)) )
```
and now it looks like this
```
smt_reward_balance_operator( const asset& value_delta, const asset& share_delta )
      : value_delta(value_delta), share_delta(share_delta), is_vesting( share_delta.amount.value != 0 )
```
which means that the obvious bug was fixed in the meantime (so a fix is not a part of this PR).

4.
> modify_balance still has a vesting case, which it should not.

In its unified form it can be safely used with any balance with no performance overhead and it helps keeping the code clean by hiding the 'ifology' inside. If you insist on splitting `modify_balance` into `modify_liquid_balance` and `modify_vesting_balance` you clutter the code with
```
if( delta.symbol.is_liquid() )
   ...
else
   ...
```
in every place `modify_balance` is called from. What would be the gain here?

mvandeberg:
With regard to the bool pointer. I don't care if a bool is passed in. All I wanted was that it takes a reference instead of a pointer. Really, this code is only called in two places, I think this interface detracts from readability rather simply checking the balance. I'd like @theoreticalbts input on this.

Another issue with the change to `create_vesting` is that we cannot distinguish the case of no voting from the case where we've lost precision. Are 0 vesting shares being created because the price is not good and we rounded down a small liquid token to 0 or are votes disabled? I think this is a subtle case that could lead to problems. I'd still rather the logic for whether or not a vesting token be created be contained in the calling code. IMO, this is a better abstraction that does not create much, if any, code duplication.

theoreticalbts:
I'm not a fan of the `smt_balance_operator` class, the logic for checking should probably be inline.

Also, agree 100% that any `smt.allow_voting` check does not belong in `create_vesting()`.  The `create_vesting()` method should do ONE thing, which is create the vesting form of an asset from the liquid form.

Kiwonik:
@mvandeberg FYI I shelved this issue/PR until https://github.com/steemit/steem/wiki/Blockchain-Code-Style is agreed as it has direct impact on requests made to this PR.

Kiwonik:
@mvandeberg Regarding the usage of `allow_voting` in `create_vesting`:
1. I do agree that the logic of the method (e.g. returning "special" 0 value) should **not** be based on value of `smt_token_object::allow_voting` for the reasons explained by you 
   > we cannot distinguish the case of no voting from the case where we've lost precision
1. I do agree that future vesting payout code should **not** blindly call `create_vesting` without prior checking (using `smt_token_object::allow_voting`) whether there should be payout at all. This should certainly be the duty of the caller.
1. For reason explained in 2. I insist that this assertion is necessary in appropriate place of `create_vesting`:
   ```FC_ASSERT( smt.allow_voting || to_reward_balance == false, "No voting - no rewards" );```
   exactly because we agreed that it should never happen unless there's a serious bug in caller's code.
   Note that this situation is very similar to what happens in other database method:
   ```
   inline const void database::push_virtual_operation( const operation& op, bool force )
   {
      (...)
      FC_ASSERT( is_virtual_operation( op ) );
   ```

mvandeberg:
The situation with `push_virtual_op` is similar, but not the same.

The assertion is more of a necessary evil than a design decision. We want virtual ops to use the same op pipeline for signals, but they are conceptually a different type entirely.

The more correct way to handle virtual ops would be to create a static_variant union type that signals could subscribe to so that this method itself could be typed properly and the assertion handled during compilation.

That being said, I think I am fine with leaving the assertion in. Please add a comment that it should never fail and it is the caller's responsibility to check the condition before calling `create_vesting`. We want to avoid creating code the incentivizes using exceptions as the preferred form of branching.

Kiwonik:
@mvandeberg Done, see commit https://github.com/steemit/steem/pull/2356/commits/e36504b466c0235240d5dde0da3b57b34c1cb1b5
This makes the output parameter style the last remaining problem of this PR as far as I can see, waiting for conclusion of https://github.com/steemit/steem/wiki/Blockchain-Code-Style discussion.

mvandeberg:
modifying vesting balances and the `smt_balance_operator` classes are currently deeply ingrained into the smt balance logic. For the time being, I believe they are necessary evils and we should consider restructuring all balance operators in the future for consistency and readability.

theoreticalbts:
```
-      FC_ASSERT( delta.amount.value >= 0, "Insufficient SMT ${smt} funds", ("smt", delta.symbol) );
+      FC_ASSERT( balance_operator.is_positive_delta(), "Cannot create a negative balance for SMT ${smt}", ("smt", balance_symbol) );
```

Please don't do this.

```
-      if( delta.amount.value == 0 )
+      if( balance_operator.is_zero_delta() )
```

Or this.

```
auto is_not_negative_is_zero = balance_operator.get_combined_balance_info( *bo );
```

Or this.

mvandeberg:
I need to double check, but I believe the reason for those changes something in templating of the methods because the two types of balances have differently named fields.

mvandeberg:
Outdated

PR #2363, opened 2018-04-19
Concatenation iterator

mariusztrela:
#2289

### DIFF
--- a/libraries/chainbase/include/chainbase/chainbase.hpp
+++ b/libraries/chainbase/include/chainbase/chainbase.hpp
@@ -21,6 +21,7 @@
 #include <boost/throw_exception.hpp>
 
 #include <chainbase/allocators.hpp>
+#include <chainbase/util/object_id.hpp>
 
 #include <array>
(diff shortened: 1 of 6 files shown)


mvandeberg:
"Concatenating data from multiple sources" is not inclusive of all scenarios that need to be included.

Chainbase stores all undo states in three BMIC, created, updated, removed.

When iterating over objects, we cannot simply combining iterators ( an n-way merge sort). We have to consider the meaning behind the different containers. Specifically if an object is modified in such a way that the iteration order changed, we cannot assume that because a certain object is next on one iterator that it should be returned next. There is a notion of temporal locality to how each of the iterators are combined that must be respected to return the actual state of the database.

The implementation in #2363 covers the append only scenario, but the not the use case of the Steem blockchain.

The problem itself would appear to be in the issue, not the PR as the PR does implement the issue. I can merge the PR as is, but most of what is written here is going to be replaced, I am not sure if it is worth merging at this point.

PR #2364, opened 2018-04-19
StorageConfigurationManager

mariusztrela:
#2291

### DIFF
--- a/libraries/utilities/CMakeLists.txt
+++ b/libraries/utilities/CMakeLists.txt
@@ -11,6 +11,7 @@ file(GLOB HEADERS "include/steem/utilities/*.hpp")
 
 set(sources
    benchmark_dumper.cpp
+   storage_configuration_manager.cpp
    key_conversion.cpp
    string_escape.cpp
    tempdir.cpp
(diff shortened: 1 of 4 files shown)


mvandeberg:
I don't like this implementation. I cannot put my finger on exactly what it is, but it doesn't feel like a clean solution to me. I am going to keep thinking about it and provide a better response. In the meantime, I am guessing part of this design was influenced by conversations not recorded on github. Can we expand the requirements in the original issue to better reflect those conversations?

i.e. What do we want the data directory to look like? What sort of configurations options need to go in plugin specific configs?

Is it possible we could implement this using some sort of inheritance/polymorphism?

The entire idea of checking the config for options that match a certain type means that a user can change their configuration with a plugin storage that is nonesense (i.e. `jsonrpc-plugin-storage`) that we now need to code for. One of the benefits of the boost program options is that any option that is not explicitly defined is ignored. This implementation violates that. I'd rather every instance of `*-plugin-storage` be explicitly defined an all others ignored.

mvandeberg:
Outdated

PR #2370, opened 2018-04-20
Rocksd Dev Deployment

### LINKED ISSUE #2301: RocksDB backed Account History returning incorrect results
Running RocksDB Account History in our dev environment does not return correct results. In some instances, account history is missing entirely. I am working on diagnosing the problems. So far I have identified two potential issues.

1. Our shutdown wait time is too short. When steemdsync (or ahnodesync) terminates, it does not wait longer for steemd to exist before compressing the state file. It looks like RocksDB Account History buffers writes and flushes them periodically. Looking at ahnodesync logs, steemd is segfaulting on shutdown. It is doing this for steemdsync as well, so we should consider increasing the timeout to ensure a clean shutdown.

2. The plugin implementation has no support for undo. For now, I am going to suggest the plugin only saves irreversible history. I don't think this will be hugely impactful to Steemit.com at all. All history will be reflected within a minute. For that data in particular, I think the latency is acceptable.

### LINKED ISSUE #2303: RocksDB AH - Save Virtual OP Undo State
The RocksDB Account History plugin needs to save virtual op undo state.

This should be done as a boost multi index container through Chainbase.

The object should be sorted on block and object id.

These objects will be consumed and then removed from memory when blocks become irreversible. (Beyond the scope of this issue)

### LINKED ISSUE #2304: RocksDB AH - Write State on Irreversibility
The RocksDB Account History plugin should only write data out to RocksDB when it becomes irreversible. 

The signal to do this should be `on_applied_block`. This would require the plugin tracking which blocks it has added to the history.

Alternatively, we could add an `on_irreversible` signal that sends the newly irreversible block so that such tracking does not need to be done in the plugin itself.

When the block is made irreversible, all operations in the block and virtual ops stored in memory should be added to the correct histories (#2303)

mvandeberg:
Closes #2301
Closes #2303
Closes #2304

### DIFF
--- a/contrib/startpaassteemd.sh
+++ b/contrib/startpaassteemd.sh
@@ -64,7 +64,7 @@ if [[ "$USE_RAMDISK" ]]; then
   if [[ "$IS_BROADCAST_NODE" ]]; then
     s3cmd get s3://$S3_BUCKET/broadcast-$VERSION-latest.tar.bz2 - | pbzip2 -m2000dc | tar x --wildcards 'blockchain/block*' -C /mnt/ramdisk 'blockchain/shared*'
   elif [[ "$IS_AH_NODE" ]]; then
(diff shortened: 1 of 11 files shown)


theoreticalbts:
I'm no RocksDB expert, but everything looks okay.

PR #2018, opened 2018-01-22
1999 follow plugin

mariusztrela:
#1999

Issue #2044, opened 2018-01-26
Possible JSON serialization change to static_variant


theoreticalbts:
Currently, `static_variant` for `operation` uses strings, all other `static_variant` use integers.

It seems like it would be better to either always use strings, or always use integers.  If we make this breaking change, it seems like the appbase release is the best time to do this, lumping it together with all the other breaking changes.

For `condenser_api` we will probably have to create a legacy type hierarchy for everything that's returned.

Related: #2036

mvandeberg:
There are many small static variants used for extensions. While I don't disagree with this change, the majority of the workload for this issue is the creation of the legacy types that would need to preserve the current serialization. I am hesitant to say we should do this because of how close Appbase is to a release candidate and the delay this would cause.

mvandeberg:
After letting this sit for a few days, I am in agreement with this change. The effect for JSON is that is turns a numeric indexed tuple into a dictionary, which I believe makes much more sense from the perspective of self-documentation.

@goldibex Thoughts?

goldibex:
are we talking about changing the shape of the object altogether? let's document this.

per @theoreticalbts, right now a `static_variant` can take on one of two shapes in json:
```json
[1, { "hello": "world" }]
["transfer", { "from": "foo", "to": "bar", "amount": "50.0 STEEM" }]
```
these 2-tuples are really not at all idiomatic JSON (in fact I'd argue they aren't idiomatic anywhere, even Python). Nevertheless, this is the shape of the object, and there's an awful lot of code out there that depends on this particular shape.

I would propose the following for immediate work:
- When represented in JSON, all `static_variant` objects should be representable using either the numeric index or the string name of the underlying type.
- the current JSON serialization for all objects should be documented and frozen.

There is a much bigger, longer-term conversation we need to have about serialization. I would argue that steemd itself should not be doing anything with JSON at all. But that's outside the scope here.

mvandeberg:
> When represented in JSON, all static_variant objects should be representable using either the numeric index or the string name of the underlying type.

Currently, only the `operation` static variant allows representation using the string name of the underlying type. And it omits `_operation` at the end of the typename (`transfer_operation` in C++ is serialized as `transfer`). And it appears that Condenser is heavily dependent upon it. In #2036 when the wrong op name was returned, Condenser simply spat out raw JSON.

The serialization is inconsistent across static variants and doesn't even follow the rule that would make the most sense when it attempts to follow a somewhat reasonable convention.

[...]

PR #2371, opened 2018-04-20
Revert "1999 follow plugin"

mvandeberg:
Reverts steemit/steem#2018

Dev is not showing re-steems. Creating this revert PR so we can reindex and see if this commit broke it or not. Do not review or merge this PR until further notice.


mariusztrela:
@mvandeberg 
I found the proof that the problem with missing follows is due to commit `2340f2a811c4269a4452b2f6defd96151e73148e` ( `Update condenser_api to use old static_variant JSON serialization #2044` )
You only need to replay 4,500,000 blocks to see the difference. Here are the commands I executed:

```
BRANCH 'develop' - with commit `2340f2a811c4269a4452b2f6defd96151e73148e`
./steemd --replay-blockchain --stop-replay-at-block 4500000
curl --data '{"jsonrpc":"2.0","id":"2","method":"follow_api.get_followers","params":{ "account":"heiditravels", "start":"", "type":"blog", "limit":1000 } }' http://127.0.0.1:8090/rpc
```
```
BRANCH 'develop' - without commit `2340f2a811c4269a4452b2f6defd96151e73148e`
./steemd --replay-blockchain --stop-replay-at-block 4500000
curl --data '{"jsonrpc":"2.0","id":"2","method":"follow_api.get_followers","params":{ "account":"heiditravels", "start":"", "type":"blog", "limit":1000 } }' http://127.0.0.1:8090/rpc
```
When you compare the results of the curl requests, you will see that current develop is lacking a single follower of `heiditravels`, while develop with your commit reverted prints it correctly. 

Here is the missing follower info:

```
{
			"follower": "jamesc",
			"following": "heiditravels",
			"what": ["blog"]
}
```

Details:
I discovered the difference at block `4,481,977`.
The follow operation of `jamesc` was introduced at block `4,450,048`

mariusztrela:
I also checked reverting my commit `1999 follow plugin`, doesn't help. The follower `jamesc` is still missing.

mvandeberg:
Thank you for diagnosing this. Can you move this to its own issue and add it to the back log?

mariusztrela:
Added #2382

Issue #1864, opened 2017-12-08
In SMT create test, use by_symbol index instead of by_nai


youkaicountry:
Details from #1852:

For every asset symbol `x` with an embedded precision value, we can define a related *canonical symbol* `CAN(x)` which is the same symbol, except the embedded precision value set to zero:

```
// return asset symbol with decimal places reset to zero
#define CAN(sym) \
   from_asset_num( sym.asset_num & 0xFFFFFFF0 )
```

Then our constraint becomes:

- If some `smt_token_object` `obj` already exists with `CAN(obj.symbol) == CAN(op.symbol)`, the op should fail.

Now `CAN(x)` has the same order as `asset_symbol_type`.  So it suffices to find the next element in the container greater than or equal to `CAN(x)`, which we can conveniently do with the `lower_bound()` function:

```
const auto& idx = _db.get_index< smt_token_index >().indices().get< by_symbol >();
auto it = idx.lower_bound( CAN( o.symbol ) );
```

If there are any objects in the container with `CAN(obj) == CAN(o.symbol)`, all such objects must be in a group together, and must start at `it`.  So we just trigger the exception when `it->symbol.asset_num == CAN( o.symbol )` like this:

```
FC_ASSERT( (it == idx.end()) || ( CAN( it->symbol.asset_num ) != CAN( o.symbol ) ), "SMT has already been created" );
```

PR #2389, opened 2018-04-26
Removed by_nai index. Hid appropriate smt searching in database methods.

Kiwonik:
#1864
Notes for reviewers:
1. Removed `by_nai` index of `smt_token_object` and used the searching code as specified in issue.
2. Encapsulated searching code in new database methods `get_smt` & `find_smt`.
3. Used the opportunity to get rid of `vesting_liquid_less` comparator from `by_symbol` index of `smt_token_object`.
4. Renamed `by_symbol` index to `by_liquid_symbol` to make it clear what symbol is stored and searched.

### DIFF
--- a/libraries/chain/smt_evaluator.cpp
+++ b/libraries/chain/smt_evaluator.cpp
@@ -17,14 +17,13 @@ namespace {
 /// Return SMT token object controlled by this account identified by its symbol. Throws assert exception when not found!
 inline const smt_token_object& get_controlled_smt( const database& db, const account_name_type& control_account, const asset_symbol_type& smt_symbol )
 {
-   const smt_token_object* smt = db.find< smt_token_object, by_symbol >( smt_symbol );
    // The SMT is supposed to be found.
-   FC_ASSERT( smt != nullptr, "SMT numerical asset identifier ${smt} not found", ("smt", smt_symbol.to_nai()) );
+   const smt_token_object& smt = db.get_smt( smt_symbol );
    // Check against unotherized account trying to access (and alter) SMT. Unotherized means "other than the one that created the SMT".
-   FC_ASSERT( smt->control_account == control_account, "The account ${account} does NOT control the SMT ${smt}",
+   FC_ASSERT( smt.control_account == control_account, "The account ${account} does NOT control the SMT ${smt}",
       ("account", control_account)("smt", smt_symbol.to_nai()) );
    
-   return *smt;
+   return smt;
 }
    
 }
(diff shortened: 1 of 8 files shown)


mvandeberg:
This PR does more than requested by the issue. The issue was simply to remove the `by_nai` index because the NAI is an external representation of the asset. 

We want to use `get< smt_object, by_symbol >` instead of a database specific method for two reasons.

1. This syntax, while slightly verbose, is extremely readable and was made the preferred method of access over a year ago. The use of this method is the exact reason why templates exist. We have many accessor methods that are all the same format. Let's use the templates the way they were designed.

2. We want to avoid adding new "global" methods to the database object. Yes, there are specific methods for other database objects, but they is a remnant of older design practices that we want to avoid moving forward. We simply have not made it a priority to clean up the old code because this design is everywhere and is a time sink to refactor.

Kiwonik:
@mvandeberg 

> We want to use get< smt_object, by_symbol > instead of a database specific method for two reasons.

But we can't use `get< smt_object, by_symbol >` here. That's the whole point of original issue. Of all the accessor methods of database this one **is not** a simple "template wrapper".

mvandeberg:
In the title of the issue,
> use by_symbol index instead of by_nai

And then you delete both `by_symbol` and `by_nai`. You should only have deleted `by_nai`. The reason we want to get rid of `by_nai` and use `by_symbol` instead is that NAI is an external reference to the asset while the symbol is the internal reference.

mvandeberg:
Besides, if we wanted to derive the data the way you are, we would have chosen to make it a const member function instead of the doing the calculations every lookup.

mvandeberg:
Ok, I see what happened here. The issue is old enough that vesting SMTs had not yet been specced.

mvandeberg:
The original `by_symbol` was using a custom comparator so that if I called `get< smt_object, by_symbol >` with either the liquid or vesting symbol it would return the correct symbol. This is the right way to do this.

Kiwonik:
OK, removed get_smt & find_smt methods, restored vesting_liquid_less. The changes are limited to by_nai index removal now.

libraries/chain/smt_evaluator.cpp
```diff
@@ -72,9 +72,12 @@ void smt_create_evaluator::do_apply( const smt_create_operation& o )
 
    // Check that SMT with given nai has not been created already.
    // Note that symbols with the same nai and different precision (decimal places) are not allowed,
-   // therefore we can't use 'by_symbol' index here, as we need to strip the symbol from precision info.
-   FC_ASSERT( (_db.find< smt_token_object, by_nai >( o.symbol.to_nai() ) == nullptr),
-      "SMT ${nai} has already been created.", ("nai", o.symbol.to_nai()));
+   // therefore we use a method that strips the symbol from precision info when searching.
+   const auto& idx = _db.get_index< smt_token_index >().indices().get< by_symbol >();
```
mvandeberg: This is a little weird, but because this is the only place we would refer to an asset symbol without precision, I think it is acceptable. Everywhere else, if the precision does not match, it is not the right asset.

Issue #2205, opened 2018-03-07
appbase witness plugin causes significant resync performance degradation


Gandalf-the-Grey:
During [Appbase Bughunting](https://github.com/steemit/steem/issues/2101) I've noticed significant increase of sync times. Surprisingly they are gone when witness plugin (recommended for all consensus nodes because of bandwidth) is not enabled.
Here I'm listing block and absolute time (minutes) from the start of sync needed to get that far while doing `--resync` comparing stable vs develop with and without witness plugin:
```
block  stable  stable+w  develop  develop+w
  5M     28       30        29        42
 10M     73       77        75       121
 15M    170      182       171       337
 20M    379      415       377       851
```
even more visible on a chart:
![resync_times_chart](https://user-images.githubusercontent.com/6341057/37088849-ee81c538-21fe-11e8-8397-ae67d0e9a194.png)

Gandalf-the-Grey:
Comparing replay time (minutes) on the same machine running appbase with and without witness plugin gives much much smaller difference:
```
block  develop  develop+w
  5M       5         6
 10M      17        18
 15M      44        50
 20M     108       126
```
![replay_times_chart](https://user-images.githubusercontent.com/6341057/37166744-3a7cfd8e-2300-11e8-8664-2b5ec0eb486d.png)

dnotestein:
Without checking the code it's just a guess, but it seems quite possible that signatures aren't validated during replays since they would have been previously checked, but signatures would require validating during a sync. A non-witness node may only be checking the block signatures by the witnesses, while a node running the witness plugin would check the signatures of all the transactions in the block as well.

Gandalf-the-Grey:
If so, that would explain `develop` vs `develop+w` but then I would expect same huge difference with `stable` vs `stable+w` (there was only less than 10% difference for resync time up to 20M block, while on develop it's more than 125% difference)

dnotestein:
I took a quick look at the witness plugin, there's a couple of functions that could be the culprit. To narrow down the problem, I suggest nullifying those functions individually by returning out of them to see what's the culprit (i.e. add a return to kill the function, then run the benchmark with the modified code). I suggest you try a reindex first before a resync, just to be sure the modified code doesn't complete.

I would start by adding a return here: 

https://github.com/steemit/steem/blob/42e2d95ec09d1695ec1b392d47a2e44612815cf0/libraries/plugins/witness/witness_plugin.cpp#L221

[...]

PR #2393, opened 2018-04-27
change moment of p2p::set_block_production call in witness plugin

mkochanowicz:
Issue #2205

### DIFF
--- a/libraries/plugins/witness/witness_plugin.cpp
+++ b/libraries/plugins/witness/witness_plugin.cpp
@@ -65,7 +65,14 @@ namespace detail {
       block_production_condition::block_production_condition_enum block_production_loop();
       block_production_condition::block_production_condition_enum maybe_produce_block(fc::mutable_variant_object& capture);
 
-      bool _production_enabled = false;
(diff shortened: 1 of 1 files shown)


abitmore:
IMHO this changed too much.

The original logic was: if witness_plugin is enabled, while syncing, the node will validate signatures of all transactions.

After this change, before caught up (in sync), the node won't validate signatures.

mkochanowicz:
@abitmore Node start validate signatures after sync completed. Is it wrong?

abitmore:
@mkochanowicz: block producing nodes (whose witness list is non-empty) should validate every block it received. We favor security over performance in this case.

Non-block-producing nodes (whose witness list is empty) can skip signature validation for better performance while syncing.

mkochanowicz:
@abitmore So, I can do sync with empty witnesses list (non-block-producing node), then restart `steemd` with built state and with non-empty witnesses list (block-producing node) - if I we talk about pre-abbase. My code changes is simply shortcut for such solution.

Maybe change in appbase is intentional, to disable such scenario? @mvandeberg ?

PR #2402, opened 2018-05-01
Boost 1.60+ Compatibility

### LINKED ISSUE #2344: Explore Upgrading to Boost 1.62
This is an exploratory issue to cost out potential future work.

Boost 1.62 is the default boost included in Ubuntu 18.04 LTS. We should determine what incompatibilities exist and the path to upgrade if we so choose.

I know fc threads are broken. Perhaps we can vendor in the specific libraries needed for fc threads and everything else can use Boost 1.62 as a stop gap until we can replace the p2p code.

mvandeberg:
Closes #2344

This works on Mac OS with Boost 1.67

The changes here are largely taken from: 
https://github.com/bitshares/bitshares-fc/pull/5/files
https://github.com/bitshares/bitshares-fc/commit/652389644c67c14dff96f4dd4d0afa9a1265ee09#diff-c9fbd6d968e071c0cbf584a61b9c0f3c

And reverting some changes made in:
https://github.com/steemit/fc/commit/e7d0d26fe2e6559400771efa989f87a6d5d5391e#diff-238ae51a67a8d3f7015b77a9c6f82c2c

The added thread tests pass building with both Boost 1.60 and 1.67


mvandeberg:
To answer the inevitable question regarding fc tests not being a part of CI: #2403

PR #2372, opened 2018-04-20
generate_block uses batch write queue

mvandeberg:
Closes #2347

PR #2404, opened 2018-05-01
Fix recursive write lock in generate_block

### LINKED ISSUE #2398: [appbase] "Lock timeout, moving to lock n" on block generation
```
steem_blockchain_version: 0.19.4
steem_git_revision:       7ecbe6b5a0ed49f6987f0a23f3d2c747697fcf9f
fc_git_revision:          7ecbe6b5a0ed49f6987f0a23f3d2c747697fcf9f
```
Testing `2347-generate-blocks-batch-write` branch, I get:
`Lock timeout, moving to lock n` everytime my node generates a block. N does increment everytime.

eg, last message:
```
Lock timeout, moving to lock 473
2557040ms witness_plugin.cpp:433        block_production_loo ] Generated block #22026046 with timestamp 2018-04-30T17:42:36 at time 2018-04-30T17:42:36
```

mvandeberg:
Closes #2398

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -813,64 +813,61 @@ signed_block database::_generate_block(
 
    signed_block pending_block;
 
-   with_write_lock( [&]()
+   //
+   // The following code throws away existing pending_tx_session and
+   // rebuilds it by re-applying pending transactions.
+   //
+   // This rebuild is necessary because pending transactions' validity
+   // and semantics may have changed since they were received, because
+   // time-based semantics are evaluated based on the current block
+   // time.  These changes can only be reflected in the database when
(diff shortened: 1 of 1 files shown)


theoreticalbts:
Removing this lock looks unsafe.  We're writing to the database with `_apply_transaction` and messing with the undo state.  Why is it safe to do that without holding the lock?

mvandeberg:
PR #2372 updated the interface by which we generate blocks. All writes need to go through the write queue. But the lock was never removed so when a `generate_block` is called it now recursively grabs the lock.

BhuzOr:
Tested on my witness node.
No `Lock timeout` log anymore and produced blocks normally as expected.

PR #2405, opened 2018-05-01
Implement additional per-block stats

theoreticalbts:
This code improves the `block_stats` plugin, and adds stats export from the witness node.  This accomplishes two objectives:

- Get bandwidth / reserve ratio data out of the witness plugin
- Show the "proper" way to use the `block_stats` plugin from another plugin


mvandeberg:
The build failure is a reflection issue

PR #2426, opened 2018-05-05
make it possible to specify blocklog url (stable)

sneak:


mvandeberg:
PR is against an old build of stable.

PR #2427, opened 2018-05-05
Reduce the comment interval from 20 seconds to 3 seconds

mejustandrew:
Changed the value of STEEM_MIN_REPLY_INTERVAL to 3.
Adjusted the assert message to match the new value.


mvandeberg:
Changing the value in this manner could fork the network. It needs to be done as part of a hard fork.

PR #2434, opened 2018-05-08
Revert "Exchange node config shared file size to 54G"

andreabarry1:
Reverts steemit/steem#2357


mvandeberg:
Why should this be reverted?

Issue #2019, opened 2018-01-22
removal of 20s comment limit


roadscape:
Users have expressed that having to wait 20 seconds between comments can be too much. IIRC, accounts can only make one post per block, so effectively the lowest comment interval is 3s.

A 3-second interval combined with bandwidth limits seems quite reasonable; we should consider removing the 20s rule.

iamsmooth:
In addition to being annoying even with the current UI, the hard coded blockchain limit fails to be UI agnostic, as certain types of UIs or blockchain applications would suffer even more from a 20 second limit. I agree with removing it and just having the 3s+bandwidth limits.

sneak:
👍

PR #2439, opened 2018-05-08
Reduce the comment interval from 20 seconds to 3 seconds

mejustandrew:
I am not sure on which branch these changes should be committed or if I missed something with this implementation so sorry and thanks for the review and the rest of the information!


libraries/chain/steem_evaluator.cpp
```diff
@@ -600,7 +600,14 @@ void comment_evaluator::do_apply( const comment_operation& o )
             FC_ASSERT( _db.calculate_discussion_payout_time( *parent ) != fc::time_point_sec::maximum(), "Discussion is frozen." );
       }
 
-      if( _db.has_hardfork( STEEM_HARDFORK_0_12__176 ) )
+      if( _db.has_hardfork( STEEM_HARDFORK_0_20__2019 ) )
+      {
+         if( o.parent_author == STEEM_ROOT_POST_PARENT )
+             FC_ASSERT( ( now - auth.last_root_post ) > STEEM_MIN_ROOT_COMMENT_INTERVAL, "You may only post once every 5 minutes.", ("now",now)("last_root_post", auth.last_root_post) );
+         else
+             FC_ASSERT( (now - auth.last_post) > STEEM_MIN_REPLY_INTERVAL_HF20, "You may only comment once every 3 seconds.", ("now",now)("auth.last_post",auth.last_post) );
```
TimCliff: A similar comparison is being done here to allow one vote every three seconds:
https://github.com/steemit/steem/blob/bb8ca476f33c08434274f2d76333fcb74cfa58b8/libraries/chain/steem_evaluator.cpp#L1229

With the vote compassion, it it is using `>=` instead of `>`. Would this need to be changed to `>=` in order to allow one comment per block (every 3 seconds)?

mejustandrew: You have a very good point, and I guess the comparison could be changed this way more to be consistent with the vote comparison, but from a functional point of view `(now - auth.last_post) > STEEM_MIN_REPLY_INTERVAL_HF20` only allows one comment to be made per block, because it is something like `time since last comment must be strictly greater than the block time`.

I will update the PR as you guys think, after all, you know best!

mvandeberg:
Related: #2019

PR #2447, opened 2018-05-14
Statsd Plugin

### LINKED ISSUE #2276: Integrate a StatsD client
To lay groundwork for exporting metrics from steemd, a StatsD protocol client should be identified and integrated.

## User Stories

I am a developer that wants to time the execution of a critical function in steemd and send it to a statsd server so that I can optimize the function.

I am an admin that only wants to record API call counts.

I am an admin that wants to log statsd rather than send messages to a statsd client. 

### Requirements

- Easy to integrate into new and existing code. (Probably Macro based)

- Runtime configurable on output destination (UDP, file, stderr, etc.)

- Runtime configurable on logged statistics.

- Must be low impact and non-blocking.

Some candidates:
https://github.com/talebook/statsd-client-cpp
https://github.com/vthiery/cpp-statsd-client

mvandeberg:
Closes #2276

Tested via branch `2338-log-write-times`.

Logging config for console/file logs works.
Confirmed UDP logging via Wireshark.
Blacklist and whitelist of metrics and namespaces work.


theoreticalbts:
LGTM

PR #2448, opened 2018-05-14
Log Chain Plugin Write Times

### LINKED ISSUE #2338: Add logged stats for `push_block` and `push_transaction` execution times
We need to better tune the interaction between read and write threads. To do this, we need to know two important pieces of information.

1. What is an acceptable latency to return results for a read request?
2. How long does it currently take to perform our longest write?

The answer to the first will come from the Applications team.

The second, steemd needs to supply.

In the write queue thread, let's record execution times for `push_block` and `push_transaction` and log them via the debug logger. For now, let's log every block and transaction. We will update later to reduce the amount of spam from transaction logging, but I want to get a good distribution of transaction application times for probabilistic analysis.

In addition to logging how long it takes to perform each write, I also want to know how many of each types of writes we do on each acquisition of the write lock so that we can calculate an expected write lock time, which will be used to determine what the read lock timeout should be with a reasonable expected error rate from lock timeouts.

mvandeberg:
Closes #2338

Should be reviewed and merged after #2447


theoreticalbts:
LGTM

Issue #2469, opened 2018-05-18
total_missed does not correctly increment in a corner case


theoreticalbts:
[This code](https://github.com/steemit/steem/blob/2ae9a31fa456d68242dca1dcc068a912b42b21a4/libraries/chain/database.cpp#L3405-L3419) skips updating `total_missed` in case `witness_missed.owner == b.owner`.  To fix the bug, transform from:

```
if( witness_missed.owner != b.witness )
{
   modify(witness_missed, [&](witness_object& w)
   {
      w.total_missed++;
      if( has_hf_14 ) { if( last_old_enough ) { shutdown_witness } }
   }
}
```

to:

```
modify(witness_missed, [&](witness_object& w)
{
   w.total_missed++;
   if( witness_missed.owner != b.witness )
   {
      if( has_hf_14 ) { if( last_old_enough ) { shutdown_witness } }
   }
}
```

We can collapse these three checks to a single `if` statement with a conjunction (`AND`) of these three conditions, reordered any we like, and I like this ordering:

```
modify(witness_missed, [&](witness_object& w)
{
   w.total_missed++;
   if( last_old_enough
     && witness_missed.owner != b.witness
     && has_hf_14 )
   { shutdown_witness }
}
```

The idea behind this reordering is that, by deferring the hardfork check (involving an object lookup) until the last, we might well avoid it entirely if one of the other checks fails (the other two checks are both only integer arithmetic and reading fields from objects already available).

I'd really like to get rid of the `witness_missed.owner != b.owner` check, but I realized we cannot remove it, otherwise the network might be bricked if all of the witnesses miss enough blocks to get shut down, then start producing again.

theoreticalbts:
@abitmore raises the point that this patch might be "cruel" to some witnesses by increasing reported `total_missed`.  For that reason, we've decided not to change the functionality.

I will therefore close this ticket and PR #2472, and remove this commit from PR #2473.

PR #2472, opened 2018-05-18
Fix corner case where total_missed is not properly updated #2469

theoreticalbts:
PR for #2469


abitmore:
This is a "cruel" fix.

Before this fix, the witness that produced the first block after **revived** a halted chain will get a good-looking block missing number which is somewhat a prize. IMHO that's why we didn't "fix" it before.

theoreticalbts:
Good point, closing.

PR #2479, opened 2018-05-21
Replaced 20s comment limit with 3s (1 block)

Kiwonik:
Implemented according to similar hardforked changes in neighboring code. #2019


libraries/protocol/include/steem/protocol/config.hpp
```diff
@@ -113,7 +113,8 @@
 #define STEEM_VOTE_DUST_THRESHOLD             (50000000)
 
 #define STEEM_MIN_ROOT_COMMENT_INTERVAL       (fc::seconds(60*5)) // 5 minutes
-#define STEEM_MIN_REPLY_INTERVAL              (fc::seconds(20)) // 20 seconds
+#define STEEM_MIN_REPLY_INTERVAL_PRE_HF20     (fc::seconds(20))   // 20 seconds
```
mvandeberg: The naming scheme we have used for such changes is to name when a particular constant was introduced. See `STEEM_MAX_VOTED_WITNESSES` as an example.

Here, the names should be `STEEM_MIN_REPLY_INTERVAL_HF0` and `STEEM_MIN_REPLY_INTERVAL_HF20`.

libraries/chain/steem_evaluator.cpp
```diff
@@ -613,14 +613,24 @@ void comment_evaluator::do_apply( const comment_operation& o )
          if( o.parent_author == STEEM_ROOT_POST_PARENT )
              FC_ASSERT( ( now - auth.last_root_post ) > STEEM_MIN_ROOT_COMMENT_INTERVAL, "You may only post once every 5 minutes.", ("now",now)("last_root_post", auth.last_root_post) );
          else
-             FC_ASSERT( (now - auth.last_post) > STEEM_MIN_REPLY_INTERVAL, "You may only comment once every 20 seconds.", ("now",now)("auth.last_post",auth.last_post) );
+         {
+            if( _db.has_hardfork( STEEM_HARDFORK_0_20__2019 ) )
```
mvandeberg: Add a todo to remove this check after HF 20. Because we are reducing the interval from 20 to 3, after HF 20 we can do all old checks as 3 seconds and they will pass. Anything to help keep the codebase clean!

libraries/chain/steem_evaluator.cpp
```diff
@@ -613,14 +613,24 @@ void comment_evaluator::do_apply( const comment_operation& o )
...
+         }
       }
       else if( _db.has_hardfork( STEEM_HARDFORK_0_6__113 ) )
       {
          if( o.parent_author == STEEM_ROOT_POST_PARENT )
              FC_ASSERT( (now - auth.last_post) > STEEM_MIN_ROOT_COMMENT_INTERVAL, "You may only post once every 5 minutes.", ("now",now)("auth.last_post",auth.last_post) );
          else
-             FC_ASSERT( (now - auth.last_post) > STEEM_MIN_REPLY_INTERVAL, "You may only comment once every 20 seconds.", ("now",now)("auth.last_post",auth.last_post) );
+         {
+            if( _db.has_hardfork( STEEM_HARDFORK_0_20__2019 ) )
```
mvandeberg: This case in not needed

youkaicountry:
This was already implemented in #2439. Closing.

PR #2484, opened 2018-05-22
Constant VESTS to STEEM Conversion rate

TianyangUnVig:
I have noticed that the conversion rate of STEEM to VESTS has been decreasing over time. I'm not sure why or if it's intentional. While looking at this, I noticed that during the block production, 15% of the inflation goes to a non-existing "vesting fund". The total_vesting_fund_steem gets increased with the 15% but total_vesting_shares remains the same, leading to a change of the conversion rate.

The 15% inflation actually goes nowhere that any user can get but gets counted in virtual supply and current supply, which doesn't seem to make sense.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -1994,7 +1994,6 @@ void database::process_funds()
 
       modify( props, [&]( dynamic_global_property_object& p )
       {
-         p.total_vesting_fund_steem += asset( vesting_reward, STEEM_SYMBOL );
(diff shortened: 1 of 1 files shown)


mvandeberg:
This is intentional.

`total_vesting_fund_steem` is the total STEEM backing all VESTS. Rather than think of the conversion rate decreasing, the amount of STEEM each VESTS is worth is increasing over time. When you power up your STEEM into VESTS and then power it back down, you receive slightly more STEEM than when you started. 

This is explained in the [Steem Whitepaper](https://steem.io/steem-whitepaper.pdf).

PR #2492, opened 2018-05-23
Increase Delegation Cooldown

### LINKED ISSUE #2428: Possible reward pool rape by cycling delegations
1) It is easier to explain the bug by showing you my "mindstorm" drawing:

[image](https://snag.gy/EHObz9.jpg)

2) If you can't read what I've shown on image:

There is possibility to take advantage of reward pool almost by x2 of somebody's SP.

User A delegates to B
After delegations VPb should be averaged by VPa% and SP accordingly, because delegation can be undone immediately, and VP can be used down to 0 very fast before that. 

If you use all your VP in few minutes or voting every day only 10 times 100% - it doesnt matter, the sum is the same.

Because of that You can:
- user A vote all power
- delegate to user B all the sp
- user B vote all power as fast as he can (!)
- undelegate to user B
- wait 7 days (in this timeline, VP recovers in 5 days also)
- repeat!

Outcome:
In about 8 days two users USED 100% vp power so used 200% power of some amount of SP.

Taking math into consideration thye shouldn be allowed to do that in such time, it should take them 14 days.

Solutions:

not ideal (you can bomb small user with VP0% to mute ;) ): 
- average the VP of user B in the moment of delegation with VP of A

not very handy: 
- add 7 days lock period to undelegate function

https://steemit.com/@rafalski

mvandeberg:
Closes #2428


roadscape:
Is it feasible to work around the issue by adjusting vote power appropriately instead of increasing delegation cooldown? In some cases this approach is ideal, and it possibly be applied to this case too.

For instance, when a user with 5 SP and 0% VP powers up 5000 SP, it makes sense their VP would increase to ~99.9%. Otherwise they have to wait days to leverage their purchase.

mvandeberg:
The solution you are suggesting does not apply only to delegation but to powering up as well. And the behavior happens currently. While this PR does not address your concern it doesn't make it any worse.

mvandeberg:
For clarity, if we implemented your proposal, I would still want to merge this PR. Your solution does address a rough spot in the current delegation/power up logic, but it does nothing to address the problem stated in #2428 without adding significant complication.

theoreticalbts:
Looks good to me

PR #2493, opened 2018-05-23
Enforce max limit order expiration of 28 days

### LINKED ISSUE #1449: Market orders should expire
We should minimize users' ability to post things that require long-lived consensus bandwidth.  Thus, all orders in the on-chain market should be required to have an expiration time that is not too far in the future, say at most `head_block_time() + MAX_ORDER_EXPIRATION_TIME`.

At the hardfork time arrives, all currently existing orders are converted to expire at `head_block_time() + MAX_ORDER_EXPIRATION_TIME`.  Thus, ancient orders will eventually be cleaned off the books.  Anyone user who wishes to have their orders continue to exist can simply re-create them.

I'm thinking `MAX_ORDER_EXPIRATION_TIME` should be 72 hours.

Thoughts?

mvandeberg:
Closes #1449


theoreticalbts:
This doesn't implement the complete spec, which is to enforce the new expiration rules on both new *and existing* orders.

So when HF20 arrives, we need to do some one-time hardfork processing.  Specifically, we need to get the last order `by_expiration` and, if its expiration time is after `new_limit = head_block_time() + MAX_ORDER_EXPIRATION_TIME`, set its expiration time to `new_limit` and repeat, until the expiration of the last order `by_expiration` is less than or equal to `new_limit`.

theoreticalbts:
Actually, the PR does implement it by a different means, it rewrites the order times when the orders are created.

This code could theoretically fork the network if *all* of the following unlikely events occur:

- HF20 doesn't occur within 28 days of its scheduled time
- The witnesses are split between HF19 and HF20 versions when that 28-day mark occurs
- Some of the affected orders match

PR #2509, opened 2018-05-29
Begin implementation of resource credit system

theoreticalbts:
No d-pools yet, and I may want to change the curve to promote basic stability issues.  But the basic architecture is there, it compiles, I've manually tested that this syncs, and the most obvious bugs have been fixed.

### DIFF
--- a/libraries/jsonball/data/resource_parameters.json
+++ b/libraries/jsonball/data/resource_parameters.json
@@ -1,0 +1,34 @@
+[
+ ["resource_history_bytes",
+  {
+   "time_unit" : "blocks",
+   "budget_time" : {"days" : 30},
+   "budget" : "15000000000",
+   "half_life" : {"days" : 15},
+   "drain_time" : {"hours" : 1},
+   "p_min" : 100.0
+  }
+ ],
+
+ ["resource_new_accounts",
+  {
+   "time_unit" : "seconds",
+   "budget_time" : {"days" : 30},
+   "budget" : 1000000,
+   "half_life" : {"days" : 20},
+   "drain_time" : {"hours" : 1},
+   "p_min" : 1000.0
+  }
+ ],
+
+ ["resource_market_bytes",
+  {
+   "time_unit" : "blocks",
+   "budget_time" : {"days" : 30},
+   "budget" : "3000000000",
+   "half_life" : {"hours" : 16},
+   "drain_time" : {"hours" : 1},
+   "p_min" : 500.0
+  }
+ ]
+]
--- a/libraries/jsonball/data/resource_parameters.json
+++ b/libraries/jsonball/data/resource_parameters.json
@@ -3,10 +3,12 @@
   {
    "time_unit" : "blocks",
    "budget_time" : {"days" : 30},
-   "budget" : "15000000000",
+   "budget" : "30000000000",
    "half_life" : {"days" : 15},
-   "drain_time" : {"hours" : 1},
-   "p_min" : 100.0
+   "a_point_num" : 37,
+   "a_point_denom" : 1000,
+   "u_point_num" : 1,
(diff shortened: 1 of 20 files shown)


theoreticalbts:
I may be doing a substantial refactor and change in the implementation soon, this probably shouldn't be merged just yet.

libraries/jsonball/templates/jsonball/rc_params.cpp.j2
```diff
@@ -0,0 +1,13 @@
...
+const char resource_parameters[] = {
+{{ resource_parameters|json_to_char_array }}, 0
+};
+
+const char* get_resource_parameters()
+{
+   return resource_parameters;
+}
+
+} }
```
mvandeberg: Missing newline

libraries/jsonball/data/resource_parameters.json
```diff
@@ -0,0 +1,34 @@
...
+   "drain_time" : {"hours" : 1},
+   "p_min" : 100.0
+  }
+ ],
+
+ ["resource_new_accounts",
+  {
+   "time_unit" : "seconds",
+   "budget_time" : {"days" : 30},
+   "budget" : 1000000,
```
mvandeberg: This budget needs to be witness votable.

The voting is NYI, but the behavior of the curve needs to not be tied to a specific budget.

libraries/plugins/rc/rc_plugin.cpp
```diff
@@ -0,0 +1,373 @@
...
+      if( export_plugin != nullptr )
+      {
+         ilog( "Registering RC export data factory" );
+         export_plugin->register_export_data_factory( STEEM_RC_PLUGIN_NAME,
+            []() -> std::shared_ptr< exportable_block_data > { return std::make_shared< exp_rc_data >(); } );
+      }
+
+      chain::database& db = appbase::app().get_plugin< steem::plugins::chain::chain_plugin >().db();
+
+      my->_post_apply_block_conn = db.add_post_apply_block_handler( [&]( const block_notification& note ){ my->on_post_apply_block( note ); }, *this, 0 );
```
mvandeberg: We need to have a conversation regarding whether RC based filtering should happen pre or post tx.

If we do it pre-tx then we can filter out a potentially costly computation. However, this requires refactoring the `get_or_create_rc_account_object` because we might create objects for accounts that don't exist in consensus. This creates the opposite scenario where RC calculations are the more costly computation.

libraries/plugins/rc/resource_user.cpp
```diff
@@ -0,0 +1,30 @@
+
+#include <steem/plugins/rc/resource_user.hpp>
+
+#include <steem/protocol/transaction.hpp>
+
+namespace steem { namespace plugins { namespace rc {
+
+using namespace steem::protocol;
+
+account_name_type get_resource_user( const signed_transaction& tx )
```
mvandeberg: This is a departure from how bandwidth is charged today. Subtle to most users, but the decision should be purposeful.

This method returns a single account to charge bandwidth to where the current algorithm charges it to all parties involved in the transaction.

theoreticalbts:
Rebased on latest `develop` and updated with `manabar` changes.  Merge #2558 first.

theoreticalbts:
Besides implementing the manabar, there are a few architecture changes which were necessitated.  Before, I tracked `usage` which was the empty portion of the manabar.  It suffered from numerous corner cases, but its main advantage was that it could be lazily implemented.

Now all account creation occurs through the `create_rc_accounts()` method which compares the states of the two tables and creates mirror `rc_account`

theoreticalbts:
- [ ] Implement SP tracking
- [ ] Write unit test
- [ ] RC regen time should be less than or equal to vote regen time

PR #2513, opened 2018-05-30
Revert account creation fee modification

### LINKED ISSUE #1761: Revert Modified Account Creation Fee
When we added `create_account_with_delegation_operation` we changed the meaning of the witness voted account creation fee to be the minimum STEEM needed as part of a creation with a delegation. The creation with 100% STEEM is 30x the account creation fee. This change should be reverted for HF 20 so the witness voted account creation fee is the actual fee.

mvandeberg:
Closes #1761


theoreticalbts:
Looks reasonable.

PR #2538, opened 2018-06-08
Prevent repeat writing of irreversible state in account history RocksDB

### LINKED ISSUE #2496: double entries on appbase get_account_history
Just one example, there are others both for my account and others account 

s = Steem(['https://api.steemitdev.com'])
```
>>> print(json.dumps(s.get_account_history('bhuz',567518,1),indent=4))
[
    [
        567517,
        {
            "trx_id": "0000000000000000000000000000000000000000",
            "block": 22337509,
            "trx_in_block": 4294967295,
            "op_in_trx": 0,
            "virtual_op": 4,
            "timestamp": "2018-05-11T13:19:45",
            "op": [
                "curation_reward",
                {
                    "curator": "bhuz",
                    "reward": "4367.999036 VESTS",
                    "comment_author": "steempostitalia",
                    "comment_permlink": "postit-n-71"
                }
            ]
        }
    ],
    [
        567518,
        {
            "trx_id": "0000000000000000000000000000000000000000",
            "block": 22337509,
            "trx_in_block": 4294967295,
            "op_in_trx": 0,
            "virtual_op": 4,
            "timestamp": "2018-05-11T13:19:45",
            "op": [
                "curation_reward",
                {
                    "curator": "bhuz",
                    "reward": "4367.999036 VESTS",
                    "comment_author": "steempostitalia",
                    "comment_permlink": "postit-n-71"
                }
            ]
        }
    ]
]
```

mvandeberg:
Closes #2496

### DIFF
--- a/libraries/plugins/account_history_rocksdb/account_history_rocksdb_plugin.cpp
+++ b/libraries/plugins/account_history_rocksdb/account_history_rocksdb_plugin.cpp
@@ -409,6 +409,17 @@ public:
 
          const auto& rocksdb_plugin = appbase::app().get_plugin< account_history_rocksdb_plugin >();
 
+         // I do not like using exceptions for control paths, but column definitions are set multiple times
+         // opening the db, so that is not a good place to write the initial lib.
+         try
+         {
+            get_lib();
+         }
+         catch( fc::assert_exception& )
+         {
+            update_lib( 0 );
+         }
+
          _on_post_apply_operation_con = _mainDb.add_post_apply_operation_handler(
             [&]( const operation_notification& note )
             {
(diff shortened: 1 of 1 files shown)


theoreticalbts:
`FC_ASSERT( s.IsNotFound() )` looks suspicious, what is this line of code trying to do?

theoreticalbts:
It seems like the indexing of the columns is determined by their insertion order.  Could the conditional column creation based on `addDefaultColumn` cause problems?

mvandeberg:
> `FC_ASSERT( s.IsNotFound() )` looks suspicious, what is this line of code trying to do?

I believe this is trying to mimic the behavior of our `get_X` db methods that throw an exception when an object is not found. It is done in other rocksdb get methods and is replicated here for internal consistency. Agreed that it is a little clunky, but this plugin is a prototype of how we might use RocksDB in a more generic manner.

> It seems like the indexing of the columns is determined by their insertion order. Could the conditional column creation based on `addDefaultColumn` cause problems?

It does not appear to be causing problems because our column families are accessed via numeric index, not name. However, I think it can probably be removed (Default column family appears to be a part of compatibility with an old RocksDB interface). It was not touched in this PR, so I'm not sure we should block this PR to change this.

theoreticalbts:
Not a big fan of this fix, but it works and this is code that won't be in there long-term once we have the `multi_index_container` replacement.

PR #2541, opened 2018-06-12
Claim Account Creation

### LINKED ISSUE #1771: Discount Account Creation
Creating an account normally requires matching the account creation fee. A discounted account will always have a fee of 0 STEEM. If an account is created in this matter, a discounted account creation token should be consumed and fail if no tokens remain.

mvandeberg:
Closes #1771


theoreticalbts:
Looks ready to merge.

PR #2551, opened 2018-06-14
Increase the minimum transfer amount

Jolly-Pirate:
Make it more expensive for spammers to send memos.

### DIFF
--- a/libraries/protocol/steem_operations.cpp
+++ b/libraries/protocol/steem_operations.cpp
@@ -159,7 +159,7 @@ namespace steemit { namespace protocol {
       validate_account_name( from );
       validate_account_name( to );
       FC_ASSERT( amount.symbol != VESTS_SYMBOL, "transferring of Steem Power (STMP) is not allowed." );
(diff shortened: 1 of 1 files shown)


iamsmooth:
Bad idea. Not long ago 0.1 STEEM was $1.20 which is far too high a limit. 

Second, any issue people have with memos should be dealt with in UI, not by impairing the ability to transfer funds.

Smittyvb:
This would also cause issues for services that send 0.001 STEEM/SBD messages as verification responses.

Also, this would cause a hardfork. You'd need to add a HF flag for this, and make this happening conditional on the hardfork flag being true.

iamsmooth:
> This would also cause issues for services that send 0.001 STEEM/SBD messages as verification responses 

I think that is somewhat by design, if sending messages via memos is considered too spam-prone. But you raise a good point about needing input from portions of the community before doing something like this.

> Also, this would cause a hardfork 

It could be considered a soft fork (not sure if it was implemented appropriately for that though) as has been the care for other anti-spam measures.

Anyway irrelevant now that it is close. Maybe open an issue to discuss?

PR #2557, opened 2018-06-15
Fix get_pending_payout when tags_api is not enabled

### LINKED ISSUE #2556: develop condenser_api not returning `pending_payout_value` with tags_api disabled.
The previous issue #2535 to remove hard requirements on the tags_api removed a bit too much in `set_pending_payout`. The tags_api is only needed for setting promoted value on the content. Everything else can and should execute without problem.

mvandeberg:
Closes #2556


theoreticalbts:
Looks like that should be `if( _tags_api )`

theoreticalbts:
Looks good now

PR #2558, opened 2018-06-15
RC support code

theoreticalbts:
This is some miscellaneous support code needed by the RC plugin.  Since it stands on its own, I thought it might be better to put it in a separate PR to make the RC plugin a smaller, easier-to-review diff.

### DIFF
--- a/libraries/fc/include/fc/int_array.hpp
+++ b/libraries/fc/include/fc/int_array.hpp
@@ -46,9 +46,8 @@ template<typename T,size_t N> struct get_typename< fc::int_array<T,N> >
 template<typename T, size_t N>
 void from_variant( const variant& var, fc::int_array<T,N>& a )
 {
-   const variants& vars = var.get_array();
    memset( a.data, 0, sizeof(a.data) );
-   variants& varray = var.get_array();    // throws if is not array
+   const variants& varray = var.get_array();    // throws if is not array
    FC_ASSERT( varray.size() == N );
 
    for( size_t i=0; i<N; i++ )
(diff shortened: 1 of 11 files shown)


libraries/fc/include/fc/int_array.hpp
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Like fc::array, but does zero-initialization and has friendly JSON format
+ */
+
+template<typename T, size_t N>
+class int_array
+{
+    public:
+       int_array(){ memset( data, 0, sizeof(data) ); }
+
```
mvandeberg: We should allow instantiation of an `int_array< T, N >` from `array< T, N >`.

For example, the witness schedule in `witness_schedule_object` to printing in `api_witness_schedule_object`.

libraries/fc/include/fc/int_array.hpp
```diff
@@ -0,0 +1,80 @@
...
+      return _name.c_str();
+   }
+};
+
+template<typename T, size_t N>
+void from_variant( const variant& var, fc::int_array<T,N>& a )
+{
+   const variants& vars = var.get_array();
+   memset( a.data, 0, sizeof(a.data) );
+   variants& varray = var.get_array();    // throws if is not array
```
mvandeberg: This line results in a compilation error in clang.

`libraries/fc/include/fc/int_array.hpp:51:14: error: binding value of type 'const vector<...>' to reference to type 'vector<...>' drops 'const' qualifier`

theoreticalbts:
See if this fixes it.

As far as `fc::array` usage in on-the-wire places and objects, I quickly checked and I don't think there's anyplace where it is natural to do this.

PR #2561, opened 2018-06-18
Discount Account Creation

### LINKED ISSUE #1770: Discount Account Creation Inventory
#1765 includes a daily discount account limit parameter. The number of discounted account creations needs to be tracked.

SMTs have logic for regular token emission. We could do a similar system to emit these "tokens" at regular intervals based on the voted parameter.

mvandeberg:
Closes #1770


theoreticalbts:
So currently, the account creation fee is *self-describing* -- the amount listed in the op is the amount that's actually charged to the account.

However, `claim_account_operation` has the following line, which makes the effective fee state-dependent:

```
paid_fee.amount.value = std::min( o.fee.amount.value, creation_fee.amount.value );
```

This breaks the self-describing nature of the fee.  A better check would be to replace this line with:

```
FC_ASSERT( o.fee.amount.value == creation_fee.amount.value );
```

This both makes the fee self-describing, and preserves the intent of preventing malfunctioning clients from overpaying.  (It is arguable whether the latter function is something the blockchain ought to do at all. 
 If not, we should just delete the `else` clause entirely.)

theoreticalbts:
There are a few changes I'd like to make this code.  Let's merge this first as a starting point.

theoreticalbts:
Also, merge conflicts.

mvandeberg:
I am currently ironing out the kinks in some unit tests based on your previous feedback.

PR #2564, opened 2018-06-19
Update statsd interface

### LINKED ISSUE #2521: Update Statsd Plugin Interface
The original macro based interface takes namespace arguments as plaintext and stringifies them. Implementation of #2510 necessitates different string key values at the same place in the code. #2510 also requires another timer macro, which includes this requirement. All other macros should be updated to follow this format.

mvandeberg:
Closes #2521

### DIFF
--- a/libraries/plugins/chain/chain_plugin.cpp
+++ b/libraries/plugins/chain/chain_plugin.cpp
@@ -108,9 +108,9 @@ struct write_request_visitor
 
       try
       {
-         STATSD_START_TIMER( chain, write_time, push_block, 1.0f )
+         STATSD_START_TIMER( "chain", "write_time", "push_block", 1.0f )
          result = db->push_block( *block, skip );
-         STATSD_STOP_TIMER( chain, write_time, push_block )
+         STATSD_STOP_TIMER( "chain", "write_time", "push_block" )
       }
       catch( fc::exception& e )
       {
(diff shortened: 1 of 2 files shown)


theoreticalbts:
It seems strange to me that the `STATSD_STOP_TIMER()` macro now throws away the `NAMESPACE`, `STAT`, `KEY` arguments.  Shouldn't it pass those to `reset()` so `statsd_timer` knows which timer to stop?

mvandeberg:
The original macro used the values to create a unique variable name. 
`NAMESPACE ## STAT ## KEY ## _timer.reset();`

This does not work now that NAMESPACE, STAT, and KEY are strings. I suppose we could put the timers in a global map indexed my a NAMESPACE, STAT, KEY tuple. The intent behind the original design was to handle destruction of the timer to automatically send the time so you could do something like this:

``` C++
void some_function()
{
   STATSD_START_TIMER( "benchmark", "foobar", "some_function" )
   . . .
}
```

And the whole function would be timed with a single macro.

Indexing the timer into a global map needs careful consideration. What if the function in question can be called by read processes? Then we need to make it thread safe. Currently all timers are on the local stack, thus no thread safety is needed. The statsd plugin itself handles thread safety for writes to the network in a non-blocking manner.

theoreticalbts:
Okay, as long as the one-timer-per-scope limitation is a conscious design decision, not an accident, we can merge this.

It will be enforced by the compiler anyway, since multiple invocations of `STATSD_START_TIMER()` at the same scope would re-declare the `statsd_timer` variable.

PR #2570, opened 2018-06-23
Manabar for voting power

### LINKED ISSUE #2539: Small double voting vulnerability (5%) in power down event at depleted voting strength.
This issue is related to https://github.com/steemit/steem/issues/2428 in the way that I believe both vulnerabilities to stem from what I believe is the same underlying idiomatic design shortcoming.

This is a smaller double voting vulnerability than the one with a delegation.  

Basically, the same problem that exists with a delegation at close to zero % VS also applies to the following sequence:

* Deplete VS
* Power-down event
* Power up with powered down amount to a different account.

Where the resulting double voting with delegation could be as big as 42%, the power down scenario only allows a little over 5% double voting. 

The below blog post describes:

* The problem scenario
* Why the two scenario's are actually the same underlying problem
* How potentially making all operations "sum of the product of SP and VS" constant could lay the foundation for weeding out any such issues at a fundamental level

https://steemit.com/steem/@mattockfs/an-other-double-voting-bug-truly-fixing-steem-double-voting-issues

mvandeberg:
Closes #2539


theoreticalbts:
Why is this against `master`?  Shouldn't it be against `develop`?

theoreticalbts:
I went ahead and changed the base branch.

theoreticalbts:
First, `hf20_vote_evaluator()` is missing `FC_ASSERT( abs_rshares > 0 )`.  Is this intentional?

Second, `claim_reward_balance_evaluator` adds `a.voting_manabar.current_mana += op.reward_vests.amount.value;`.  There are multiple problems with this:

- This patch repurposes `voting_manabar.current_mana` to replace `voting_manabar.current_mana` for pre-HF20 computations.  (Whether or not this the best design is arguable, but it's what this patch does.)  So this line probably needs an HF check so it only executes post-HF20.
- The `claim_reward_balance_evaluator` changes `vesting_shares`, so it needs to call `regenerate_mana()`.
- For consistency with the rest of the patch, this should probably be `regenerate_mana()` followed by `use_mana( -op.reward_vests );`

Third, in `delegate_vesting_shares()`, I have doubts about the computation of `available_shares`.  Specifically, `current_mana` is a manabar whose maximum value is of course `get_effective_vesting_shares()`, which is defined as `vesting_shares + received_vesting_shares - delegated_vesting_shares - shares_about_to_power_down`.  The intent here is to only allow powered-up VESTS to be delegated.  So the computation should actually be much simpler, namely, `available_shares = current_mana - received_vesting_shares` (which may, in general, be negative, which is fine -- the subsequent `available_shares` check will then forbid any increase in delegation).

Fourth, the patch is inconsistent in a couple other places about whether to use `current_mana += x` or `use_mana( -x )`.

theoreticalbts:
So upon reading the patch, I had some insights.

First, some terminology:  Let's divide our VESTS into "mana-charged" VESTS that contain mana, and "uncharged" VESTS that do not.  Then (assuming no blocks have passed since `regenerate_mana()` was last called), `current_mana` is simply the number of mana-charged VESTS, and `get_effective_vesting_shares() - current_mana` is the number of uncharged VESTS.

To prevent "free voting power" exploits, we want to enforce conservation of mana:  Mana cannot be created or destroyed, except by the passage of time (as computed by `regenerate_mana()`).

Of course, `get_effective_vesting_shares()` is a four-term sum:  `vesting_shares + received_vesting_shares - delegated_vesting_shares - shares_about_to_power_down`.  The two positive terms are different "lumps" of VESTS owned by the account, and the two negative terms are "debits" against these two lumps.

The rules implemented by this patch are as follows:

- (a) The two debit terms cannot be taken against `received_vesting_shares`.
- (b) Mana is allocated to `received_vesting_shares` first.
- (c) Only mana-charged VESTS may be delegated.
- (d) When un-delegating, mana-charged VESTS are preferred to un-charged VESTS.
- (e) Un-charged VESTS may be un-delegated.

mvandeberg:
I want to hold off on merging this. I just thought of a corner case regarding delegation that I want to investigate first.

theoreticalbts:
The main thing here is adding a new clamp:

```
available_shares.amount = std::min( available_shares.amount, max_mana - delegator.received_vesting_shares.amount );
```

The `if` statement following this clamp should be deleted entirely:

- The `if()` branch takes out the vesting shares about to power down, but these have *already been taken out* since `max_mana = vesting_shares + received_vesting_shares - delegated_vesting_shares - shares_about_to_power_down`.
- The `else` branch *sets* `available_vesting_shares` to `vesting_shares - delegated_vesting_shares - shares_about_to_power_down`, which is of course just `max_mana - received_vesting_shares`.  So it does almost the same thing as the newly added clamp statement -- but *it doesn't clamp to `current_mana`*.  Which means if the `else` clause is hit, you can delegate no-mana shares.

mvandeberg:
The `else` clause is the pre-hardfork branch. And yes, you can delegate no-mana shares, which is what we are fixing.

mvandeberg:
> The if() branch takes out the vesting shares about to power down, but these have already been taken out since max_mana = vesting_shares + received_vesting_shares - delegated_vesting_shares - shares_about_to_power_down.

`shared_about_to_power_down` is only the next week of powered down shares. The semantics of delegation mean that you cannot delegate any VESTS that are a part of a power down. It would not make sense for me to power down all of my VESTS, then delegate them to you. Perhaps we could implement the logic to automatically remove the delegation from you over time, but that is an added layer of complication that we have decided to avoid for the time being. The comment in this block explains what is happening. (Edited for remove reference to defunct power shares name)

```
/*
current voting mana does not include the current week's power down:

std::min(
   account.vesting_withdraw_rate.amount.value,           // Weekly amount
   account.to_withdraw.value - account.withdrawn.value   // Or remainder
   );

But an account cannot delegate **any** VESTS that they are powering down.
The remaining withdrawal needs to be added in but then the current week is double counted.
*/
```

The next line

``` c++
available_shares += delegator.vesting_withdraw_rate - asset( delegator.to_withdraw - delegator.withdrawn, VESTS_SYMBOL );
```

Removes the entire power down, but then we've removed `delegator.vesting_withdraw_rate` twice, so we add it back. There is a bug here that if the account is in the last week of power down then we add in `vesting_withdraw_rate` when only `to_withdraw - withdrawn` was removed. The min check needs to be done in this block as well.

theoreticalbts:
Checked out the latest changes, the only issue I have is maybe `a.voting_manabar.use_mana( op.reward_vests.amount.value )` should be `a.voting_manabar.use_mana( -op.reward_vests.amount.value );` ?

mvandeberg:
FIxed

PR #2572, opened 2018-06-25
Symmetric Vote Lockout

### LINKED ISSUE #1267: Symmetrical/gradual vote lockout
Previously discussed in https://github.com/steemit/steem/issues/900

The lockout period could be a linearly ramping restriction on rshares granted from a vote.. e.g. from 0% to 100% over the last 12 hours on a post. It should affect downvotes and upvotes the same.

mvandeberg:
Closes #1267

#2570 should be reviewed first


theoreticalbts:
As this is dependent on #2570, I am adding the BLOCKED label.

mvandeberg:
Working on resolving merge conflicts

theoreticalbts:
```
-            FC_ASSERT( _db.head_block_time() < comment.cashout_time - STEEM_UPVOTE_LOCKOUT_HF17, "Cannot increase payout within last twelve hours before payout." );
+            FC_ASSERT( _db.head_block_time() < comment.cashout_time - STEEM_UPVOTE_LOCKOUT_HF17, "G." );
```

- Looks like leftover debugging code, can we revert this change?
- Maybe `cashout_dt` instead of `cashout_delta`?
- How confident are we that comments will always be pruned before `comment.cashout_time` ?  AFAIK since that pruning is done in post-block processing, we could have a vote op get into the chain after its comment's nominal expiration.
- I think we should simply `FC_ASSERT( now < comment.cashout_time );` near the beginning of the method.
- If we don't do that, we should have some way of sanely handling overflow:

```
uint32_t cashout_delta = ( comment.cashout_time - _db.head_block_time() ).to_seconds();
```

theoreticalbts:
> The following code at the beginning of the function handles the general case

That code only handles the case where it's infinite (i.e., equal to `time_point::maximum()` which we're using as a kind of "infinity" sentinel value here).  What if it's finite and greater?

Specifically, I'm thinking that maybe we have the following sequence of events:

- Witness A produces a block at `1:00:00`.
- Comment expires at `1:00:03`
- Witness B produces a block at `1:00:06` which includes an upvote.

I looked through the code and I *think* this scenario can't happen (basically `update_global_dynamic_data()` which updates the time is done post-transactions, so the `head_block_time()` observed by the upvote evaluator would still be `1:00:00`; no user ops can sneak into the gap between updating the time and expiring comments that are in the past as of the update.)

But for safety I think it's still better to add a failsafe check to sanely handle a negative result.

mvandeberg:
Is it better to fail the vote or allow it and create a delta lower bound of 0?

theoreticalbts:
Looks good to me

PR #2583, opened 2018-07-03
Fix negative vesting withdrawals

mvandeberg:


Smittyvb:
Won't this cause forking among non-upgraded witnesses if another such attack were to happen?

PR #2599, opened 2018-07-06
Virtual op refactoring

theoreticalbts:
Currently, virtual ops are implemented by directly modifying the DB in per-block processing methods, and then pushing a virtual op after the fact.

Really what ought to be done is (1) have the virtual op generated based on state, (2) pre-notify, (3) do all state modification by applying the virtual op, (4) post-notify.

That's not what this patch does.  The main immediate need for this kind of refactoring is to use vops for pre/post notification of all vesting changes, so the RC plugin can do its thing with the old value of vesting shares.

So the `create_vesting()` function gets the vesting share price, computes how much to create, and then modifies the balance and supply fields.  This patch hacks `create_vesting()` so that it has a callback after computing how much to create, but before anything is modified.

Now what is done is ideally (1) call `create_vesting()` before state is modified, (2) have the callback issue the pre-op, (3) do the state modification, (4) post-notify.  I say "ideally" because, in some cases, state is modified beforehand.  But since the only immediate pre/post op consumer that cares is the RC plugin, and it only cares about vesting, as long as these are in the proper place with respect to vesting, modifying other state before the pre-notify doesn't matter to the RC plugin.

This will let us get HF20 out the door, but this code will probably need to be replaced before HF21.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -1153,7 +1153,7 @@ std::pair< asset, asset > database::create_sbd( const account_object& to_account
 // we modify the database.
 // This allows us to implement virtual op pre-notifications in the Before function.
 template< typename Before >
-asset create_vesting2( database& db, const account_object& to_account, asset liquid, bool to_reward_balance, Before&& f )
+asset create_vesting2( database& db, const account_object& to_account, asset liquid, bool to_reward_balance, Before&& before_vesting_callback )
 {
    try
    {
@@ -1186,7 +1186,7 @@ asset create_vesting2( database& db, const account_object& to_account, asset liq
          price vesting_share_price = to_reward_balance ? smt.get_reward_vesting_share_price() : smt.get_vesting_share_price();
          // Calculate new vesting from provided liquid using share price.
          asset new_vesting = calculate_new_vesting( vesting_share_price );
-         f( new_vesting );
+         before_vesting_callback( new_vesting );
          // Add new vesting to owner's balance.
          if( to_reward_balance )
             db.adjust_reward_balance( to_account, liquid, new_vesting );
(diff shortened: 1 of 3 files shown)


libraries/chain/database.cpp
```diff
@@ -1172,18 +1181,19 @@ asset database::create_vesting( const account_object& to_account, asset liquid,
       {
          FC_ASSERT( liquid.symbol.is_vesting() == false );
          // Get share price.
-         const auto& smt = get< smt_token_object, by_symbol >( liquid.symbol );
+         const auto& smt = db.get< smt_token_object, by_symbol >( liquid.symbol );
          FC_ASSERT( smt.allow_voting == to_reward_balance, "No voting - no rewards" );
          price vesting_share_price = to_reward_balance ? smt.get_reward_vesting_share_price() : smt.get_vesting_share_price();
          // Calculate new vesting from provided liquid using share price.
          asset new_vesting = calculate_new_vesting( vesting_share_price );
+         f( new_vesting );
```
mvandeberg: Can this be a different name? Reading through this I thought it was a bad and buggy conditional (`if( new_vesting );`)

libraries/chain/database.cpp
```diff
@@ -1206,17 +1216,18 @@ asset database::create_vesting( const account_object& to_account, asset liquid,
       FC_ASSERT( liquid.symbol == STEEM_SYMBOL );
       // ^ A novelty, needed but risky in case someone managed to slip SBD/TESTS here in blockchain history.
       // Get share price.
-      const auto& cprops = get_dynamic_global_properties();
+      const auto& cprops = db.get_dynamic_global_properties();
       price vesting_share_price = to_reward_balance ? cprops.get_reward_vesting_share_price() : cprops.get_vesting_share_price();
       // Calculate new vesting from provided liquid using share price.
       asset new_vesting = calculate_new_vesting( vesting_share_price );
+      f( new_vesting );
```
mvandeberg: Same as above

PR #2555, opened 2018-06-15
Closes #2022 - Add beneficiaries SBD payout

TimCliff:
This PR closes issue #2022 by updating the beneficiaries payout to use the same payout setting as the author. For example, if the author selects the `50% / 50%` payout setting for the post, then all of the beneficiary payouts will be paid using the `50% / 50%` payout setting. This is a change from the current behavior which always pays beneficiaries using `100% SP`.

RPC nodes with the `legacy_comment_benefactor_reward_operation` or `comment_benefactor_reward_operation` options enabled will return data for those virtual operations in a different format when this change is applied. Instead of a single `reward` asset, there will now be three: `sbd_payout`, `steem_payout`, and `vesting_payout`.

Issue #2022, opened 2018-01-22
Community Request - Reward Beneficiaries should be paid based on payout setting


TimCliff:
Currently if an author chooses 50/50 rewards for their post/comment, and they include beneficiaries, the beneficiaries do not receive their rewards as 50/50. This should be updated so that the beneficiaries payout is the same as the author’s.

joticajulian:
Yes, I agree. To be fair, also curators should receive it in 50/50. Because currently, with the high prices of sbd, the authors are receiving around 92% of the total payment (see [my proposal and math details](https://steemit.com/utopian-io/@jga/steem-proposal-possibility-to-receive-50-50-sbd-sp-for-curators)).

Zappl:
We at Zappl also Support's this idea as well. Currently, for new platform developers, 13 weeks power down is going to be hard on paying server bills or developers if you're not popular with the Steem community. Lucky enough most of us devs have good communities but the current distribution of payment for Beneficiaries is enough to not want to make a project.

It's not really clear to developers how much they can't depend on post rewards from their users. Considering most projects make a large number of their earnings from their own update posts.

Even 25/75 is okay in our eyes.

hrosspet:
I agree beneficiary's payout should follow if the author chooses 50% SBD / 50% SP.

@mvandeberg : any updates if this has been discussed at steemit team?

TimCliff:
I propose the following changes to update the beneficiaries payout to use the same payout setting as the comment author:

In `libraries/protocol/hardfork.d/0_20.hf`, add a new HF20 condition:
```
#define STEEM_HARDFORK_0_20__2022 (STEEM_HARDFORK_0_20)
```

In `libraries/chain/database.cpp`, inside the `cashout_comment_helper` method, update the beneficiaries payout to give the same portion of rewards in SP/SBD/STEEM as the author payout:
```
            for( auto& b : comment.beneficiaries )
            {
               auto benefactor_tokens = ( author_tokens * b.weight ) / STEEM_100_PERCENT;

               if( has_hardfork( STEEM_HARDFORK_0_20__2022 ) )
               {
                              auto benefactor_sbd_steem     = ( benefactor_tokens * comment.percent_steem_dollars ) / ( 2 * STEEM_100_PERCENT ) ;
                              auto benefactor_vesting_steem = benefactor_tokens - benefactor_sbd_steem;
[...]

PR #2607, opened 2018-07-09
#2022 Fix Boost Tests

TimCliff:
This PR fixes the Boost tests that were broken in https://github.com/steemit/steem/pull/2555.

- The two tests to validate that the `bob` beneficiary was not paid in STEEM/SBD were removed, since beneficiary payments in STEEM/SBD are allowed now.
- The test that validates the beneficiary rewards for `bob` + `sam` are equal to the beneficiary payout for the `alice` comment were updated to include the SBD payout as part of the beneficiary rewards amounts.
- The two tests that validate the payout amounts for `bob` and `sam` are correct when compared to the reward for `alice` were updated to include the SBD payout as part of the beneficiary rewards amounts.

Note: The previous version of the test cases were written assuming that the liquid portion of the author payout was 100% SBD, and did not include a portion of liquid STEEM as well. I.E. they did not include `.reward_steem_balance.amount` in any of the calculations.
```
db->get_account( "alice" ).reward_sbd_balance.amount + db->get_account( "alice" ).reward_vesting_steem.amount
```

The new version of the test cases are written in the same way.

### DIFF
--- a/tests/tests/operation_tests.cpp
+++ b/tests/tests/operation_tests.cpp
@@ -6720,11 +6720,9 @@ BOOST_AUTO_TEST_CASE( comment_beneficiaries_apply )
 
       generate_block();
 
-      BOOST_REQUIRE( db->get_account( "bob" ).reward_steem_balance == ASSET( "0.000 TESTS" ) );
-      BOOST_REQUIRE( db->get_account( "bob" ).reward_sbd_balance == ASSET( "0.000 TBD" ) );
-      BOOST_REQUIRE( db->get_account( "bob" ).reward_vesting_steem.amount + db->get_account( "sam" ).reward_vesting_steem.amount == db->get_comment( "alice", string( "test" ) ).beneficiary_payout_value.amount );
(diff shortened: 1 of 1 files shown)


mvandeberg:
In addition to these changes, the proportion of SBD:SP payout should be checked for each beneficiary.

TimCliff:
Just an update - I'm still working on this. 

I've tried this:
```
      BOOST_REQUIRE( db->get_account( "bob" ).reward_vesting_steem.amount == db->get_account( "bob" ).reward_sbd_balance.amount );
      BOOST_REQUIRE( db->get_account( "sam" ).reward_vesting_steem.amount == db->get_account( "sam" ).reward_sbd_balance.amount );
```

but for some reason, it is failing on the `sam` case. (The `bob` case is passing.) I'll continue working on it tonight/tomorrow.

TimCliff:
Updates made.

The reason the `sam` case was failing was because the reward amount could not be split exactly equally between SP and SBD due to precision / rounding. `sam` had a `reward_vesting_steem` amount of `15965` and a `reward_sbd_balance` amount of `15964`. The solution was to +1 to the `reward_sbd_balance` side of the equation. This is in line with the existing tests, which do the same thing.

PR #2615, opened 2018-07-11
Move start_write_processing

### LINKED ISSUE #2613: Chain plugin write processing thread startup
The start of the write processing thread should happen at the end of `plugin_startup`, not the beginning. It currently busy waits the entirety of a reindex. It should slow down the reindex but gives the illusion of parallel reindexing that is not occurring.

mvandeberg:
Closes #2613


theoreticalbts:
Looks reasonable

Issue #2621, opened 2018-07-11
Debug fixture shouldn't use debug_update() to create vests


theoreticalbts:
[This code](https://github.com/steemit/steem/blob/97640ae24e9b8820b1c7194d20638e1b51693779/tests/db_fixture/database_fixture.cpp#L485-L501), which creates VESTS directly for unit tests, is creating false positives of the new `rc_plugin` code which finds unexpected changes to VESTS (which indicate a bug in the plugin).

So let's replace this method with a transfer from `initminer` followed by a normal vesting op.

One problem I encountered implementing this, is that some tests, e.g. `withdraw_vesting_apply`, will mix [this method](https://github.com/steemit/steem/blob/97640ae24e9b8820b1c7194d20638e1b51693779/tests/tests/operation_tests.cpp#L1472) with [other debug updates](https://github.com/steemit/steem/blob/97640ae24e9b8820b1c7194d20638e1b51693779/tests/tests/operation_tests.cpp#L1564-L1580) which causes the op to fail when the debug plugin re-pushes the op with less strict skip flags.

mvandeberg:
It turns out this solution is something I just came to as a work around to get tests working for #1770

PR #2622, opened 2018-07-11
Don't use debug_update in vest() override #2621

theoreticalbts:
PR for issue #2621


theoreticalbts:
Your change looks fine

PR #2630, opened 2018-07-13
742 to filter the history of user operations

RostislavVel:
To filter the history of user operations. #742.


mvandeberg:
In order to make a PR from Golos to Steem, you'll need to back port the change specifically for Steem.

PR #2634, opened 2018-07-13
Additional rc_plugin implementation

theoreticalbts:
New `rc_plugin` features:

- Skip flags which allows easily (de)activating different ways of handling various error conditions
- Stricter checking for unexpected account creation
- Visitor-based implementation of `get_resource_user()` with special cases for a few ops
- Pre/post visitor to regenerate worker accounts
- Pre-apply checks that vesting has not changed since post-apply of last operation, call this the "unexpected delta vesting check" or UDV check.  (A failed UDV check indicates a bug in the plugin logic.)
- `rc_plugin::validate_database()` executes the UDV check on all accounts

Integration of `rc_plugin` with unit tests:

- Skip flags set by database fixture so the UDV check will throw an exception, so any failed UDV check triggered by the unit tests will automatically result in failing CI
- On test completion, database fixture `validate_database()` calls `rc_plugin::validate_database()` to run the UDV check on all accounts created during the unit test

### DIFF
--- a/libraries/plugins/rc/rc_plugin.cpp
+++ b/libraries/plugins/rc/rc_plugin.cpp
@@ -13,7 +13,7 @@
 
 #include <steem/jsonball/jsonball.hpp>
 
-#define STEEM_RC_REGEN_TIME   (60*60*24*15)
+#define STEEM_RC_REGEN_TIME   (60*60*24*5)
 
 namespace steem { namespace plugins { namespace rc {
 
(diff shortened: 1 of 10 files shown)


theoreticalbts:
This PR is based on #2633; merge #2633 first.

libraries/plugins/rc/rc_plugin.cpp
```diff
@@ -440,11 +430,335 @@ void rc_plugin_impl::on_first_block()
...
+   {
+      regenerate( op.author );
+   }
+
+   void operator()( const curation_reward_operation& op )const
+   {
+      regenerate( op.curator );
+   }
+
+   // Is this one actually necessary?
```
mvandeberg: Probably not. But an account that is active since before the reward balance might have a different of a few satoshi.

mvandeberg:
I am assuming the testing of the RC plugin was checking it against all of the other unit tests.

PR #2637, opened 2018-07-13
Fix block size corner case on generation

### LINKED ISSUE #2632: Got two "Block Size is too Big" errors running v0.19.10
You can see the relevant logs here:

https://gist.github.com/lukestokes/7421398c19efacb0f31ee6315f3fff83

```
{"next_block_num":24111132,"block_size":65537,"max":65536}
{"next_block_num":24117399,"block_size":65540,"max":65536}
```

My hunch is v0.19.10 is doing something strange to bring in more transactions than it should and is blowing up when it tries to produce the block.

I may switch back to my v0.19.5 producing server if this continues.

mvandeberg:
Closes #2632

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -834,11 +834,38 @@ signed_block database::_generate_block(
    if( !(skip & skip_witness_signature) )
       FC_ASSERT( witness_obj.signing_key == block_signing_private_key.get_public_key() );
 
-   static const size_t max_block_header_size = fc::raw::pack_size( signed_block_header() ) + 4;
-   auto maximum_block_size = get_dynamic_global_properties().maximum_block_size; //STEEM_MAX_BLOCK_SIZE;
-   size_t total_block_size = max_block_header_size;
-
    signed_block pending_block;
+
+   pending_block.previous = head_block_id();
+   pending_block.timestamp = when;
+   pending_block.witness = witness_owner;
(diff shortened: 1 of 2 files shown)


theoreticalbts:
This should fix the problem.  Approved.

libraries/chain/database.cpp
```diff
@@ -834,11 +834,40 @@ signed_block database::_generate_block(
...
+      }
+   }
+
+   // The 4 is for the max size of the transaction vector length
+   size_t total_block_size = fc::raw::pack_size( pending_block ) + 4;
    auto maximum_block_size = get_dynamic_global_properties().maximum_block_size; //STEEM_MAX_BLOCK_SIZE;
-   size_t total_block_size = max_block_header_size;
 
-   signed_block pending_block;
+   idump( (total_block_size) );
```
lukestokes: Is this a debugging statement? Sorry for the newbie question, I couldn't find much about what "idump" does and was curious.

mvandeberg: Yes, I thought I had removed them. `idump` is info level, dump object. There is also `ilog` for log at info level. And `ddump`, `wdump`, and `edump` for debug, warning, and error and their corresponding logs as well.

lukestokes: Thanks so much for explaining! :)

PR #2640, opened 2018-07-13
Fix reindex on manabar refactor

mvandeberg:
The type of `elapsed_seconds` was changed and had major impacts on the results of the voting algorithm. Also, the updating of `last_update_time` was changed to only require one db read.


theoreticalbts:
Looks good

PR #2645, opened 2018-07-16
To filter the history of user operations #742

RostislavVel:


mvandeberg:
I will repeat what I said last time.

> In order to make a PR from Golos to Steem, you'll need to back port the change specifically for Steem.

Golos and Steem have diverged enough that you cannot make PRs directly from Golos to Steem.

PR #2657, opened 2018-07-18
Burn Account Creation Fee

### LINKED ISSUE #1762: Burn Account Creation Fee
Instead of powering up the account creation fee to the new account, it should be "burned" by transferring it to the `null` account.

mvandeberg:
Closes #1762


theoreticalbts:
Approved.

PR #2662, opened 2018-07-20
RC Plugin Account Inventory Feedback

### LINKED ISSUE #2628: rc_plugin should take feedback from consensus state for discount account pool size
The consensus code implements its own pool for discounted account creation.  The non-consensus pool used by the RC plugin for discount account creation should be copied from this consensus pool every block, so as the consensus pool depletes, the non-consensus RC price signal reflects the true remaining amount of discounted account creations which would be allowed by consensus.

The effect is to ensure that, as long as sufficient numbers of p2p nodes / witnesses are running the RC plugin, high RC requirements will apply a continuous soft limit to discount account creation, before the consensus hard-limit starts rejecting discount account creations.

mvandeberg:
Closes #2628

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -2310,7 +2310,8 @@ void claim_account_evaluator::do_apply( const claim_account_operation& o )
       if( !_db.is_pending_tx() )
       {
          const auto& current_witness = _db.get_witness( gpo.current_witness );
-         FC_ASSERT( current_witness.schedule == witness_object::elected, "Subsidized accounts can only be claimed by elected witnesses" );
+         FC_ASSERT( current_witness.schedule == witness_object::elected, "Subsidized accounts can only be claimed by elected witnesses. current_witness:${w} witness_type:${t}",
+            ("w",current_witness.owner)("t",current_witness.schedule) );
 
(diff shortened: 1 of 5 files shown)


theoreticalbts:
```
pool = _db.get_dynamic_global_properties().available_account_subsidies / STEEM_ACCOUNT_SUBSIDY_PRECISION;
```

You need to multiply by `params.resource_unit` in order for this to have the correct units.

```
"half_life" : {"days" : 1},
```

If the intent here is to disable decay, it would be better to set `"days" : 1000000` or something.

mvandeberg:
> change the assert() (which dangerously turns into a no-op in release builds) to FC_ASSERT() (which safely enforces its condition regardless of build settings).

That was the whole point of using `assert`. I want to enforce the check on the debug build in CI, which will fail CI if it is not true and then skip it in prod. If I could `static_assert` against a variable, I would.

Likewise, my intent behind hind that implementation (which contains a bug as you mentioned) was to optimize out the multiplication and division. Again, if we could optimize math between a constant and a variable (that just so happens to be constant) I would.

We can bring up the argument of pre-mature optimization, but us having the conversation is now burning more time and developer resources than it is worth. Because I have to make a change in the commit I'll implement in the generic manner. If I had gotten it right, I might fight this. But again, the argument is not worth the time.

theoreticalbts:
Looks good.

PR #2665, opened 2018-07-20
Add logic to switch from bandwidth to rc_plugin for HF 20

### LINKED ISSUE #2468: Fix `broadcast_transaction_synchronous` dup reversion
The assertion to prevent duplicate transactions was reverted in the recent merge to `master`.

``` c++
FC_ASSERT( _callbacks.find( txid ) == _callbacks.end(), "Transaction is a duplicate" );
```

c5f4e16b520f9d71f9a0d949016e393fee8fad84 was merged to `master`, but an merge was never made to `develop`, so when #2278 was implemented, this check was not moved to `condenser_api`.

### LINKED ISSUE #2648: Update Witness Plugin for RC Plugin
The witness plugin contains all of the bandwidth logic which is to be deprecated. The RC Plugin is to be used instead.

The old bandwidth from the witness plugin needs to be toggle-able via a command line argument. The argument will define if the witness plugin actively rejects transactions due to bandwidth or not. Even when disabled the plugin will continue to track bandwidth for the time being. 

The RC Plugin needs to be made a dependency of the witness plugin.

mvandeberg:
Closes #2648

### DIFF
--- a/libraries/plugins/rc/include/steem/plugins/rc/rc_plugin.hpp
+++ b/libraries/plugins/rc/include/steem/plugins/rc/rc_plugin.hpp
@@ -35,6 +35,8 @@ class rc_plugin : public appbase::plugin< rc_plugin >
       virtual void plugin_shutdown() override;
 
       void set_rc_plugin_skip_flags( rc_plugin_skip_flags skip );
+      const rc_plugin_skip_flags& get_rc_plugin_skip_flags() const;
+
       void validate_database();
 
    private:
--- a/libraries/plugins/rc/rc_plugin.cpp
+++ b/libraries/plugins/rc/rc_plugin.cpp
@@ -764,7 +764,12 @@ void rc_plugin_impl::validate_database()
 rc_plugin::rc_plugin() {}
 rc_plugin::~rc_plugin() {}
 
-void rc_plugin::set_program_options( options_description& cli, options_description& cfg ){}
+void rc_plugin::set_program_options( options_description& cli, options_description& cfg )
+{
+   cfg.add_options()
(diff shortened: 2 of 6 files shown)


mvandeberg:
What needs to be added is a way to set `skip_reject_not_enough_rc` from the config. For whatever I assumed there was a way to do this, but upon further review that does not appear to be the case.

Additionally, if we want to prevent overly restrictive semantics, we can perform and xor between !`rc_skip_reject_not_enough_rc` and `witness-enforce-bandwidth` with the default being `rc_ski_reject_not_enough_rc=false` and `witness_enforce_bandwidth=false` when HF 20 activates.

theoreticalbts:
The `plugin_initialize` check in `witness_plugin`, in pseudocode, is essentially:

```
witness_plugin._enforce_bandwidth != rc_plugin.skip_reject_not_enough_rc
```

Unfortunately, these two flags have opposite "polarity":

- `_enforce_bandwidth = true` means the `witness_plugin` will check bandwidth
- `skip_reject_not_enough_rc = false` means the `rc_plugin` will check bandwidth

The "proper" fix, architecturally, is probably to refactor so that both flags the same "polarity" [1].

The "quick hack" fix is to just change the inequality from `!=` to `==`, although it makes the code hard to follow (so a short comment explaining the situation would not be amiss).

My suggestion is to put the "quick hack" fix in this PR, then make a separate issue/PR for the flag rename / repolarization.

[1]   Since this code's never been in a numbered release, we should be able to just rename the config option without breaking anyone's deployments.  We won't have this luxury once the code makes it into a numbered release.

mvandeberg:
The amount of work to get #2703 implemented over the quick fix was negligible. This PR now has the fix described in #2703 as well.

theoreticalbts:
Looks like it's ready to merge

Issue #2563, opened 2018-06-19
RC Bandwidth API plugin


youkaicountry:
Implement RC bandwidth system RPC API plugin

PR #2678, opened 2018-07-30
Implement rc_api_plugin #2563

theoreticalbts:
PR for issue #2563

### DIFF
--- a/libraries/plugins/apis/rc_api/CMakeLists.txt
+++ b/libraries/plugins/apis/rc_api/CMakeLists.txt
@@ -1,0 +1,23 @@
+file(GLOB HEADERS "include/steem/plugins/rc_api/*.hpp")
+add_library( rc_api_plugin
+             rc_api_plugin.cpp
+             rc_api.cpp
+           )
+
+target_link_libraries( rc_api_plugin rc_plugin json_rpc_plugin )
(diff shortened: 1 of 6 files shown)


theoreticalbts:
> works well for a small number of pools, but if it increases

The number of pools is three.  (Insert Monty Python reference here.)

Each resource type must have its own entry in the `rc_resource_types` enum, and custom coding in `resource_count.cpp` to measure how many of the new resource a transaction consumes.  Which means a new resource always means code changes and a new release.

*Even if* we restrict it to a subset of pools, we can add a parameter for that in a future release.  This is backwards compatible:  Old servers will return all of the pools, instead of only pools that the client asked for.  Sanely coded clients should ignore pools they don't care about.

theoreticalbts:
Force-pushed to fix a bug.

mvandeberg:
> The number of pools is three.

We have plans for more already before release and the plugin is designed to support more.

> Sanely coded clients should ignore pools they don't care about.

This thinking worries me. It is **not** a problem with this API, but pushed to its extreme leads us to calls like `get_state`.

> we can add a parameter for that in a future release.

I am approving because of this aspect of the current API design. In future APIs I think it should be a goal that the client can make an API call to retrieve only the data they need (granular to the object). @youkaicountry To note my proposal to make this an API design goal.

Issue #2679, opened 2018-07-31
Disable rc_plugin op processing until after first block


theoreticalbts:
The `rc_plugin` throws an error when a startup hardforks occur (e.g. on a private testnet).  The reason for the error is because the `rc_account` objects for genesis accounts are created by `on_first_block()` which processes the first block.  However, `on_first_block()` is triggered in a post-block handler, meaning ops in the first block are missed (i.e. virtual ops generated by hardfork processing).

Instead, let's disable op processing entirely in the first block.

PR #2682, opened 2018-07-31
rc_plugin:  Ignore operations until first block is processed #2679

theoreticalbts:
PR for #2679

### DIFF
--- a/libraries/plugins/rc/rc_plugin.cpp
+++ b/libraries/plugins/rc/rc_plugin.cpp
@@ -53,11 +53,33 @@ class rc_plugin_impl
 
       void on_first_block();
       void validate_database();
+
       bool before_first_block()
-      { return (_db.count< rc_account_object >() == 0); }
+      {
+         //
+         // This method returns _db.count< rc_account_object >() == 0.
+         // But we know that if this check ever returns false, all
+         // subsequent executions of the check will return false.
+         //
+         // So we can do an optimization which saves the per-op count()
+         // call in the common case with a simple caching algorithm:
+         //
+         // - Initialize the cached check result to true
+         // - Cache a false check result forever
+         // - Don't cache a true check result (i.e. re-run the check
+         // if the cached result is true)
+         //
+         if( _before_first_block_last_result )
+         {
+            _before_first_block_last_result = (_db.count< rc_account_object >() == 0);
+         }
+         return _before_first_block_last_result;
+      }
 
       database&                     _db;
       rc_plugin&                    _self;
+
+      bool                          _before_first_block_last_result = true;
 
       // State is shared between the pre- and post- visitor
       visitor_shared_state          _shared_state;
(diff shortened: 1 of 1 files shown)


libraries/plugins/rc/rc_plugin.cpp
```diff
@@ -53,6 +53,8 @@ class rc_plugin_impl
 
       void on_first_block();
       void validate_database();
+      bool before_first_block()
+      { return (_db.count< rc_account_object >() == 0); }
```
mvandeberg: I think this should be done without a db read. While technically O(1) there is a good amount of indirection to read a value that we can know through observation and once set does not change.

The rc plugin already listens to the post apply block notification.

I would simply track this locally as a bool. On startup perform this check and set the bool accordingly. Then, after the first block set the bool and read from it instead of the database.

Unless there is some corner case when `_db.count< rc_account_object >() == 0` could be true after the first block. In which case, `before_first_block` should be renamed to something more accurate.

Admittedly, my solution does not cover the case when the first block is forked and an alternate first block is generated. However, because that can only happen if the initminer is misconfigured, I would consider that an ops problem and not a legitimate consensus corner case that we need to cover.

theoreticalbts:
> On startup perform this check and set the bool accordingly.

I took a crack at this, the semantics are a little tricky, and I am not super confident I didn't introduce a subtle bug in this refactoring.

I default-initialize the flag to `false`, then set it to `true` in `plugin_startup()` if the head block number is greater than zero.

There are several cases it has to work correctly in:

# Replay

- If a replay is requested, the replay executes in `chain_plugin::plugin_startup()`, which happens *before* `rc_plugin::plugin_startup()`.
- So during the replay, `_before_first_block` will have its default-initialized value of `false`.
- This is always correct, as replays always start at block `1`.

# New datadir

- If the node is started with a new datadir, `rc_plugin::plugin_startup()` sees `_db.head_block_num() == 0`.
- This leads it to correctly set `_before_first_block`.

# Existing datadir

- If the node is started with an existing datadir, `rc_plugin::plugin_startup()` sees `_db.head_block_num() > 0`.
- This leads it to correctly clear `_before_first_block`.

Is there a race condition if a post-handler is applied before `plugin_startup()` is executed? 
 I.e. the p2p plugin could be started before the RC plugin, and some transaction/block incoming on the p2p network could cause a database notification (and activate `rc_plugin`'s pre/post notification hooks before its `rc_plugin::plugin_startup()` method is called)?

I *think* this code works, and I am sure that if it does work, it is an optimization that saves a database read per op.  But I have significantly less confidence in its correctness compared to the old code.

theoreticalbts:
Good idea.  I force-pushed with a different patch, which is essentially this, but with a clearer explanation (to me, anyway) in the comments.  And instead of checking `head_block_num()` and possibly having an off-by-one error, I check the existence of any `rc_account` objects, which is what this code cares about anyway.

Issue #683, opened 2016-12-08
Get schema working with shared_db


theoreticalbts:
Schemas were broken by shared_db changes.  They need to be unbroken.

theoreticalbts:
I'm still actively working on this when I have time to spare, so I'll keep my assignment for now.

Issue #2549, opened 2018-06-14
State Size RC


youkaicountry:
Implement a resource credit that represents state size.

theoreticalbts:
I think I can repurpose some of my old code to help with this.

mvandeberg:
#2693 fixed the json schema. There are tests for the schema code that should be added to our CI progress to ensure it is working properly. This is critical to ensure that this issue works.

PR #2693, opened 2018-08-01
Modernize schema

theoreticalbts:
This gets the schema code compiling with modern `steemd`, see #683.

It doesn't do any integration.

### DIFF
--- a/programs/util/CMakeLists.txt
+++ b/programs/util/CMakeLists.txt
@@ -55,17 +55,17 @@ install( TARGETS
    ARCHIVE DESTINATION lib
 )
 
-#add_executable( schema_test schema_test.cpp )
-#target_link_libraries( schema_test
-#                       PRIVATE steem_chain fc ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
-#
-#install( TARGETS
-#   schema_test
-#
-#   RUNTIME DESTINATION bin
-#   LIBRARY DESTINATION lib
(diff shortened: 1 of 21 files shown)


theoreticalbts:
Tests were failing because `js_operation_serializer` didn't compile, due to missing `fc::get_typename` overrides for `fixed_string` and `legacy_asset_symbol_type`.

I added the missing overrides, this should be ready to merge as soon as it passes CI.

programs/util/CMakeLists.txt
```diff
@@ -55,17 +55,17 @@ install( TARGETS
...
-#   RUNTIME DESTINATION bin
-#   LIBRARY DESTINATION lib
-#   ARCHIVE DESTINATION lib
-#)
+add_executable( schema_test schema_test.cpp )
+target_link_libraries( schema_test
+                       PRIVATE steem_chain fc ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
+
+install( TARGETS
+   schema_test
```
mvandeberg: We should add these tests to the Docker file to run as part of CI. Especially because the rc state size code is going to rely on it. We need to make sure this works properly and is not broken by a code change.

mvandeberg:
I will merge, but am adding passing these tests to the acceptance criteria of #2549 

Presumably these tests worked and passed at some point in the past. It seems like a no-brainer to pass them now if we are going to use this code.

PR #2702, opened 2018-08-03
Implement state size count #2549

theoreticalbts:
PR for #2549.  This compiles and should pass CI, but I haven't tested it against mainnet.

### DIFF
--- a/libraries/plugins/rc/resource_count.cpp
+++ b/libraries/plugins/rc/resource_count.cpp
@@ -3,8 +3,21 @@
 
 #include <steem/protocol/transaction.hpp>
 
-#define STATE_BYTES_SCALE              10000
-#define STATE_TRANSACTION_BYTE_SIZE      175
+#define STATE_BYTES_SCALE                       10000
+
+// These numbers are obtained from computations in state_byte_hours.py script
+//
+// $ ./state_byte_hours.py
+// {"seconds":196608}
+// 174
+// {"days":3}
+// 229
+// {"days":28}
+// 1940
+
+#define STATE_TRANSACTION_BYTE_SIZE               174
+#define STATE_TRANSFER_FROM_SAVINGS_BYTE_SIZE     229
+#define STATE_LIMIT_ORDER_BYTE_SIZE              1940
 
 namespace steem { namespace plugins { namespace rc {
 
@@ -45,7 +58,10 @@ struct state_object_size_info
    int64_t escrow_object_base_size            = 119    *STATE_BYTES_SCALE;
 
    // limit_order_object
-   int64_t limit_order_object_base_size       = 76     *STATE_BYTES_SCALE;
+   int64_t limit_order_object_base_size       = 76     *STATE_LIMIT_ORDER_BYTE_SIZE;
+
+   // savigns_withdraw_object
+   int64_t savings_withdraw_object_byte_size  = 64     *STATE_TRANSFER_FROM_SAVINGS_BYTE_SIZE;
 
    // transaction_object
    int64_t transaction_object_base_size       =  35    *STATE_TRANSACTION_BYTE_SIZE;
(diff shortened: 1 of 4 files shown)


libraries/plugins/rc/resource_count.cpp
```diff
@@ -3,35 +3,238 @@
...
 #include <steem/protocol/transaction.hpp>
 
+#define STATE_BYTES_SCALE              10000
+#define STATE_TRANSACTION_BYTE_SIZE      175
+
 namespace steem { namespace plugins { namespace rc {
 
 using namespace steem::protocol;
 
+struct state_object_size_info
```
mvandeberg: The whitespace in this struct is kind of all over the place.

I am all for adding whitespace to align constants. Not sure whats up with `*STATE_BYTES_SCALE`. I don't think we've ever had multiplication without whitespace. Looks like a pointer to me.

libraries/plugins/rc/resource_count.cpp
```diff
@@ -3,35 +3,238 @@
...
+   void operator()( const comment_operation& op )const
+   {
+      state_bytes_count +=
+           _w.comment_object_base_size
+         + _w.comment_object_permlink_char_size * op.permlink.size()
+         + _w.comment_object_parent_permlink_char_size * op.parent_permlink.size()
+         ;
+   }
+
+   void operator()( const comment_payout_beneficiaries& bens )const
```
mvandeberg: There is no refund of cost for deleting an account. Was this overlooked or a limitation of the rc plugin?

libraries/plugins/rc/resource_count.cpp
```diff
@@ -3,35 +3,238 @@
...
+      state_bytes_count += std::max(
+         _w.vesting_delegation_object_base_size,
+         _w.vesting_delegation_expiration_object_base_size
+         );
+   }
+
+   void operator()( const escrow_transfer_operation& op )const
+   { state_bytes_count += _w.escrow_object_base_size;                            }
+
+   void operator()( const limit_order_create_operation& op )const
```
mvandeberg: Same with limit order cancel

libraries/plugins/rc/resource_count.cpp
```diff
@@ -3,35 +3,238 @@
...
-   void operator()( const account_create_operation& op )const
-   { new_account_op_count++; }
-   void operator()( const account_create_with_delegation_operation& op )const
-   { new_account_op_count++; }
+   void operator()( const transfer_to_savings_operation& )const
+   {
+      FC_TODO( "Change RC state bytes computation to take SMT's into account" )
+   }
+
+   void operator()( const transfer_from_savings_operation& )const
```
mvandeberg: This creates an object

theoreticalbts:
A couple quick notes here:  What about that 1.75% constant?  It's because I did some back-of-the-envelope calculation that shows state bytes that last less than three days should cost 1.75% as much as state bytes that last forever.

Think about measuring state bytes as a time-weighted cost of the "I consume a byte at the point in time `t` hours in the future".  If these costs don't fall quickly for values far in the future, then "consume a byte of state forever" will have an infinite time-weighted cost, which we don't want.  So we exponentially discount the future costs; I picked 90 days as a reasonable half-life for the discounting.  So it works out that the cost of consuming is about 1.74% for the 54.613 hours of TaPoS, or 2.29% for three days for `transfer_from_savings`.

Here's some Python code for obtaining these values:

```
>>> import math
>>> k = math.log(2) / (90.0*24)
>>> f = lambda x : (1/k) * (1-math.exp(-k*x))
>>> tapos_hours = (65536 * 3) / (60*60)
>>> tapos_hours
54.61333333333334
>>> f(tapos_hours) * 10000 / f(365*24*1000)
173.72820747158624
>>> f(24*3) * 10000 / f(365*24*1000)
228.40031565754094
```

theoreticalbts:
Updated with missing cases and a Python script to implement the computation

PR #2711, opened 2018-08-05
Fail entire build fast when one part of the job fails

jredbeard:
This fixes two things with CI.

One: if one of the parllel jenkins jobs fails, the other will immediately halt, which is the desired effect.

Two: since jobs now run on multiple instances, docker cleanup needs to happen at the end of each build script in case instances are recycled and used again before being terminated.


Smittyvb:
Why 2 PRs? You should close one

jredbeard:
@smittyvb - these are for CI. One for master, one for develop. This is intended.

Issue #2511, opened 2018-05-29
StatsD push_block / push_transaction timing


youkaicountry:
The time `push_block` and `push_transaction` take to execute should be logged, as well as any useful information (block id, txn id, etc)

Acceptance criteria:
- [ ] Time `push_block` takes should be measured.
- [ ] Time `push_transaction` takes should be measured
- [ ] Data should be sent via StatsD plugin

mvandeberg:
Some of this work may have been done already in the `chain_plugin`

PR #2713, opened 2018-08-06
Use statsd key push_transaction instead of push_tx

sgerbino:
As per the acceptance criteria stated in #2511, we change the statsd key from `push_tx` to `push_transaction`.


theoreticalbts:
Looks reasonable.

PR #2764, opened 2018-08-07
Appbase build instruction fix: checkout master branch

Jolly-Pirate:


TimCliff:
@Jolly-Pirate it may be best to leave it as stable. AppBase is kind of a unique situation in that we are supporting two stable versions at the same time (stable and master). Under normal conditions though, there should only be one stable - which is probably the right one to have as a default.

Jolly-Pirate:
I thought it would be less confusing for those testing both versions.

PR #2765, opened 2018-08-07
Add statsd timers to jsonrpc plugin

### LINKED ISSUE #2512: StatsD API timing
The time for each API call should be measured, as well as any useful data regarding that call such as API call type.

Acceptance Criteria:
- [ ] Response time of each API call is measured
- [ ] Data sent via StatsD plugin.

mvandeberg:
Closes #2512

Names/frequencies are open to discussion if something things obviously wrong. This is a best guess at what is useful.


theoreticalbts:
Looks good.  Might be interesting to check `ddump()` -- does it still do JSON formatting when logging isn't set to debug?

Also, it might be good to have a separate measurement of how long it takes to obtain a lock.

But all this could be done in future PR's.

PR #2771, opened 2018-08-08
If ZLIB_ROOT_DIR and/or BZIP2_ROOT_DIR are not set, use environment variables if available

### LINKED ISSUE #2770: Eliminate flags required to build RocksDB with CMake
It would be ideal to not require passing flags to get RocksDB to compile. For other dependencies it seems we require setting environment variables (as we can see from building.md), we should probably do the same for RocksDB.

For example, instead of invoking CMake like this: 
```
cmake -DBUILD_STEEM_TESTNET=ON -DZLIB_ROOT_DIR="$ZLIB_ROOT_DIR" -DBZIP2_ROOT_DIR="$BZIP2_ROOT_DIR" -DCMAKE_BUILD_TYPE=Debug .. -G Xcode
```
 
We should only have to CMake as follows:
```
cmake -DBUILD_STEEM_TESTNET=ON -DCMAKE_BUILD_TYPE=Debug .. -G Xcode
```

sgerbino:
Closes #2770.

### DIFF
--- a/libraries/vendor/CMakeLists.txt
+++ b/libraries/vendor/CMakeLists.txt
@@ -9,6 +9,18 @@ SET(WITH_ZLIB ON CACHE BOOL "build with ZLIB")
 SET(WITH_BZ2 ON CACHE BOOL "build with BZ2")
 SET(WITH_BENCHMARKS OFF CACHE BOOL "build with BENCHMARKS")
 
+# If we don't have CMake variables defined, try to get them from the environment
+if (NOT DEFINED ZLIB_ROOT_DIR)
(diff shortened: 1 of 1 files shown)


sgerbino:
After exporting these variables, it is not required to pass additional flags to CMake with this change.

```
export BZIP2_ROOT_DIR=$(brew --cellar bzip2)/$(brew list --versions bzip2 | tr ' ' '\n' | tail -1)
export ZLIB_ROOT_DIR=$(brew --cellar zlib)/$(brew list --versions zlib | tr ' ' '\n' | tail -1)
```

PR #2772, opened 2018-08-08
2642 add check for max block size

rajatsteemit:
Added check to maximum_block_size for HF20


libraries/chain/steem_evaluator.cpp
```diff
@@ -95,16 +95,10 @@ void witness_update_evaluator::do_apply( const witness_update_operation& o )
       wlog( "Wrong fee symbol in block ${b}", ("b", _db.head_block_num()+1) );
    }
 
-   #pragma message( "TODO: This needs to be part of HF 20 and moved to validate if not triggered in previous blocks" )
-   if( _db.is_producing() )
+   // May be we can move this check to - validate.hpp
+   if( _db.is_producing() && _db.has_hardfork( STEEM_MAX_BLOCK_SIZE_HF20 ))
```
mvandeberg: What you're saying here is when Steem has hardfork 2097152 then enforce the rule. We want to enforce this on HF 20.

libraries/protocol/include/steem/protocol/config.hpp
```diff
@@ -234,7 +234,7 @@
 #define STEEM_MAX_TRANSACTION_SIZE            (1024*64)
 #define STEEM_MIN_BLOCK_SIZE_LIMIT            (STEEM_MAX_TRANSACTION_SIZE)
 #define STEEM_MAX_BLOCK_SIZE                  (STEEM_MAX_TRANSACTION_SIZE*STEEM_BLOCK_INTERVAL*2000)
-#define STEEM_SOFT_MAX_BLOCK_SIZE             (2*1024*1024)
+#define STEEM_MAX_BLOCK_SIZE_HF20             (2*1024*1024)
```
mvandeberg: This name doesn't need to change because we aren't changing the value during the hardfork, just adding a new enforcement case.

libraries/chain/database.cpp
```diff
@@ -4853,6 +4853,7 @@ void database::apply_hardfork( uint32_t hardfork )
                gpo.reverse_auction_seconds = STEEM_REVERSE_AUCTION_WINDOW_SECONDS_HF20;
                gpo.sbd_stop_percent = STEEM_SBD_STOP_PERCENT_HF20;
                gpo.sbd_start_percent = STEEM_SBD_START_PERCENT_HF20;
+               gpo.maximum_block_size = STEEM_MAX_BLOCK_SIZE_HF20;
```
mvandeberg: This is not good. The max block size is already voted on by the witnesses and enforced. We don't want to change that simply limit what the maximum max block size is.

libraries/chain/steem_evaluator.cpp
```diff
@@ -95,14 +95,7 @@ void witness_update_evaluator::do_apply( const witness_update_operation& o )
...
 
-   #pragma message( "TODO: This needs to be part of HF 20 and moved to validate if not triggered in previous blocks" )
-   if( _db.is_producing() )
-   {
-      FC_ASSERT( o.props.maximum_block_size <= STEEM_SOFT_MAX_BLOCK_SIZE, "Max block size cannot be more than 2MiB" );
-   }
-
-   #pragma message( "TODO: This needs to be part of HF 20 and moved to validate if not triggered in previous blocks" )
-   if( _db.is_producing() )
+   if( _db.is_producing() && _db.has_hardfork( STEEM_HARDFORK_0_20 ))
```
mvandeberg: The only thing to do is to make this a specific hardfork constant for the issue. Add `STEEM_HARDFORK_0_20__2642` to `protocol/hardfork.d/0_20.hf`

PR #2779, opened 2018-08-10
Remove duplicate p2p nodes

sgerbino:


abitmore:
Why not change `vector` to `set`?

mvandeberg:
> Why not change vector to set?

That is a perfectly reasonable solution, but this code executes once on load. There are usually not more than a couple dozen seeds specified. I am not particularly worried about this being a highly optimized piece of code. Sending the PR back to be changed uses more time than this code will ever spend executing.

PR #2780, opened 2018-08-10
Fix varint unpack

mvandeberg:


pmconrad:
I think changing unpack() can break consensus.

mvandeberg:
In this case it does not. This particular change is to sanitize inputs from a varint larger than 32 bits. Yes, that input may be a block on the p2p network, but we deserialize it first and then when we check the signature we serialize, hash, and verify against that. Our serialization code is well formed so if serialization and deserialization are not symmetric (as they are prior to this fix) we will not have a matching signature. In order for a block to be accepted by consensus now it already needs to not break this rule.

pmconrad:
Makes sense, thanks.

theoreticalbts:
Looks reasonable.

PR #2781, opened 2018-08-10
Closes #2478 - Update default shared-file-size

### LINKED ISSUE #2478: First time run of steemd starts config with 24G shared file
Based on the changes for issue https://github.com/steemit/steem/issues/2176, the first time steemd launches, it is supposed to load with reasonable defaults to load steemd without any additional changes.

The default config file that is getting created at `.steemd/config.ini` after running `steemd` for the first time has `shared-file-size` set to `24G` and the autoscaling set to off:
```
# Size of the shared memory file. Default: 24G. If running a full node, increase this value to 200G.
shared-file-size = 24G

# A 2 precision percentage (0-10000) that defines the threshold for when to autoscale the shared memory file. Setting this to 0 disables autoscaling. Recommended value for consensus node is 9500 (95%). Full node is 9900 (99%)
shared-file-full-threshold = 0

# A 2 precision percentage (0-10000) that defines how quickly to scale the shared memory file. When autoscaling occurs the file's size will be increased by this percent. Setting this to 0 disables autoscaling. Recommended value is between 1000-2000 (10-20%)
shared-file-scale-rate = 0
```

Should this be increased to 54G? Also, should autoscaling be enabled with the recommended values for consensus nodes?

mvandeberg:
Internal PR for #2778
Closes #2478

### DIFF
--- a/libraries/plugins/chain/chain_plugin.cpp
+++ b/libraries/plugins/chain/chain_plugin.cpp
@@ -296,7 +296,7 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
    cfg.add_options()
          ("shared-file-dir", bpo::value<bfs::path>()->default_value("blockchain"),
             "the location of the chain shared memory files (absolute path or relative to application data dir)")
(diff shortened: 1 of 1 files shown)


theoreticalbts:
Looks reasonable.  I'm surprised it wasn't 54G already.

PR #2784, opened 2018-08-13
2783 automate tinman start time

### LINKED ISSUE #2783: TN: Accomodate non-optimal snapshot dates when creating a testnet.
As the creator of a testnet (TN) using contributed shell scripts, I want to be able to create one that will run for a reasonable length of time before hitting the block limit because a TN that does not run long enough is a sad puppy.

**AC**
- [ ] `/contrib/testnetinit.sh` sets the start_time to something from the recent past, far enough that the TN can catch up, run and use the `--realtime` flag in [tinman](https://github.com/steemit/tinman/)

relativityboy:
Closes #2783


relativityboy:
Error adding this PR to the board. We're tracking the issue.

PR #2790, opened 2018-08-14
Remove partial subsidization

### LINKED ISSUE #2785: Remove Partial Account Subsidization
For simplicity we are removing partial subsidization for new accounts.

mvandeberg:
Closes #2785

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -2351,56 +2351,8 @@ void claim_account_evaluator::do_apply( const claim_account_operation& o )
 
    FC_ASSERT( creator.balance >= o.fee, "Insufficient balance to create account.", ( "creator.balance", creator.balance )( "required", o.fee ) );
 
-   FC_TODO( "Remove min() after HF20" );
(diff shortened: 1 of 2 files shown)


theoreticalbts:
Looks reasonable.

libraries/chain/steem_evaluator.cpp
```diff
@@ -2432,13 +2384,20 @@ void claim_account_evaluator::do_apply( const claim_account_operation& o )
...
       _db.modify( gpo, [&]( dynamic_global_property_object& gpo )
       {
-         gpo.available_account_subsidies -= current_subsidy;
+         gpo.available_account_subsidies -= STEEM_ACCOUNT_SUBSIDY_PRECISION;
       });
    }
+   else
+   {
+      FC_ASSERT( o.fee == wso.median_props.account_creation_fee,
+         "Cannot pay more than account creation fee. paid: ${p} fee: ${f}",
```
plink01001: Can this error condition also be hit if the paid fee is less than the account creation fee?

mvandeberg: Yes,

```
if( o.fee.amount == 0 )
{
 ...
}
else
{
 FC_ASSERT( ... )
}
```

mvandeberg: And there is a unit test for the case of the op fee being between 0 and the account creation fee.

PR #2803, opened 2018-08-17
2801 keep tn bootstrap node from mcfly time 2

### LINKED ISSUE #2801: TN: Keep Boostrap node from running past the present
As a developer or a shell script working with a testnet initialized with a `Bootstrap` node and [Tinman] I want the testnet to last more than an hour after all the transactions are moved over from Mainnet so that I can use the testnet for weeks instead of 30 minutes.

Relates to steemit/tinman#68

**AC**
- [x] add the `--realtime` flag to `testnetinit.sh`

relativityboy:
Closes #2801 

relates to steemit/tinman#68


relativityboy:
NM. Moving too fast. Already there.

Issue #2786, opened 2018-08-14
Implement Account Resource Dynamics in Consensus


mvandeberg:
Change how subsidized accounts are tracked by using the same resource dynamics used in the RC Plugin to increase supply stability.

PR #2808, opened 2018-08-20
Use same resource dynamics for consensus #2786

theoreticalbts:
PR for #2786


tests/tests/operation_tests.cpp
```diff
@@ -6858,6 +6858,8 @@ BOOST_AUTO_TEST_CASE( witness_set_properties_validate )
       prop_op.props[ "url" ] = fc::raw::pack_to_vector( "\xE0\x80\x80" );
       STEEM_REQUIRE_THROW( prop_op.validate(), fc::assert_exception );
 
+      FC_TODO( "Rewrite account subsidy tests" );
+/*
```
goldibex: why weren't the tests rewritten here?

theoreticalbts:
Pushed WIP

theoreticalbts:
> why acquire a new scope here?

I want to re-use the names `decay`, `budget`, `max_pool_size`, `pool` in the per-witness logic.  The code for the two blocks is similar, but it's short enough and different enough that I don't think it's worth refactoring into a single function.

An alternative solution would be to get rid of the scope and use different names for the global/per-witness variables.  If this kind of scoping's bad coding style, I don't mind changing it

theoreticalbts:
> if this log statement is useful, retain it with the correct log level. otherwise, remove the code

It's not useful until it's needed.  I don't want to rewrite it from scratch when I need it.  But I also commonly run with `info` level and I don't want the log to be super noisy with per-block messages.  I also recall being told sometime in the very distant past about negative performance impact of intensive FC logging, even if the log level is suppressed.

My own opinion is that commented-out logging statements are acceptable coding style, especially for new code.

tests/tests/operation_tests.cpp
```diff
@@ -7117,30 +7167,34 @@ BOOST_AUTO_TEST_CASE( claim_account_apply )
    try
    {
       BOOST_TEST_MESSAGE( "Testing: claim_account_apply" );
-
+      FC_TODO( "Rewrite account subsidy tests" );
```
mvandeberg: Remove erroneous TODO

libraries/chain/witness_schedule.cpp
```diff
@@ -59,36 +65,63 @@ void update_median_witness_props( database& db )
...
-           * ( uint64_t( STEEM_ACCOUNT_SUBSIDY_PRECISION )
-           *   uint64_t( STEEM_WITNESS_SUBSIDY_PERCENT   ) )
-         ) / STEEM_100_PERCENT;
+      _wso.median_props.account_subsidy_budget     = median_account_subsidy_budget;
+      _wso.median_props.account_subsidy_decay      = median_account_subsidy_decay;
+
+      rd_setup_dynamics_params( account_subsidy_user_params, account_subsidy_system_params, _wso.account_subsidy_rd );
+      rd_setup_dynamics_params( account_subsidy_per_witness_user_params, account_subsidy_system_params, _wso.account_subsidy_witness_rd );
+
+      int64_t median_decay = rd_compute_pool_decay( _wso.account_subsidy_witness_rd.decay_params, median_available_witness_account_subsidies, 1 );
```
mvandeberg: This is not what I had envisioned at all for witness subsidy restrictions, but if I am reading this correctly, then this will attempt to distribute account subsidy claims evenly across all witnesses.

libraries/chain/util/rd.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+int64_t rd_compute_pool_decay(
+   const rd_decay_params& decay_params,
+   int64_t current_pool,
+   uint32_t dt
+   )
+{
+   /*
+   ilog( "compute_pool_decay( ${params}, ${pool}, ${dt} )",
```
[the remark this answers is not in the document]
mvandeberg: +1

libraries/protocol/include/steem/protocol/config.hpp
```diff
@@ -272,9 +272,34 @@
...
+#define STEEM_RD_MAX_DECAY_BITS              32
+#define STEEM_RD_DECAY_DENOM_SHIFT           36
+#define STEEM_RD_MAX_POOL_BITS               64
+#define STEEM_RD_MAX_BUDGET_1                ((uint64_t(1) << (STEEM_RD_MAX_POOL_BITS + STEEM_RD_MIN_DECAY_BITS - STEEM_RD_DECAY_DENOM_SHIFT))-1)
+#define STEEM_RD_MAX_BUDGET_2                ((uint64_t(1) << (64-STEEM_RD_DECAY_DENOM_SHIFT))-1)
+#define STEEM_RD_MAX_BUDGET_3                (uint64_t( std::numeric_limits<int32_t>::max() ))
+#define STEEM_RD_MAX_BUDGET                  (int32_t( std::min( { STEEM_RD_MAX_BUDGET_1, STEEM_RD_MAX_BUDGET_2, STEEM_RD_MAX_BUDGET_3 } )) )
+#define STEEM_RD_MIN_DECAY                   (uint32_t(1) << STEEM_RD_MIN_DECAY_BITS)
+#define STEEM_RD_MIN_BUDGET                  1
+#define STEEM_RD_MAX_DECAY                   (uint32_t(0xFFFFFFFF))
```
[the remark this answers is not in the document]
mvandeberg: +1

goldibex:
@mvandeberg lgty?

PR #2823, opened 2018-08-22
Document Witness Parameters

### LINKED ISSUE #2810: Create Witness Parameter Documentation
Create documentation to help witnesses understand and set the new parameters.

mvandeberg:
Closes #2810


theoreticalbts:
- `account_subsidy_budget` units are `STEEM_PRECISION`
- `account_subsidy_decay` has 100% at `2^36`, not `2^34` (this means decays of faster than 1/16 per block cannot be represented, this is on purpose, if you need to restrict usage on very short timescales, `rc_plugin` is the wrong tool for the job -- you should be enforcing a per-block limit instead)

mvandeberg:
https://github.com/steemit/steem/blob/77e05e8f1a0de4d2e6743eafbd92cf5a6a941e4c/libraries/protocol/include/steem/protocol/config.hpp#L282

`account_subsidy_decay` has 100% decay at 2^32 according to the config.

Feel free to make your own modifications to the documentation.

theoreticalbts:
I got a start on this the other day, but it's not anywhere near finished.  I'm not actively working on these notes now, so feel free to edit them or combine them.

mvandeberg:
> At the end of this document there are sections on setting parameter values. This could potentially be audited/removed in favor of the information in witness_parameters.md.

This section literally links directly to `witness_parameters.md`. I think the mention of more documentation along with the link is better than not saying anything. I would like to see more links in our documentation so that users do not need to look through every file to find what they are looking for.

> It may be a good idea to include a reference to the resource-notes document, to help the reader get more information.

Where do you want this link?

youkaicountry:
Okay, I'm comfortable with leaving the parameter values section in `2018-08-20-resource-notes.md`.
As for linking to `2018-08-20-resource-notes.md` in `witness-parameters.md`, there should be a note following `account_subsidy_budget` and `account_subsidy_decay` directing the reader to find more details regarding resources in `2018-08-20-resource-notes.md`. It could be added in a footnote / reference fashion.

PR #2828, opened 2018-08-23
Change --chain-id option to accept hex string instead of sha-seed

### LINKED ISSUE #2827: Chain plugin chain ID command line parameter isn't specified in hex
`--chain-id` option to `steemd` [here](https://github.com/steemit/steem/blob/develop/libraries/plugins/chain/chain_plugin.cpp#L318) is specified as a string, not a hex value.  It should be specified as a hex value for consistency with the rest of the tooling (`get_config` API call, `sign_transaction` CLI).

mvandeberg:
Closes #2827


theoreticalbts:
Approved.  A couple notes:

- I'm not sure if the better way to treat the `fc::sha256::hash` constructor is to zero-pad, or to error out.

mvandeberg:
I want to zero out so that we can use a git commit hash as a chain id.

Issue #2807, opened 2018-08-20
percent_steem_dollars in comment_options shows random results on AppBase


Jolly-Pirate:
When querying block # 25220874 for example and checking transaction_num 15, the `percent_steem_dollars` in `comment_options` shows 10000. Working properly with non-appbase:
```
curl -s -d '{"id":0,"jsonrpc":"2.0","method":"call","params":["condenser_api","get_block",[25220874]]}' https://rpc.steemviz.com | jq -r ".result.transactions[15]"
curl -s -d '{"id":0,"jsonrpc":"2.0","method":"call","params":["condenser_api","get_block",[25220874]]}' https://rpc.buildteam.io | jq -r ".result.transactions[15]"
curl -s -d '{"id":0,"jsonrpc":"2.0","method":"call","params":["condenser_api","get_block",[25220874]]}' https://gtg.steem.house:8090/ | jq -r ".result.transactions[15]"
```
(requires jq on linux, `apt install jq`)

![percent_steem_dollars bug nonappbase](https://user-images.githubusercontent.com/4613678/44346185-5d5e9f80-a463-11e8-8319-c1d7ded5c28b.jpg)

But on appbase nodes, it's giving random results every time it's queried! Interestingly on api.steemit.com it gives one value only: 55832 (which is still wrong).
```
curl -s -d '{"id":0,"jsonrpc":"2.0","method":"call","params":["condenser_api","get_block",[25220874]]}' https://api.steemit.com | jq -r ".result.transactions[15]"
curl -s -d '{"id":0,"jsonrpc":"2.0","method":"call","params":["condenser_api","get_block",[25220874]]}' https://appbasetest.timcliff.com | jq -r ".result.transactions[15]"
curl -s -d '{"id":0,"jsonrpc":"2.0","method":"call","params":["condenser_api","get_block",[25220874]]}' https://appbase.buildteam.io | jq -r ".result.transactions[15]"
```
![percent_steem_dollars bug appbase](https://user-images.githubusercontent.com/4613678/44346200-63548080-a463-11e8-8af5-b4e38e6f250c.jpg)

mvandeberg:
My gut says to double check that the condenser API version of the comment options operation is copying `percent_steem_dollars` properly.

PR #2833, opened 2018-08-23
	modified:   libraries/plugins/apis/condenser_api/include/steem/plugi…

rajatsteemit:
…ns/condenser_api/condenser_api_legacy_operations.hpp


theoreticalbts:
The code itself is good.  But the commit message needs improvement.

- 2-10 words to explain what problem was fixed
- Add the issue number #2807
- Putting the filename in the commit message is optional.
- If you do put the filename in the commit message, only the filename exists.

Look at some recent commits for examples of what kinds of commit messages we use.

In order to change the message of a pushed commit, you can do `git commit --amend` which changes your local branch.  Then use `git push -f` to force-push and overwrite the branch.  (Please don't force-push other people's branches without asking them first.)

PR #2795, opened 2018-08-15
Implement rc curve generation in c++

mvandeberg:
Closes #2787

Due to my decision to keep this as a local header (to prevent accidental inclusion in consensus code) testing is difficult. Below are the results of the existing curves from the python and c++ implementations.

Python Reference Implementation:
```
{
 "resource_parameters":[
  [
   "resource_history_bytes",
   [
    {
     "time_unit":"rc_time_unit_blocks",
     "resource_unit":1,
     "budget_per_time_unit":34722,
     "pool_eq":21640431401,
     "max_pool_size":"43280862801",
     "curve_params":{
      "coeff_a":"9736235291562360832",
      "coeff_b":"169065870",
      "coeff_d":"0",
      "shift":47
     },
     "decay_params":{
      "decay_per_time_unit":"3613026481",
      "decay_per_time_unit_denom_shift":51
     }
    },
    {}
   ]
  ],
[
   "resource_new_accounts",
   [
    {
     "time_unit":"rc_time_unit_blocks",
     "resource_unit":10000,
     "budget_per_time_unit":0,
     "pool_eq":14426950409,
     "max_pool_size":"28853900818",
     "curve_params":{
      "coeff_a":"9904210705121218560",
      "coeff_b":"112710550",
      "coeff_d":"0",
      "shift":31
     },
     "decay_params":{
      "decay_per_time_unit":"3551752401",
      "decay_per_time_unit_denom_shift":67
     }
    },
    {}
   ]
  ],
  [
   "resource_market_bytes",
   [
    {
     "time_unit":"rc_time_unit_blocks",
     "resource_unit":100,
     "budget_per_time_unit":578704,
     "pool_eq":16030041350,
     "max_pool_size":"32060082699",
     "curve_params":{
      "coeff_a":"13847090192444250112",
      "coeff_b":"125234698",
      "coeff_d":"0",
      "shift":52
     },
     "decay_params":{
      "decay_per_time_unit":"2540365427",
      "decay_per_time_unit_denom_shift":46
     }
    },
    {}
   ]
  ],
  [
   "resource_state_bytes",
   [
    {
     "time_unit":"rc_time_unit_blocks",
     "resource_unit":1,
     "budget_per_time_unit":23148148,
     "pool_eq":14426954266915,
     "max_pool_size":"28853908533829",
     "curve_params":{
      "coeff_a":"9736235291562360832",
      "coeff_b":"112710580210",
      "coeff_d":"0",
      "shift":47
     },
     "decay_params":{
[...]

Issue #2788, opened 2018-08-14
RC Plugin Account Subsidization Integration


mvandeberg:
Blocked on #2786 #2787

Finish integration of consensus and non-consensus portions of subsidized account creation.

Issue #2787, opened 2018-08-14
Port RC Curve Generation to C++


mvandeberg:
Port the Python script that generates RC curves to C++.

This will be intended for non-consensus use only and will contain some floating point calculations to make porting easier.

PR #2839, opened 2018-08-23
RC plugin subsidy integration #2788

theoreticalbts:
Combined PR for #2787 #2788.  Integrates and obsoletes PR #2795.

### DIFF
--- a/libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
+++ b/libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
@@ -10,12 +10,12 @@ namespace steem { namespace plugins { namespace rc {
 
 struct rc_curve_gen_params
 {
-   uint16_t                inelasticity_threshold_num    = 1;
-   uint16_t                inelasticity_threshold_denom  = 128;
-   uint16_t                a_point_num                   = 37;
-   uint16_t                a_point_denom                 = 1000;
-   uint16_t                u_point_num                   = 1;
-   uint16_t                u_point_denom                 = 2;
+   uint64_t                inelasticity_threshold_num    = 1;
+   uint64_t                inelasticity_threshold_denom  = 128;
+   uint64_t                a_point_num                   = 37;
+   uint64_t                a_point_denom                 = 1000;
+   uint64_t                u_point_num                   = 1;
+   uint64_t                u_point_denom                 = 2;
 
    void validate()const;
 };
--- a/libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
+++ b/libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
@@ -12,8 +12,8 @@ struct rc_curve_gen_params
 {
    uint16_t                inelasticity_threshold_num    = 1;
    uint16_t                inelasticity_threshold_denom  = 128;
-   uint16_t                a_point_num                   = 1;
-   uint16_t                a_point_denom                 = 32;
+   uint16_t                a_point_num                   = 37;
+   uint16_t                a_point_denom                 = 1000;
    uint16_t                u_point_num                   = 1;
    uint16_t                u_point_denom                 = 2;
 
--- a/libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
+++ b/libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
@@ -1,10 +1,3 @@
-/**
- * Note:
- *
- * This header is contained within a source directory because it is not meant to be used anywhere
- * EXCEPT the rc plugin. The implementation uses floating point arithmetic which can be
- * non-deterministic. This implementation is NOT suitable for a consensus level protocol.
- */
 
 #pragma once
 #include <steem/plugins/rc/rc_utility.hpp>
--- a/libraries/plugins/rc/rc_curve.cpp
+++ b/libraries/plugins/rc/rc_curve.cpp
@@ -23,7 +23,8 @@ void rc_curve_gen_params::validate()const
 }
 
 void generate_rc_curve_params(
-   rc_resource_params& resource_params,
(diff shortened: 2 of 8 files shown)


theoreticalbts:
I rebased #2787 on `develop`, then rewrote the logic to not use `double`, then integrated it with the new RC / RD split.  This should be everything, but there's still some numerical testing to be done.

libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <steem/protocol/types.hpp>
+
+#include <fc/time.hpp>
+
+namespace steem { namespace plugins { namespace rc {
+
+struct rc_curve_gen_params
+{
+   uint16_t                inelasticity_threshold_num    = 1;
```
goldibex: why not one of `size_t`, `uint64_t`, or `unsigned int`?

libraries/plugins/rc/rc_curve.cpp
```diff
@@ -0,0 +1,90 @@
...
+   //      = (a - a*u) / (r*u)
+   //      = a / (r*u) - a/r
+   //
+   //      = (a_point_num / a_point_denom) / ((N / decay_per_time_unit) * (u_point_num / u_point_denom)) - (a_point_num / a_point_denom) / (N / decay_per_time_unit)
+   //      = (a_point_num * decay_per_time_unit * u_point_denom) / (a_point_denom * N * u_point_num ) - (a_point_num * decay_per_time_unit) / (a_point_denom * N)
+   //      = (a_point_num * decay_per_time_unit * u_point_denom - a_point_num * decay_per_time_unit * u_point_num) / (a_point_denom * N * u_point_num)
+
+   FC_ASSERT( resource_dynamics_params.decay_params.decay_per_time_unit > 0 );
+
+   uint64_t A_num_cf = uint64_t( curve_gen_params.a_point_num ) * uint64_t( resource_dynamics_params.decay_params.decay_per_time_unit );
```
goldibex: can we find a way to clean all these casts out of here?

libraries/plugins/rc/rc_curve.cpp
```diff
@@ -0,0 +1,90 @@
...
+   uint64_t A_num = A_num_cf * (curve_gen_params.u_point_denom - curve_gen_params.u_point_num);
+   uint32_t A_denom = uint32_t( curve_gen_params.a_point_denom ) * uint32_t( curve_gen_params.u_point_num );
```
goldibex: any risk of an overflow in any of these multiplications? you're assigning products of 32-bit multiplications to 32-bit values

libraries/plugins/rc/rc_curve.cpp
```diff
@@ -0,0 +1,90 @@
...
+   A_denom = std::max( A_denom, uint32_t(1) );
+
+   // How many places are needed to shift A_num left until leading 1 bit is in the top bit position of uint64_t?
+   int8_t shift1 = 63 - boost::multiprecision::detail::find_msb( A_num );
+   uint64_t A_1 = (A_num << shift1) / A_denom;
+   // n.b. A_1 cannot be smaller than 2^31 because we are dividing a number at least 2^63 by uint32_t
+   // How many places are needed to shift A_1 left until leading 1 bit is in the top bit position of uint64_t?
+   int8_t shift2 = 63 - boost::multiprecision::detail::find_msb( A_1 );
+   fc::uint128_t u128_A = A_num;
+   price_curve_params.shift = uint8_t( shift1+shift2 );
```
goldibex: i'm a little surprised we're being so granular about bit width for some of these parameters. Is there a reason for not just making them standard machine-word types?

libraries/plugins/rc/rc_curve.cpp
mvandeberg: `a_point_denom` and `u_point_num` are `uint16_t`.

libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
mvandeberg: Let's make these uint64_t and then bound them to uint16_t in `validate`. That way we can remove casting in `generate_rc_curve_params`.

libraries/plugins/rc/rc_curve.cpp
mvandeberg: Let's set the field types to `uint64_t` and bounds check them in `validate`. Removes casting. `rc_curve_gen_params` is not stored in db state. We aren't saving that much, if any, by making them 16 bit fields. Just use word length and call it good enough.

libraries/plugins/rc/rc_curve.cpp
mvandeberg: +1

libraries/plugins/rc/include/steem/plugins/rc/rc_curve.hpp
```diff
@@ -0,0 +1,38 @@
...
+   uint16_t                inelasticity_threshold_denom  = 128;
+   uint16_t                a_point_num                   = 1;
```
mvandeberg: These are not the values that we currently use but are the default script values.

```
"a_point_num" : 37,
"a_point_denom" : 1000,
"u_point_num" : 1,
"u_point_denom" : 2
```

theoreticalbts:
> Let's make these uint64_t and then bound them to uint16_t in validate. That way we can remove casting in generate_rc_curve_params.

Then there would have to be a bunch of `FC_ASSERT()` that protect against additional possible overflows.

mvandeberg:
If I assert that `uint64_t a` and `uint64_t b` are both < 2^16 and multiply them together, I can still cast `a * b` back down to a `uint32_t` without needing the check the value against overflow.

theoreticalbts:
> 23c4a42

Looks fine

Issue #2658, opened 2018-07-18
Not producing block because node didn't wake up within 500ms of the slot time.


abitmore:
Log:
>2018-07-18T18:58:27 th_a:invoke handle_block         handle_block ] Got 34 trx on block 24291322 by someguy123 -- 400 ms                        application.cpp:514
2018-07-18T18:58:34 th_a:invoke handle_block         handle_block ] Got 29 trx on block 24291323 by aggroed -- 4468 ms                  application.cpp:514
2018-07-18T18:58:34 th_a:Witness Block Production block_production_loo ] Not producing block because node didn't wake up within 500ms of the slot time.                 witness_plugin.cpp:624
2018-07-18T18:58:35 th_a:Witness Block Production block_production_loo ] Not producing block because node didn't wake up within 500ms of the slot time.                 witness_plugin.cpp:624
2018-07-18T18:58:36 th_a:invoke handle_block         handle_block ] Got 56 trx on block 24291324 by ausbitbank -- 855 ms                        application.cpp:514

Possible scenario:
* a block arrived with high latency
* before it arrived, my node didn't try to produce a block (perhaps because it's not its turn?)
* after it arrived (witness list reshuffled or somehow?), my node suddenly found it's its turn to produce next block, however, it's already in the 2nd second, so it tried twice to produce a block
* the result (fail) was guaranteed by this code (time diff was always more than 500ms on the 2nd and 3rd try):

https://github.com/steemit/steem/blob/1b53cf6258890461b322cc566e2e03ad4b1cbfb5/libraries/plugins/witness/witness_plugin.cpp#L702-L706

mvandeberg:
What version of steemd did this happen on?

abitmore:
Was on 0.19.5. Although in OP the quoted code is master branch.

PR #2855, opened 2018-08-29
Not producing block because node didn't wake up within 500ms of the slot time

rajatsteemit:
Increase the block producing LAG time to 'close to block interval' (2500ms), to avoid witness not producing the block during his slot time. This problem is due to arrival of high latency block just before the witness schedule slot time.

Fix for Issue #2658


mvandeberg:
I think the larger issue is that the loop frequency is currently equal to the lag delta (lag delta is abs(delta) < 500ms, so a total range of 1s). If there are delays at all, then the node could not produce.

Instead of increasing the lag delta, let's increase the loop frequency from 1 Hz to 5 Hz.

libraries/plugins/witness/witness_plugin.cpp
```diff
@@ -466,11 +466,13 @@ namespace detail {
 
    void witness_plugin_impl::schedule_production_loop() {
       //Schedule for the next second's tick regardless of chain state
-      // If we would wait less than 50ms, wait for the whole second.
+      // If we would wait less than 50ms, wait for at least another 200ms.
       fc::time_point now = fc::time_point::now();
       int64_t time_to_next_second = 1000000 - (now.time_since_epoch().count() % 1000000);
```
mvandeberg: Simply change this to `time_to_next_wakeup` and replace `1000000` with `200000`. The values here does be replaced with defined constants at the top of the source file instead of "magic" numbers.

libraries/plugins/witness/witness_plugin.cpp
```diff
@@ -603,7 +605,7 @@ namespace detail {
          return block_production_condition::low_participation;
       }
 
-      if( llabs((scheduled_time - now).count()) > fc::milliseconds( 500 ).count() )
+      if( llabs((scheduled_time - now).count()) > fc::milliseconds( 750 ).count() )
```
mvandeberg: Change 750 to a defined constant at the top of the source file. Yes, this isn't what was written when you started your change, but it was done wrong previously.

libraries/plugins/witness/witness_plugin.cpp
```diff
@@ -465,14 +467,8 @@ namespace detail {
...
-      //Schedule for the next second's tick regardless of chain state
-      // If we would wait less than 50ms, wait for the whole second.
-      fc::time_point now = fc::time_point::now();
-      int64_t time_to_next_second = 1000000 - (now.time_since_epoch().count() % 1000000);
-      if(time_to_next_second < 50000)      // we must sleep for at least 50ms
-         time_to_next_second += 1000000;
-
-      _timer.expires_from_now( boost::posix_time::microseconds( time_to_next_second ) );
+      // Sleep for 200ms, before checking the block production
+      _timer.expires_from_now( boost::posix_time::microseconds( BLOCK_PRODUCTION_LOOP_SLEEP_TIME ) );
```
mvandeberg: Let's still keep the offset logic to of `now % 200000`. That way the thread doesn't drift over time. Because our blocks are sent to other nodes, it is important to try and wake on the second as often as possible to reduce block propagation latency.

PR #2861, opened 2018-08-30
Move block generation to the witness plugin

### LINKED ISSUE #2717: Move Block Generation to Witness Plugin
Block generation code should be moved to witness plugin.

sgerbino:
Closes #2717.


libraries/plugins/witness/witness_plugin.cpp
```diff
@@ -621,6 +623,55 @@ namespace detail {
...
+      if( !(skip & chain::database::skip_witness_signature) )
+         FC_ASSERT( witness_obj.signing_key == block_signing_private_key.get_public_key() );
+
+      signed_block pending_block;
+
+      pending_block.previous = _db.head_block_id();
+      pending_block.timestamp = when;
+      pending_block.witness = witness_owner;
+
+      if( _db.has_hardfork( STEEM_HARDFORK_0_5__54 ) )
```
sgerbino: Is this conditional still necessary?

libraries/chain/include/steem/chain/database.hpp
```diff
@@ -364,6 +364,12 @@ namespace steem { namespace chain {
          /** this updates the vote of a single witness as a result of a vote being added or removed*/
          void adjust_witness_vote( const witness_object& obj, share_type delta );
 
+         /** this is called during block generation to ensure the witness votes for a hardfork that matches the binary configuration */
+         void adjust_witness_hardfork_version_vote( const witness_object& witness, signed_block& pending_block );
```
mvandeberg: Do these need to be done in database?

mvandeberg:
You re-created `generate_block` in the witness plugin, but did not remove it from `database`

libraries/chain/include/steem/chain/database.hpp
sgerbino: I think `apply_pending_transactions()` should be in database because it manipulates `_pending_tx_session`.
On the other hand, `adjust_witness_hardfork_version_vote()` could probably be extracted out with some interfaces to access `_hardfork_versions` and `_hardfork_times`.

libraries/plugins/witness/witness_plugin.cpp
sgerbino: Thanks, this conditional is now removed.

libraries/chain/include/steem/chain/database.hpp
sgerbino: These functions are now integrated in to the block_producer class.

mvandeberg:
Let's add a method in `chain_plugin` called, `register_block_generator` that accepts a function pointer/lambda and we'll call that on a block generate write request. This does not remove the ping pong between the chain and witness plugin on the call stack, but it will remove the source dependency between them. This is the pattern APIs and the jsonrpc plugin use so that the jsonrpc plugin can call APIs without needing to include any information about them during compilation.

libraries/plugins/chain/include/steem/plugins/chain/chain_plugin.hpp
```diff
@@ -42,6 +42,16 @@ class chain_plugin : public plugin< chain_plugin >
...
 
+   typedef std::function
+      < signed_block (
+         fc::time_point_sec,
+         const account_name_type&,
+         const fc::ecc::private_key&,
+         uint32_t ) >
+            block_generator_func;
+
+   void register_block_generator( block_generator_func func );
```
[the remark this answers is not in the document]
mvandeberg: We need this exposed for debug_plugin and unit tests. A comment should be added documenting this is a lockless write and it is a bad idea and `chain_plugin::generate_block` is called instead.

libraries/plugins/witness/include/steem/plugins/witness/witness_plugin.hpp
```diff
@@ -55,6 +56,7 @@ class witness_plugin : public appbase::plugin< witness_plugin >
    virtual void plugin_startup() override;
    virtual void plugin_shutdown() override;
 
+   witness::block_producer& block_producer();
```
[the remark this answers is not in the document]
mvandeberg: We need this exposed for debug_plugin and unit tests. A comment should be added documenting this is a lockless write and it is a bad idea and `chain_plugin::generate_block` is called instead.

libraries/plugins/chain/chain_plugin.cpp
```diff
@@ -630,4 +636,9 @@ void chain_plugin::check_time_in_block( const steem::chain::signed_block& block
    FC_ASSERT( block.timestamp.sec_since_epoch() <= max_accept_time );
 }
 
+void chain_plugin::register_block_generator( block_generator_func func )
+{
+   my->block_generator = func;
```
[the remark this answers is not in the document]
sgerbino: A warning will now be emitted when overriding the block generator.

libraries/plugins/witness/include/steem/plugins/witness/witness_plugin.hpp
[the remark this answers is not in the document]
sgerbino: We can still hide away the block producer within the witness, this accessor has now been removed.

theoreticalbts:
There's a race condition here.  The write processing thread is started by `chain_plugin::plugin_startup()`, which AFAIK because plugins execute in dependency order, occurs before `witness_plugin::plugin_startup()`.  The write processing thread reads the `block_generator` field once when it starts, it never reads the block generator field again.

I suggest handling this by the following strategy:

- Get rid of the block generator fields in `chain_plugin`, have it only exist on the `write_request_visitor` object which is local to the write processing thread
- To get the block generator into the `write_request_visitor` field, send it as a queue item (i.e., `write_request_ptr` member type)

It is also tricky to keep races from occurring during shutdown.  Can a block production request be in the queue when the witness plugin is destroyed?  If it could, then when executing the block production request, it would try to reference a non-existing `block_producer`, since the `block_producer` is "owned" by the witness plugin.

I don't think this shutdown race can occur if we only call `chain_plugin::generate_block()` from `witness_plugin::generate_block()` .  However, I don't think it's good to rely on that.

Fortunately there's a straightforward refactor which will prevent this shutdown race:

- Replace the `block_producer_func` typedef with a class which has a `generate_block()` abstract virtual method with the same signature as `block_generator_func`.  Let's call this class `abstract_block_producer`.
- Instead of passing around `block_generator_func*`, instead use `shared_ptr< abstract_block_producer >`.
- Since `abstract_block_producer` subclasses can be destroyed through a pointer to the base class, `abstract_block_producer` must define a virtual destructor.

theoreticalbts:
Turns out the startup race can't occur because it's registered in `plugin_initialize`, and the thread is started in `plugin_startup`.

PR #2863, opened 2018-08-31
Stub automated actions

### LINKED ISSUE #2681: Automated Action Framework stubs
Create stubs for automated action framework.

Acceptance:
- [x] static_variant
- [x] Block extension

mvandeberg:
Closes #2681

### DIFF
--- a/libraries/protocol/include/steem/protocol/automated_actions.hpp
+++ b/libraries/protocol/include/steem/protocol/automated_actions.hpp
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <steem/protocol/steem_actions.hpp>
+
+#include <steem/protocol/operation_util.hpp>
+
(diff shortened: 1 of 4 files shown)


theoreticalbts:
We need to add code to reject automated actions on mainnet.  Without that check, this could break consensus.

mvandeberg:
https://github.com/steemit/steem/blob/27db90b11ee031d3157f5e3ec9835a5938088daf/libraries/chain/database.cpp#L3262-L3267

All extensions that are not explicitly defined in this visitor are rejected. Broadcasting a block with the added extension to nodes that do not know of the extension will be read in, treated like garbage data, and rejected.

theoreticalbts:
I'd be in favor of removing that default, and requiring every extension type to have an explicit visitor method.  But since this is functionally correct, I'll approve it.

Issue #2458, opened 2018-05-16
Transaction polling API


youkaicountry:
Create an api call which takes: transaction id, expiration time.

It returns one of the following:
- Expiration time in future, transaction not included in block or mempool
- Transaction in mempool
- Transaction has been included in block, block not irreversible
- Transaction has been included in block, block is irreversible
- Transaction has expired, transaction is not irreversible (transaction could be in a fork)
- Transaction has expired, transaction is irreversible (transaction cannot be in a fork)
- Transaction is too old, I don't know about it

If you give it a transaction id but no expiration time, it can only return results indicating whether or not it is aware of the transaction.

theoreticalbts:
This needs to track the time of LIB and the first block in the TaPoS window.

The choices are:

- Add ~8 bytes of non-consensus fields and one `get_block()` call per block to consensus objects to track these times
- Add ~256k of consensus fields to blocks in the TaPoS window to track these times and avoid `get_block()` call
- Implement the API call and extra objects in a plugin (it is then unavailable unless you run the plugin)
- Change the spec of the API call so that it doesn't need these fields

PR #2868, opened 2018-09-06
Transaction polling api implementation

rajatsteemit:
Issue #2458 

This is a first cut to implement the transaction polling api.
Added to database_api, not sure whether this belong to condenser_api or wallet_api.
Just added the frame work. Please guide me, whether the approach is correct or not.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -296,6 +296,36 @@ bool database::is_known_transaction( const transaction_id_type& id )const
    return trx_idx.find( id ) != trx_idx.end();
 } FC_CAPTURE_AND_RETHROW() }
 
+//uint32_t database::find_transaction( const transaction_id_type& id, const time_point_sec& expiration_time ) const
+uint32_t database::find_transaction( const transaction_id_type& id, const time_point_sec& expiration_time )const
+{
+    /*
+     * Return values
+     * 0 - Not aware of the transaction (if transaction not found)
+     * 1 - Aware of the transaction (if transaction is found)
+     * 2 - Expiration time in future, transaction not included in block or mempool
+     * 3 - Transaction in mempool
+     * 4 - Transaction has been included in block, block not irreversible
+     * 5 - Transaction has been included in block, block is irreversible
+     * 6 - Transaction has expired, transaction is not irreversible (transaction could be in a fork)
+     * 7 - Transaction has expired, transaction is irreversible (transaction cannot be in a fork)
+     * 8 - Transaction is too old, I don't know about it
+     */
+
+    
+    uint32_t result = 0;
+
+    // Transaction in mempool
+    const auto& trx_idx = database::get_index<transaction_index>().indices().get<by_trx_id>();
+    if (trx_idx.find( id ) != trx_idx.end())
+    {
+        result = 3;
+        return result; // Transaction in mempool
+    }
+    
+    return result;
+}
+
 block_id_type database::find_block_id_for_num( uint32_t block_num )const
 {
    try
--- a/libraries/plugins/apis/database_api/database_api.cpp
+++ b/libraries/plugins/apis/database_api/database_api.cpp
@@ -63,6 +63,7 @@ class database_api_impl
          (verify_authority)
          (verify_account_authority)
          (verify_signatures)
+         (find_transaction)
 #ifdef STEEM_ENABLE_SMT
          (get_smt_next_identifier)
 #endif
@@ -1422,6 +1423,47 @@ DEFINE_API_IMPL( database_api_impl, verify_signatures )
    return result;
 }
 
+DEFINE_API_IMPL( database_api_impl, find_transaction )
+{
+    
+    // Yet to check the argument list
+    // FC_ASSERT ( args.size() == 1 || args.size() == 2, "Expected 1-2 arguments, was ${n}", ("n", args.size()) );
+    // auto trx_id = { args[0].as< transaction_id_type >() };
+    // auto expiration_time = { args[1].as< time_point_sec >() };
+    
+    find_transaction_return result;
+    result.trx_status_code = 0; // 0 - Not aware of the transaction
+    
+    auto trx_id = args.trx_id.as<transaction_id_type>();
+    
+    // Only transaction id is provided
+    //if (args.size() == 1)
+    {
+        try {
+            if (_db.is_known_transaction(trx_id))
+            {
+                result.trx_status_code = 1; // 1 - Aware of the transaction
+            }
+            else
+            {
+                result.trx_status_code = 0; // 0 - Not aware of the transaction
+            }
+        
+        } catch ( const fc::exception&  ) {
+            result.trx_status_code = 0; // 0 - Not aware of the transaction
+        }
+        return result;
+    }
+    //else
+    {
+        auto expiration_time = args.expiration.as<time_point_sec>();
+        
+        result.trx_status_code = _db.find_transaction(trx_id, expiration_time);
+    }
+    
+    return result;
+}
+
 #ifdef STEEM_ENABLE_SMT
 //////////////////////////////////////////////////////////////////////
 //                                                                  //
@@ -1485,6 +1527,7 @@ DEFINE_READ_APIS( database_api,
    (verify_authority)
    (verify_account_authority)
    (verify_signatures)
+   (find_transaction)
 #ifdef STEEM_ENABLE_SMT
    (get_smt_next_identifier)
 #endif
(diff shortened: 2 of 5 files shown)


libraries/plugins/apis/database_api/include/steem/plugins/database_api/database_api.hpp
```diff
@@ -132,6 +132,25 @@ class database_api
           */
          (verify_signatures)
 
+         /**
+         * This is a transaction ipolling api. This api takes transaction id and expiration (optional) as arguments
```
mvandeberg: Type: ipollling -> polling

libraries/plugins/apis/database_api/include/steem/plugins/database_api/database_api_args.hpp
```diff
@@ -556,6 +556,23 @@ struct verify_signatures_return
...
+};*/
+struct find_transaction_args
+{
+    fc::variant  trx_id;
+    fc::variant  expiration;
+};
+    
+struct find_transaction_return
+{
+   uint32_t trx_status_code;
```
mvandeberg: Return the enum and reflect the enum. In binary, it will be the int status, but will serialize as the enum types in JSON.

libraries/chain/database.cpp
```diff
@@ -296,6 +296,36 @@ bool database::is_known_transaction( const transaction_id_type& id )const
    return trx_idx.find( id ) != trx_idx.end();
 } FC_CAPTURE_AND_RETHROW() }
 
+//uint32_t database::find_transaction( const transaction_id_type& id, const time_point_sec& expiration_time ) const
+uint32_t database::find_transaction( const transaction_id_type& id, const time_point_sec& expiration_time )const
```
mvandeberg: There is nothing here that requires this be done in database. It can just as easily be done in `database_api`

libraries/plugins/apis/database_api/database_api.cpp
```diff
@@ -1422,6 +1423,47 @@ DEFINE_API_IMPL( database_api_impl, verify_signatures )
    return result;
 }
 
+DEFINE_API_IMPL( database_api_impl, find_transaction )
```
mvandeberg: I am assuming this implementation is WIP because it does not conform to the documented behavior at all.

rajatsteemit: What is WIP?

libraries/chain/database.cpp
rajatsteemit: Moved to database_api

libraries/plugins/apis/database_api/include/steem/plugins/database_api/database_api_args.hpp
rajatsteemit: done, returning enum.

libraries/chain/database.cpp
```diff
@@ -296,6 +296,36 @@ bool database::is_known_transaction( const transaction_id_type& id )const
...
+     * 6 - Transaction has expired, transaction is not irreversible (transaction could be in a fork)
+     * 7 - Transaction has expired, transaction is irreversible (transaction cannot be in a fork)
+     * 8 - Transaction is too old, I don't know about it
+     */
+
+    
+    uint32_t result = 0;
+
+    // Transaction in mempool
+    const auto& trx_idx = database::get_index<transaction_index>().indices().get<by_trx_id>();
```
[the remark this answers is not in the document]
rajatsteemit: Not sure, how to look through transactions to get the transaction state. Shall we talk when you get a chance. Thanks

libraries/plugins/apis/database_api/database_api.cpp
mvandeberg: WIP = Work In Progress

libraries/plugins/apis/database_api/database_api.cpp
```diff
@@ -1425,40 +1425,94 @@ DEFINE_API_IMPL( database_api_impl, verify_signatures )
 
 DEFINE_API_IMPL( database_api_impl, find_transaction )
 {
-    // Yet to check the argument list
-    // FC_ASSERT ( args.size() == 1 || args.size() == 2, "Expected 1-2 arguments, was ${n}", ("n", args.size()) );
-    // auto trx_id = { args[0].as< transaction_id_type >() };
-    // auto expiration_time = { args[1].as< time_point_sec >() };
+    FC_ASSERT ( args.size() == 1 || args.size() == 2, "Expected 1-2 arguments, was ${n}", ("n", args.size()) );
```
mvandeberg: This is not how arguments in `database_api` work. This was copied from `condenser_api`, which is an old API format that parsed args as a json array. Now args are passed as a struct. Please see other `database_api` methods as an example of what to do.

libraries/plugins/apis/database_api/database_api.cpp
```diff
@@ -1425,40 +1425,94 @@ DEFINE_API_IMPL( database_api_impl, verify_signatures )
...
+    bool trx_in_mempool = itr_pending_tx != _db._pending_tx.end() ? true : false;
+    bool trx_in_index = itr_idx_tx != trx_idx.end() ? true : false;
+    
+    switch (args.size())
     {
-        try {
-            if (_db.is_known_transaction(trx_id))
+        case 1:
+            // 0 - Not aware of the transaction
+            if (!trx_in_mempool && !trx_in_index)
```
mvandeberg: `trx_in_mempool == true` implies `trx_in_index == true` because the current state is the head block plus all pending transactions, which means they will exist in the index.

libraries/plugins/apis/database_api/database_api.cpp
```diff
@@ -1425,40 +1425,94 @@ DEFINE_API_IMPL( database_api_impl, verify_signatures )
...
+            {
+                result.trx_status_code = transaction_status_codes::aware_of_trx;
+                break;
+            }
+            
+        case 2:
+            // 2 - Expiration time in future, transaction not included in block or mempool
+            if (!trx_in_mempool && !trx_in_index)
+            {
+                result.trx_status_code = transaction_status_codes::exp_time_future_trx_not_in_block_or_mempool;
```
mvandeberg: You cannot jump to this conclusion yet without looking at the expiration. A transaction that has not been included and is not in the mempool could have an expiration in the past. Depending on how far back we could conclude the transaction is irreversibly excluded.

libraries/plugins/apis/database_api/database_api.cpp
```diff
@@ -1425,40 +1425,94 @@ DEFINE_API_IMPL( database_api_impl, verify_signatures )
...
+            
+            // 5 - Transaction has been included in block, block is irreversible
+            if (!trx_in_mempool && trx_in_index
+                && expiration <= now - fc::seconds(STEEM_MAX_TIME_UNTIL_EXPIRATION))
+            {
+                result.trx_status_code = transaction_status_codes::trx_included_in_block_and_block_irreversible;
+                break;
+            }
+            // 6 - Transaction has expired, transaction is not irreversible (transaction could be in a fork)
+            if (!trx_in_mempool && !trx_in_index
```
mvandeberg: This case is unreachable due to case above.

libraries/plugins/apis/database_api/database_api.cpp
```diff
@@ -1425,40 +1425,94 @@ DEFINE_API_IMPL( database_api_impl, verify_signatures )
...
+                && expiration > now - fc::seconds(STEEM_MAX_TIME_UNTIL_EXPIRATION))
+            {
+                result.trx_status_code = transaction_status_codes::trx_expired_trx_is_not_irreversible;
+                break;
+            }
+            // 7 - Transaction has expired, transaction is irreversible (transaction cannot be in a fork)
+            if (!trx_in_mempool && !trx_in_index
```
mvandeberg: Also unreachable.

libraries/plugins/apis/database_api/include/steem/plugins/database_api/database_api_args.hpp
```diff
@@ -600,6 +595,15 @@ struct get_smt_next_identifier_return
 };
 #endif
 
+#define DEFINE_API_ARGS( api_name, arg_type, return_type )  \
+typedef arg_type api_name ## _args;                         \
+typedef return_type api_name ## _return;
+    
+/*               API,                                    args,                return */
+DEFINE_API_ARGS( find_transaction,                      vector< variant >,    find_transaction  )
```
mvandeberg: Do not use `vector<variant>` as argument type. Use a struct. Follow example of other `database_api` methods, not `condenser_api`.

Issue #2869, opened 2018-09-06
RPC calls are missing Content-Type


relativityboy:
As a developer using steem's api calls, I want to have a correctly set Content-Type header in the response to my RPC requests _so that_

* my application can determine what mechanism to use when parsing the response
* the generic applications I use in development can determine what mechanism to use when parsing the response, which makes me more efficient and frees up resources to invest in Steem value producing activities
* I have a sense of confidence in the blockchain I'm using

**AC**
- [ ] all JSON responses have the `Content-Type` header present and set to `application/json`

PR #2871, opened 2018-09-06
Set Content-Type header in RPC responses #2869

theoreticalbts:
PR for #2869

### DIFF
--- a/libraries/plugins/webserver/webserver_plugin.cpp
+++ b/libraries/plugins/webserver/webserver_plugin.cpp
@@ -251,11 +251,10 @@ void webserver_plugin_impl::handle_http_message( websocket_server_type* server, 
    {
       auto body = con->get_request_body();
 
-      con->append_header( "Content-Type", "application/json" );
-
       try
       {
          con->set_body( api->call( body ) );
+         con->append_header( "Content-Type", "application/json" );
          con->set_status( websocketpp::http::status_code::ok );
       }
       catch( fc::exception& e )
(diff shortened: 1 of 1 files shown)


libraries/plugins/webserver/webserver_plugin.cpp
```diff
@@ -251,6 +251,8 @@ void webserver_plugin_impl::handle_http_message( websocket_server_type* server,
    {
       auto body = con->get_request_body();
 
+      con->append_header( "Content-Type", "application/json" );
```
mvandeberg: This should be done after `con->set_body( api->call( body ) );` because the payload is only JSON after this point.

theoreticalbts:
Is there any path where a non-JSON payload could be returned?

mvandeberg:
Yes, the catch block. If there is some really bad error, we just return text.

relativityboy:
I don't know how challenging this would be, (taking it on is not appropriate for this issue) but typically a webserver will attempt to return information in the format requested. HTML/HTML JSON/JSON. Both success and error responses are in the Content-Type requested. 

In an ideal situation  internal messages would bubble up to the webserver and be encoded in the appropriate type. This provides a basis for very consumable messaging of all types.

A nice example

**Plain Text**
```
Error 5000000 - Bad param applesauce - did you check your config?
```

**JSON**
```json
{
  "success":false,
  "error": 5000000,
  "message": "Bad param applesauce",
  "userMessage": "did you check your config?"
}
```

Applications that can consume structured data have the opportunity to quickly check the `error` and utilize a recovery strategy, potentially logging (for developers) the `message` and /or choosing a 'nice' message for users to see (possibly something like `userMessage` though that is not a strong suggestion, just something we did at one of the places I worked that turned out to yield pretty excellent dividends in-terms of debugging time and thus overall productivity)

Issue #2722, opened 2018-08-06
Block Space Partitioning


youkaicountry:
The amount of block space that can be used for automated actions should be a fixed fraction of a block.

mvandeberg:
#3098 specifies 25% as the automated action reserved space.

mvandeberg:
Block validation for space partitioning needs to be done.

Issue #1360, opened 2017-08-08
Check for concurrent modification of boost multi-index


mvandeberg:
Several places in our code use the following pattern:

``` c++
while( /* condititon */ )
{
   auto obj = *itr;
   ++itr;
   db.remove( obj );
}
```

They should be refactored to a two pass algorithm that stores const pointers to objects to be removed in a vector and then iterates through the vector removing them.

https://github.com/steemit/steem/blob/master/libraries/chain/database.cpp#L3850-L3866 is an example of the correct form.

See discussion in #1355

PR #2903, opened 2018-09-14
AA Process Queue

### LINKED ISSUE #2709: Process AA Schedule Queue
Process the schedule queue each block. Examine the top to find actions that should go into the block.

mvandeberg:
Closes #2709

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -84,12 +84,14 @@ class database_impl
    public:
       database_impl( database& self );
 
-      database&                              _self;
-      evaluator_registry< operation >        _evaluator_registry;
+      database&                                       _self;
+      evaluator_registry< operation >                 _evaluator_registry;
+      evaluator_registry< required_automated_action > _req_action_evaluator_registry;
+      evaluator_registry< optional_automated_action > _opt_action_evaluator_registry;
 };
 
 database_impl::database_impl( database& self )
-   : _self(self), _evaluator_registry(self) {}
+   : _self(self), _evaluator_registry(self), _req_action_evaluator_registry(self), _opt_action_evaluator_registry(self) {}
 
 database::database()
    : _my( new database_impl(*this) ) {}
@@ -3204,9 +3206,10 @@ void database::_apply_block( const signed_block& next_block )
       dgp.current_witness = next_block.witness;
    });
 
-   automated_actions actions;
+   required_automated_actions req_actions;
+   optional_automated_actions opt_actions;
    /// parse witness version reporting
-   process_header_extensions( next_block, actions );
+   process_header_extensions( next_block, req_actions, opt_actions );
 
    if( has_hardfork( STEEM_HARDFORK_0_5__54 ) ) // Cannot remove after hardfork
    {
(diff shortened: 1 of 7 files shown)


theoreticalbts:
Reviewed.  Issues I found:

- When using `std::copy()`, you can't provide the `end()` iterator as an argument, because writing to that iterator will write past the end of the vector.  Instead you have to use `back_inserter` like the [cppreference std::copy example](https://en.cppreference.com/w/cpp/algorithm/copy).
- In `process_required_actions()`, if the queue empties but the block has more actions, an exception should be thrown (currently the method just returns).  I suggest changing the `while` loop to `while( actions_itr != actions.end() )` and inside the loop, `FC_ASSERT( pending_itr != pending_actions_index.begin() );`
- The `process_required_actions()` loop can be cleaned up to only assign `pending_itr = pending_actions_idx.begin()` once at the beginning of the loop (as opposed to twice, once outside the loop and again at the bottom of the loop)
- `#{o}` typo in format string
- AFACIT, `process_optional_actions()` applies actions without actually checking that the actions exist in the queue
- `generate_required_actions()` seems to be missing some boilerplate (`database::add_generate_required_actions_handler()` doesn't exist, `database::generate_required_actions()` is empty, contrast this with the corresponding optional actions code)
- `pushed_block_num` is non-default-initialized integer field in a database object
- `required_action` / `optional_action` should renamed to something like `example_required_action` / `example_optional_action`, it is confusing to have types `required_action` and `required_automated_action` which are different things

mvandeberg:
There is some explanation required here to motivate the PR as it exists. The caveat I have with this PR is that it is a contained part of a larger set of changes and some of the incremental pieces don't make sense outside of the larger whole. I will address those individually.

> In `process_required_actions()`, if the queue empties but the block has more actions, an exception should be thrown (currently the method just returns). I suggest changing the while loop to `while( actions_itr != actions.end() )` and inside the loop, `FC_ASSERT( pending_itr != pending_actions_index.begin() );`

This is absolutely correct. I was unsure where to place this requirement as it is related to #2722. There is a corner case where not all required actions are included in the block but we allow it because the partition of the block for required actions is full. @youkaicountry Please comment on the AC of this issue in relation to this requirement.

> The `process_required_actions()` loop can be cleaned up to only assign `pending_itr = pending_actions_idx.begin()` once at the beginning of the loop (as opposed to twice, once outside the loop and again at the bottom of the loop)

This would result in concurrent modification of the iterator. #1360

> AFACIT, `process_optional_actions()` applies actions without actually checking that the actions exist in the queue

This is correct and intentional. For simplicity of implementation, we will require required actions to be included in the same order. This will not be the case for optional actions. Optional actions can be delayed. It the first optional action is delayed and the next one is included then how do we verify inclusion? We could store the hash of an action, but it is possible that we could have duplicate actions. Probably a poor design, but I haven't ruled that out at this point. It makes the most sense for optional actions to have a simple `validate` and state verification in the evaluator to determine validity. This also allows a node that is brought online fresh to not need to keep track of optional actions as they are generated but still verify the chain state.

> `generate_required_actions()` seems to be missing some boilerplate (`database::add_generate_required_actions_handler()` doesn't exist, `database::generate_required_actions()` is empty, contrast this with the corresponding optional actions code)

If any node can verify optional actions then only witness nodes need to generate optional actions. I made the decision that optional action generation should be event driven and live in plugins that are required by the witness plugin and required actions be baked in to database logic.

> required_action / optional_action should renamed to something like example_required_action / example_optional_action, it is confusing to have types required_action and required_automated_action which are different things

This is fine. The structs exist as placeholder actions and will probably be used for testing.

theoreticalbts:
Approved, but see #3070.

PR #2914, opened 2018-09-19
Cleanup Irreversible and Checkpoint

mvandeberg:
Cleans up irreversible logic so reversible state is deleted last during block application.

Checkpoint logic was moved to `push_block` instead of `_apply_block`. This will short circuit on known bad blocks.

Made default fork_item constructor private to ensure field initialization on construction.

Fixed logic of `fork_db::fetch_block_by_number`


theoreticalbts:
All this looks reasonable.

abitmore:
As mentioned in https://github.com/bitshares/bitshares-core/issues/1333#issuecomment-423458124, after removed checkpoint code from `_apply_block`, replay could fail under certain circumstances.

Issue #2910, opened 2018-09-17
get_block API returns null


abitmore:
The API calls `database::fetch_block_by_number()` 

https://github.com/steemit/steem/blob/017698e615096afa04412582ec0a77369f7be6c2/libraries/plugins/apis/block_api/block_api.cpp#L49-L58

https://github.com/steemit/steem/blob/017698e615096afa04412582ec0a77369f7be6c2/libraries/chain/database.cpp#L356-L367

A block in fork db means it's not in block_log. When there are more than one block with same height in fork db, aka `results.size() > 1`, the function calls `_block_log.read_block_by_num` but won't find the block, thus will return null.

mvandeberg:
Closed in PR #2915

PR #2915, opened 2018-09-19
Fix fetch block by number #2910

theoreticalbts:


libraries/chain/database.cpp
```diff
@@ -3836,38 +3843,66 @@ void database::update_last_irreversible_block()
       }
    }
 
-   commit( dpo.last_irreversible_block_num );
-
    for( uint32_t i = old_last_irreversible; i <= dpo.last_irreversible_block_num; ++i )
```
fisher2015: When i= old_last_irreversible or last_irreversible_block_num, it will be notified twice.

PR #2925, opened 2018-09-21
NAI generation

### LINKED ISSUE #2886: NAI Generation
Numerical Asset Identifiers should be pseudorandomly generated. 
NAIs should be taken from a pool, and replaced upon taking - to avoid a race condition.
The NAI should be checked for collision via `by_symbol`.

sgerbino:
Closes #2886.


theoreticalbts:
LGTM

libraries/chain/database.cpp
```diff
@@ -4979,9 +4982,17 @@ void database::apply_hardfork( uint32_t hardfork )
             });
          }
          break;
-   #ifdef IS_TEST_NET
+FC_TODO( "We need to handle this if we have HF21 that is NOT an SMT hardfork" );
+#if 0
```
mvandeberg: The `#if 0` is not needed. That's why a default case is specified.

libraries/chain/database.cpp
```diff
@@ -5410,31 +5421,66 @@ void database::retally_witness_vote_counts( bool force )
...
+void database::replenish_nai_pool()
+{
+   try
+   {
+      nai_generator n;
+      while ( count< nai_pool_object >() < SMT_MAX_NAI_POOL_COUNT )
+      {
+         asset_symbol_type next_sym;
+         auto block_id = head_block_id();
+         auto seed = block_id._hash[0];
```
mvandeberg: Don't use `_hash[0]` as the seed. It is an encoding of the block num and is deterministic. I would probably use `_hash[4]`.

https://github.com/steemit/steem/blob/a8cd260e819dccca173993e4dc295229072082c5/libraries/protocol/block.cpp#L17-L25

libraries/chain/database.cpp
```diff
@@ -5410,31 +5421,66 @@ void database::retally_witness_vote_counts( bool force )
...
+         uint32_t collisions = 0;
+         do
+         {
+            if ( collisions++ >= SMT_MAX_NAI_GENERATION_TRIES )
```
mvandeberg: I am not a fan of nesting pre/post increment in an expression. Yes, that is what they were designed for, but I've seen too many OBO of this form and this is easily rewritten as:

```c++
if( collisions >= SMT_MAX_NAI_GENERATION_TRIE ) { ... }
collisions++;
```

This does the same thing, is just as efficient, and is much clearer.

libraries/chain/database.cpp
```diff
@@ -5410,31 +5421,66 @@ void database::retally_witness_vote_counts( bool force )
...
+            {
+               wlog( "Encountered ${collisions} collisions while attempting to generate NAI",
```
mvandeberg: Add to this message that the pool will attempt to regenerate next block. If this message every shows up there will be a flurry of issues created about it simply because it is not a common message. Make it as friendly as possible so if it does show up, admins don't freak out as much. I might even put it at the info level. It really isn't a warning. The code is working as intended. But we definitely want this logged for debugging.

libraries/chain/include/steem/chain/util/nai_generator.hpp
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include <steem/protocol/asset_symbol.hpp>
+
+namespace steem { namespace chain {
+   class nai_generator {
+      /** FNV-1a implementation https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function#FNV-1a_hash */
+      struct hasher {
```
mvandeberg: Is this struct used?

libraries/chain/include/steem/chain/util/nai_generator.hpp
```diff
@@ -0,0 +1,48 @@
...
+         }
+         
+      private:
+         // Recommended by http://isthe.com/chongo/tech/comp/fnv/
+         static const uint32_t prime         = 0x01000193; //   16777619
+         static const uint32_t offset_basis  = 0x811C9DC5; // 2166136261
+      } hash;
+
+   public:
+      asset_symbol_type operator()( uint32_t seed )
```
mvandeberg: Why the need for operator overloading? It reads strangely in database.cpp

libraries/chain/include/steem/chain/smt_objects/nai_pool_object.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <steem/chain/steem_object_types.hpp>
+#include <steem/protocol/asset_symbol.hpp>
+
+#ifdef STEEM_ENABLE_SMT
+
+namespace steem { namespace chain {
+
+   class nai_pool_object : public object< nai_pool_object_type, nai_pool_object >
```
mvandeberg: I am interested in your reasoning behind making each NAI in the pool a separate object rather than a singleton object that contains a fixed size array.

libraries/chain/database.cpp
```diff
@@ -5410,31 +5421,66 @@ void database::retally_witness_vote_counts( bool force )
...
+         });
 
-   return vector< asset_symbol_type >( 1, new_symbol );
+         FC_ASSERT( asset_symbol_exists_in_nai_pool( next_sym ), "Asset symbol should exist in NAI pool." );
+      }
+   }
+   FC_CAPTURE_AND_RETHROW()
 }
+
+bool database::asset_symbol_exists_in_nai_pool( asset_symbol_type a )
```
mvandeberg: If the NAI pool was a singleton, this method (with different logic) could live in the singleton.

libraries/chain/database.cpp
```diff
@@ -5410,31 +5421,66 @@ void database::retally_witness_vote_counts( bool force )
...
+   auto stripped_symbol_num = a.get_stripped_precision_smt_num();
+   auto it = idx.lower_bound( asset_symbol_type::from_asset_num( stripped_symbol_num ) );
+   return it != idx.end() && it->asset_symbol.get_stripped_precision_smt_num() == stripped_symbol_num;
+}
+
+bool database::asset_symbol_is_an_smt_token( asset_symbol_type a )
+{
+   const auto& idx = get_index< smt_token_index >().indices().get< by_symbol >();
+   auto stripped_symbol_num = a.get_stripped_precision_smt_num();
+   auto it = idx.lower_bound( asset_symbol_type::from_asset_num( stripped_symbol_num ) );
```
mvandeberg: Why use `lower_bound` and not `find`? You're checking for existence and not needing to iterate.

libraries/chain/database.cpp
```diff
@@ -5410,31 +5421,66 @@ void database::retally_witness_vote_counts( bool force )
...
+
+bool database::asset_symbol_exists_in_nai_pool( asset_symbol_type a )
+{
+   const auto& idx = get_index< nai_pool_index >().indices().get< by_symbol >();
+   auto stripped_symbol_num = a.get_stripped_precision_smt_num();
+   auto it = idx.lower_bound( asset_symbol_type::from_asset_num( stripped_symbol_num ) );
+   return it != idx.end() && it->asset_symbol.get_stripped_precision_smt_num() == stripped_symbol_num;
+}
+
+bool database::asset_symbol_is_an_smt_token( asset_symbol_type a )
```
mvandeberg: Is there any way we can rewrite this to not live in the database? Considering how short the method is, it seems like this could just be inlined wherever needed...

libraries/chain/include/steem/chain/util/nai_generator.hpp
sgerbino: The sole purpose is to set clear boundaries between the FNV-1a implementation and the NAI logic.

mvandeberg: The operator overloading makes this difficult to follow.

If you use bits from the block id, it is already a hash and can be used as is. There is no reason to hash the hash. At that point, if you know the first hash, the second hash is still deterministic. There is no extra entropy being added.

libraries/chain/smt_evaluator.cpp
```diff
@@ -118,6 +115,15 @@ void smt_create_evaluator::do_apply( const smt_create_operation& o )
       token.control_account = o.control_account;
       token.market_maker.token_balance = asset( 0, token.liquid_symbol );
    });
+
+   auto nai_from_pool = asset_symbol_type::from_asset_num( o.symbol.get_stripped_precision_smt_num() );
+   const nai_pool_object* npo = _db.find< nai_pool_object, by_symbol >( nai_from_pool );
+   FC_ASSERT( npo, "NAI must have existed in the pool." );
+   _db.remove( *npo );
+   FC_ASSERT( !_db.asset_symbol_exists_in_nai_pool( o.symbol ), "Asset symbol should no longer be in NAI pool." );
```
mvandeberg: This seems like an assertion that should live in a test case rather than the evaluator.

mvandeberg: Also, at this point in the method, you've looked up this symbol 3 times, when you only need one lookup.

libraries/chain/database.cpp
```diff
@@ -5410,31 +5421,66 @@ void database::retally_witness_vote_counts( bool force )
...
+{
+   try
+   {
+      nai_generator n;
+      while ( count< nai_pool_object >() < SMT_MAX_NAI_POOL_COUNT )
+      {
+         asset_symbol_type next_sym;
+         auto block_id = head_block_id();
+         auto seed = block_id._hash[0];
+         uint32_t collisions = 0;
```
mvandeberg: This allows for SMT_MAX_NAI_GENERATION_TRIES collisions per NAI needing to be generated. Is this intentional? In our earlier conversation I had envisions this as a max per block.

This implementation is also slightly problematic in the worst case because seed is always set on call. If there are multiple SMT creations in a block, the same collisions will be hit on every call.

Perhaps static variables, last_seed? That resets collisions and seed only when there is a new block?

libraries/chain/smt_evaluator.cpp
```diff
@@ -118,6 +115,15 @@ void smt_create_evaluator::do_apply( const smt_create_operation& o )
...
+
+   /* After consumption we replenish our NAI pool */
+   _db.replenish_nai_pool();
```
mvandeberg: We probably only want to replenish the pool when the transaction is not pending because the pool may generate an NAI that doesn't end up ever being in the pool in consensus. Use `_db.is_pending_tx()`.

libraries/chain/include/steem/chain/database.hpp
```diff
@@ -489,8 +489,11 @@ namespace steem { namespace chain {
          /**
           * @return a list of available NAIs.
          */
-         vector< asset_symbol_type > get_smt_next_identifier();
+         vector< asset_symbol_type > get_nai_pool();
```
mvandeberg: As best I can tell, this is only needed for the API and testing. Let's not have it as a method on the database then. Also, if the NAI pool object is a singleton, this vector is stored on the object and easily retrievable by any code that needs it without extra work.

libraries/chain/include/steem/chain/smt_objects/nai_pool_object.hpp
sgerbino: Changed this to be a singleton, similar to witness schedule.

libraries/chain/include/steem/chain/smt_objects/nai_pool_object.hpp
```diff
@@ -0,0 +1,46 @@
...
+   public:
+      template< typename Constructor, typename Allocator >
+      nai_pool_object( Constructor&& c, allocator< Allocator > a ) : nai_pool( a )
+      {
+         c( *this );
+      }
+
+      id_type id;
+
+      using t_nai_pool = chainbase::t_vector< asset_symbol_type >;
```
mvandeberg: use fc::array so there are no dynamic allocations needed.

libraries/chain/smt_objects/nai_pool.cpp
```diff
@@ -0,0 +1,93 @@
...
+   try
+   {
+      static block_id_type last_block_id = db.head_block_id();
+      static uint32_t collisions_per_block = 0;
+      static uint32_t attempts_per_block = 0;
+
+      // If this is the first time we're encountering this block, reset our variables and set the last block */
+      if ( last_block_id != db.head_block_id() )
+      {
+         last_block_id = db.head_block_id();
```
mvandeberg: This is minor, but `db.head_block_id()` does a database lookup. While it is likely in cache, it is still common for us to save such values locally, even if they are only used a couple of times.

``` c++
auto head_block_id = db.head_block_id();

if( last_block_id != head_block_id )
{
   last_block_id = head_block_id;
   collisions_per_block = 0;
   attempts_per_block = 0;
}
```

libraries/protocol/include/steem/protocol/asset_symbol.hpp
```diff
@@ -63,6 +70,7 @@ class asset_symbol_type
       };
 
       asset_symbol_type() {}
+      operator uint32_t() { return to_nai(); }
```
mvandeberg: I am a little worried about this introducing subtle bugs later on. uint32_t is typedeffed to a lot of things. What if this was made an explicit operator?

libraries/protocol/include/steem/protocol/asset_symbol.hpp
```diff
@@ -207,20 +216,34 @@ inline void to_variant( const steem::protocol::asset_symbol_type& sym, fc::varia
...
    {
-      auto v = var.as< std::vector< variant > >();
-      FC_ASSERT( v.size() == 2, "Expected tuple of length 2." );
+      FC_ASSERT( var.is_object(), "Expected an object." );
+
+      auto& o = var.get_object();
+
+      auto nai = o.find( ASSET_SYMBOL_NAI_KEY );
+      FC_ASSERT( nai != o.end(), "Expected key '${key}'.", ("key", ASSET_SYMBOL_NAI_KEY) );
+      FC_ASSERT( nai->value().is_string(), "Expected a string type." );
```
mvandeberg: Say which field you expected a string type in. (i.e. "Expected a string type for value 'nai'"). Do the same thing for the decimals field. Also, please check the asset serialization and if such error messages don't exist there, please add them while you're in this code.

libraries/protocol/include/steem/protocol/asset_symbol.hpp
sgerbino: I shared the same sentiment, which is why I brought this up to you earlier. However adding explicit to the operator prevents compilation.

mvandeberg: It doesn't build because there is an implicit conversion (precisely what we want to avoid).

Make the following change:

```c++
/**Comparison operators that allow to return the same object representation
 * for both liquid and vesting symbol/nai.
 */
struct vesting_liquid_less
{
   bool operator ()( const asset_symbol_type& lhs, const asset_symbol_type& rhs ) const
   {
      // Compare as if both symbols represented liquid version.
      return ( lhs.is_vesting() ? lhs.get_paired_symbol() : lhs ) <
             ( rhs.is_vesting() ? rhs.get_paired_symbol() : rhs );
   }

   bool operator ()( const uint32_t& lhs, const asset_symbol_type& rhs ) const
   {
      // Use the other operator, adding the same precision to both NAIs.
      return operator()( asset_symbol_type::from_nai( lhs, rhs.decimals() ), rhs );
   }

   bool operator ()( const asset_symbol_type& lhs, const uint32_t& rhs ) const
   {
      return operator()( lhs, asset_symbol_type::from_nai( rhs, lhs.decimals() ) );
   }
};
```

By removing the double uint32_t comparison we remove the implicit conversion to uint32_t only to convert back to an asset symbol.

libraries/chain/smt_objects/nai_pool.cpp
```diff
@@ -77,13 +80,16 @@ void replenish_nai_pool( database& db )
...
+   const auto end = nais.begin() + npo.num_available_nais;
+   auto it = std::find( nais.begin(), end, asset_symbol_type::from_asset_num( a.get_stripped_precision_smt_num() ) );
+   if ( it != end )
    {
+      auto index = std::distance( nais.begin(), it );
       db.modify( npo, [&] ( nai_pool_object& obj )
       {
-         obj.nai_pool.erase( it );
+         obj.num_available_nais--;
+         std::swap( obj.nais[ index ], obj.nais[ obj.num_available_nais ] );
```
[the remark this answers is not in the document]
sgerbino: I think it is correct, we are decrementing the `num_available_nais` before using it as an index.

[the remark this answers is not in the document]
sgerbino: See c76361a4f077a5a1fcd82f2ec0849d01aeaf16d3, if you really don't like it; I can revert that.

libraries/protocol/include/steem/protocol/asset_symbol.hpp
sgerbino: Makes sense, applying/testing now.

libraries/chain/smt_objects/nai_pool.cpp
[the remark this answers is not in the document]
mvandeberg: I still think it is worth setting the empty space to `asset_symbol_type()` and add an accompanying check in the existing unit tests.

PR #2944, opened 2018-09-25
Remove unused max_rc field

theoreticalbts:
This removes the unused `max_rc` field from the `rc_account_object` and properly sets the return value in the API.


avinash9992:
nice

PR #2946, opened 2018-09-25
gitignore additions and LF changes

reggaemuffin:
Add CLion configurations to gitignore to make setup easier. Also fix all cases of CRLF to support builds on WSL with CLion.

Nothing critical but I thought I contribute my findings on trying to build steem in WSL. Once I got it working completely (with all tests) I will add windows build instructions.


mvandeberg:
This is a stale PR and .gitignore was recently updated.

PR #2966, opened 2018-09-26
Require RC plugin to produce blocks or broadcast transactions #2965

### LINKED ISSUE #2965: Require RC plugin to produce blocks or broadcast transactions
Now that we know that RC basically works on mainnet and we're pretty sure we won't need to switch back to the old bandwidth algorithm in a hurry, we can remove some of the contingency options which allowed nodes to easily switch between the old and new bandwidth algorithms (e.g. #2648).

Node misconfigurations that don't check RC are a prerequisite for situations like #2963.  Let's make it harder for that kind of misconfiguration to occur, by no longer exposing options for broadcast and witness nodes to disable checking RC's.

theoreticalbts:
Closes #2965.

### DIFF
--- a/libraries/plugins/apis/network_broadcast_api/network_broadcast_api_plugin.cpp
+++ b/libraries/plugins/apis/network_broadcast_api/network_broadcast_api_plugin.cpp
@@ -11,6 +11,8 @@ void network_broadcast_api_plugin::set_program_options( options_description& cli
 void network_broadcast_api_plugin::plugin_initialize( const variables_map& options )
 {
    api = std::make_shared< network_broadcast_api >();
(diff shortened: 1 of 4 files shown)


libraries/plugins/apis/network_broadcast_api/network_broadcast_api_plugin.cpp
```diff
@@ -11,6 +11,8 @@ void network_broadcast_api_plugin::set_program_options( options_description& cli
 void network_broadcast_api_plugin::plugin_initialize( const variables_map& options )
 {
    api = std::make_shared< network_broadcast_api >();
+   FC_ASSERT( !appbase::app().get_plugin< rc::rc_plugin >().get_rc_plugin_skip_flags().skip_reject_not_enough_rc,
```
mvandeberg: I am assuming this addition in particular is just until the transaction confirmation API is implemented?

PR #2975, opened 2018-09-27
Properly count execution time #2972

### LINKED ISSUE #2972: count_resources() does not set result.resource_count[ resource_execution_time]
Inside the count_resources() function the array for resource_execution_time is not set.

https://github.com/steemit/steem/blob/master/libraries/plugins/rc/resource_count.cpp#L503

Something like
```
result.resource_count[ resource_execution_time ] += vtor.execution_time_count ;
```
is missing. 

A search https://github.com/steemit/steem/search?q=resource_execution_time&unscoped_q=resource_execution_time
showed that is not set somewhere else.

At the moment the RC costs does not depend on compute time.

The RC system uses two measurements to determine how much an operation should cost in terms of RCs: blockchain size, and state size. 

https://github.com/steemit/steem/blob/master/libraries/plugins/rc/rc_plugin.cpp#L292
`tx_info.usage.resource_count[resource_execution_time]` is always zero.

theoreticalbts:
Fixes #2972.

### DIFF
--- a/libraries/plugins/rc/resource_count.cpp
+++ b/libraries/plugins/rc/resource_count.cpp
@@ -395,6 +395,8 @@ void count_resources(
         size_info.transaction_object_base_size
       + size_info.transaction_object_byte_size * tx_size
       + vtor.state_bytes_count;
+
+   result.resource_count[ resource_execution_time ] += vtor.execution_time_count;
 }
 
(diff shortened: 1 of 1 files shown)


mvandeberg:
This merge is being postponed temporarily.

PR #2977, opened 2018-09-27
Increment rc_account.manabar immediately for positive RC changes #2961

### LINKED ISSUE #2961: Delegating SP does not increase current RC mana
I just delegated 100 SP to the @sm-qa account in order to allow it to be able to transact on the network. After the delegation it appears that the "max_rc" was updated appropriately however the "current_mana" was not updated and the account is still unable to transact until it recharges.

I don't know if this is how it was intended to work or not, but if it is intended then I will at least say that I disagree with it.

theoreticalbts:
Fixes #2961.

### DIFF
--- a/libraries/plugins/rc/rc_plugin.cpp
+++ b/libraries/plugins/rc/rc_plugin.cpp
@@ -883,7 +883,7 @@ void update_modified_accounts( database& db, const std::vector< account_name_typ
       db.modify( rc_account, [&]( rc_account_object& rca )
       {
          rca.last_max_rc = new_last_max_rc;
-         rca.rc_manabar.current_mana += std::min( drc, int64_t( 0 ) );
+         rca.rc_manabar.current_mana += std::max( drc, int64_t( 0 ) );
       } );
    }
 }


libraries/plugins/rc/rc_plugin.cpp
```diff
@@ -870,15 +870,20 @@ void rc_plugin_impl::on_pre_apply_operation( const operation_notification& note
...
       const rc_account_object& rc_account = db.get< rc_account_object, by_name >( name );
+
+      int64_t new_last_max_rc = get_maximum_rc( account, rc_account );
+      int64_t drc = new_last_max_rc - rc_account.last_max_rc;
+
       db.modify( rc_account, [&]( rc_account_object& rca )
       {
-         rca.last_max_rc = get_maximum_rc( account, rca );
+         rca.last_max_rc = new_last_max_rc;
+         rca.rc_manabar.current_mana += drc;
```
mvandeberg: This should only modify if drc is positive. Negative changes are already accounted for in the per satoshi mana changes for voting mana.

libraries/plugins/rc/rc_plugin.cpp
```diff
@@ -870,15 +870,20 @@ void rc_plugin_impl::on_pre_apply_operation( const operation_notification& note
...
       const rc_account_object& rc_account = db.get< rc_account_object, by_name >( name );
+
+      int64_t new_last_max_rc = get_maximum_rc( account, rc_account );
+      int64_t drc = new_last_max_rc - rc_account.last_max_rc;
+
       db.modify( rc_account, [&]( rc_account_object& rca )
       {
-         rca.last_max_rc = get_maximum_rc( account, rca );
+         rca.last_max_rc = new_last_max_rc;
+         rca.rc_manabar.current_mana += std::min( drc, int64_t( 0 ) );
```
mvandeberg: This should be std::max

PR #3016, opened 2018-10-01
Fixes build errors using clang 10 #3015

### LINKED ISSUE #3015: Build errors after updating to Apple LLVM version 10.0.0 (clang-1000.11.45.2)
There are two instances where classes with virtual functions are not final and do not have virtual destructors. This causes the latest Clang to fail during build time.

sgerbino:
Closes #3015.


theoreticalbts:
LGTM.

PR #3022, opened 2018-10-03
Disable bandwidth system

### LINKED ISSUE #3006: Remove old bandwidth system from witness plugin
Removing the old bandwidth system from the witness plugin will greatly simplify the code and should speed replays.

mvandeberg:
Closes #3006

### DIFF
--- a/libraries/plugins/witness/witness_plugin.cpp
+++ b/libraries/plugins/witness/witness_plugin.cpp
@@ -89,7 +89,7 @@ namespace detail {
       void on_pre_apply_operation( const chain::operation_notification& note );
       void on_post_apply_operation( const chain::operation_notification& note );
 
-      void update_account_bandwidth( const chain::account_object& a, uint32_t trx_size, const bandwidth_type type );
+//      void update_account_bandwidth( const chain::account_object& a, uint32_t trx_size, const bandwidth_type type );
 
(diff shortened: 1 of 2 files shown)


theoreticalbts:
LGTM

bobinson:
@nedsteem : IMHO a plan to releasing this, timelines etc must be announced to the witnesses. And also, will it need a replay ?

Smittyvb:
@bobinson Why? This will have no impact on Steem users, just cleaning up unused code.

nedsteem:
With this change, Bandwidth, which is already deprecated and replaced by RCs, is functionally being commented out. No replay would be needed. The benefit of this however is any subsequent replays would occur faster due to not needing to calculate available bandwidths.

bobinson:
Thank you @nedsteem  for the clarifications. The speed improvements are a welcome change and will save not just computational power but also save man hours spend as testing, deployment by witnesses everything will be lot more faster now.

@Smittyvb 

> Why?

Side effects, whether replay needed etc were not clear thus asked for the details.

>  This was already disabled, this just completely removes it from the code.

As discussed in #3029 we are not completely removing the code.

KINGdotNET:
@nedsteem Are we expecting a new update/release anytime soon?

Issue #3025, opened 2018-10-03
Fix command line interface compatibility breakage


theoreticalbts:
#3014 was unintentionally an incompatible change to the command-line interface.  The problem here is that, in some places, the option needs to be separately added to the config and command line options.  This cannot be done because [here](https://github.com/steemit/steem/blob/4a5142c47ee2d0ae886fec122b875eba86ae26d7/libraries/appbase/application.cpp#L85-L86) the plugin-specified options are added to both the config and command-line options.

So I propose the following refactoring:

- In every plugin's `set_program_options()`, have the common options for `cli` / `cfg` added to `cfg` as presently
- Add the line `cli.add(cfg);` to the plugin
- This is a pure refactoring, we've just moved the logic of "add everything in `cfg` to `cli`" from `appbase` to each individual plugin
- For every former `bool_switch()` option, after `cli.add(cfg);`, add `bool_switch()` to `cli`
- For every former `bool_switch()` option, after `cli.add(cfg);`, add `value<bool>` to `cfg`

theoreticalbts:
#3024 will probably have to be reverted when this is implemented.

PR #3026, opened 2018-10-03
Make tracking start configurable #3007

### LINKED ISSUE #3007: Don't calculate RC before HF20
RC are computed from the beginning of time.  This was useful for testing, but it should be disabled to speed replays.

theoreticalbts:
Fixes #3007.


theoreticalbts:
This adds a `value<bool>` option, so fixing #3025 should be based on this PR.

mvandeberg:
It only needs to be based on this if we want the option to be a switch in the cli.

Issue #2991, opened 2018-09-28
Community Suggestion - RC Budget Witness Parameters


TimCliff:
The current implementation of RC budgets requires all witnesses to use the same budgets, otherwise there will be inconsistencies of behavior between nodes. Also, each time a budget is changed, it requires a code change and replay.

It seems that a more optimal way of handling changes in the RC budgets would be for witnesses to vote on each budget via a witness parameter, and have the blockchain use the median value as the current budget.

- [ ] Create new witness parameters for the RC budgets.
- [ ] Update the RC budgets to be based on the median value across the witnesses.
- [ ] Update the witness parameter documentation to include the new parameters.

Smittyvb:
Just to be clear, would RCs still be implemented as a soft fork?

TimCliff:
The witness parameter values would need to become part of consensus (afaik) but their enforcement through the RC plugin could still be done via non-consensus / soft-fork.

iamsmooth:
I don't think this should be witness settable. One of the factors that go into RC limits is ability to run nodes which implicates all of those running nodes in the ecosystem, including those whose economic influence may far exceed their stake (such as large exchanges). Require all members of the ecosystem to buy into these limits via a node upgrade is more sensible than allowing witnesses to arbitrarily adjust (which in effect puts the decision mostly in the hands of the largest stakeholders).

Issue #308, opened 2016-08-19
Smart Contract Sidechain based Upon Wren?


bytemaster:
![image](https://cloud.githubusercontent.com/assets/347290/17824100/ff639090-662d-11e6-9822-5f3ec9481ca6.png)

TheDarkLightX:
Implemented as a side chain, why not? As long as it's not the only smart contract sidechain option, this way developers get to choose better languages in the future. WREN has good specs in terms of speed, but speed may not be an issue depending on what kind of smart contract you want to write.

theoreticalbts:
Remember the software development aphorism: "Plan to throw one away; you will anyway."

For the first sidechain(s), we'll probably have lots of issues and need to be able to iterate quickly. Then we'll end up with an ugly, un-extensible mess because we changed our minds four times about how it should be designed. Then we'll decide to rewrite it cleanly, using a sane, coherent architecture, now that we fully understand all the issues involved.

This means that, for the initial pass, developer productivity will be much more important than runtime performance, since it will be rewritten from scratch anyway. So instead of some new unfamiliar language, we should use Python (first choice), C++ (second choice) or JS (third choice).

I think I would go so far as to say that we don't even have a good enough idea of the requirements of a blockchain scripting language to be able to state with any certainty whether Wren is an appropriate choice.

Also, I question this performance benchmark -- it seems this only includes CPython. Pypy will be much much faster.

Finally, Wren is not widely used -- I certainly haven't heard of it. That means we may have to deal with:

- Library deficiencies (we need X, but no implementation is readily available)
- Support deficiencies (any Python problem can usually be solved with a Stack Overflow search, but this is much less likely for obscure languages)
- Network effect deficiencies (anyone looking to build a sidechain will first have to learn a new scripting language, increases barrier to entry)

liondani:
https://steemit.com/steem/@dana-edwards/formal-verification-of-smart-contracts-and-wren

talerecursion:
> This means that, for the initial pass, developer productivity will be much more important than runtime performance, since it will be rewritten from scratch anyway. So instead of some new unfamiliar language, we should use Python (first choice), C++ (second choice) or JS (third choice).

Or we could simply integrate the EVM and be done with it in no time. Many advantage in addition to shorter time to market: a lot of reuse from existing Ethereum smart contracts, many developers familiar with the architecture, compatible with ETH, ETC, Rootstock, Counterparty and other future projects using the EVM.

[...]

PR #3027, opened 2018-10-03
Reduce cost for all custom ops except follow custom ops

mvandeberg:


nedsteem:
Zooming out for context of this issue, could we begin or point to the proposal for cost parameters set by Witnesses?

theoreticalbts:
LGTM

theoreticalbts:
@nedsteem This PR is basically tweaking a number.  Making it settable by the witnesses is something that's been proposed by the community (issue #2991 ), but it involves more code and dev time than this quick fix.

bobinson:
@nedsteem It will be helpful to explain how we derived these values & their explanations IMHO.

1. custom_binary_operation_exec_time : what exactly is a custom binary ?
2. The comment says: `These numbers are obtained from computations in state_byte_hours.py script` but what exactly is script doing I feel should be explained to witnesses and they will explain to the end users/communities.

Smittyvb:
Custom binary is like custom JSON, but for any binary blob.

mvandeberg:
The state byte numbers are calculated from the python script.

The execution time numbers were derived from empirical testing. The previous custom json time was based on our full node (all plugins enabled) execution times, which includes significant execution times for custom ops used by the follow plugin. We want to account for those because they do incur significant costs to operating Steemit.com but we acknowledge that other Dapps make extensive use of custom ops and do not increase operating costs for Steemit.com at all. This change reduces the cost by 20x for all other custom ops, bring the cost in line with other lower impact operations.

bobinson:
@mvandeberg So this is a way to find the time complexity ? Further by limiting the time certain actions/operations takes, are we trying to achieve Turing Completeness ?  

Doubt:
> The execution time numbers were derived from empirical testing. 

So these numbers are like acceptable values based on observation and may vary based on CPU types but we expect the max values to hold hood in majority of the scenarios. Is this how the values are decided ?

Smittyvb:
@bobinson
> is a way to find the time complexity 

It determines how computationaly hard operations are.

> limiting the time

RCs don't limit how long nodes will process transactions. They just prevent a user from making too many operations that would overload nodes.

> Turing Completeness

What do you mean by this? C++ has been, and always has been turing complete. Steem users have never been able to run smart contracts.

bobinson:
thank you @Smittyvb 

> Steem users have never been able to run smart contracts.


Ok, I was trying to understand as there was #308 and also https://github.com/steemit/steem/commit/abc27b3f4c3e740cfc7e1265bc298bfc47163f81

PR #3028, opened 2018-10-03
cli options override cfg in cli

### LINKED ISSUE #3025: Fix command line interface compatibility breakage

mvandeberg:
Closes #3025

### DIFF
--- a/contrib/startpaassteemd.sh
+++ b/contrib/startpaassteemd.sh
@@ -127,7 +127,7 @@ if [[ $finished == 0 ]]; then
 fi
 
 # for appbase tags plugin loading
-ARGS+=" --tags-skip-startup-update=true"
+ARGS+=" --tags-skip-startup-update"
 
 cd $HOME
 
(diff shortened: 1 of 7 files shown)


theoreticalbts:
LGTM

PR #3030, opened 2018-10-03
Revert "Set value for non-switched config options"

mvandeberg:
Reverts steemit/steem#3024


theoreticalbts:
LGTM, not sure how Git/Github ate the previous version of this commit...

Issue #3004, opened 2018-09-30
accounts with 0 voting mana can't vote


roadscape:
Received a report that a claim-created account was unable to cast votes.

Smittyvb:
``@evoscoin`` is once such user. This is probably because claim-created users have 0 SP.

roadscape:
Should "advisory"/no-op votes be allowed at all? Given that SP delegation seems to be a core concern for UIs, faucets, and communities for the foreseeable future, how many legit users would ever have 0SP?

Other than it seems strange to allow no-op reward distribution core consensus actions, one practical concern is that there is no way for external services to efficiently filter spammy no-op votes (until we have something like https://github.com/steemit/steem/issues/2545)

If the goal is to allow 0SP users to interact with content, there are alternative solutions:
 - off-chain, up to each UI
 - on-chain in the form of some custom_json-based "advisory vote" standard

Issue #1764, opened 2017-11-13
Remove vote dust threshold


mvandeberg:
We are removing the vote dust threshold in HF 20.

We have two options in how to implement this. 

1. Do the vote as is
2. Floor the vote to 0 rshares

I am inclined towards 1. The vote dust threshold was instated during quadratic rewards. Votes under the threshold could add up to decent reward and should not be overlooked.

TimCliff:
I would much prefer option 2 (Floor the vote to 0 rshares), as it will discourage users from making lots of small dust votes which increases voting spam. If a user wants to make a 'dust vote', it would still be allowed and counted (so the user gets to express their like) but it will not contribute to the rewards until the user acquires more SP.

One other thing that would be good to consider along with this issue would be to apply the "floor of dust votes" shift amount to **all** votes. This would provide a (very small) disincentive for users to split their votes across multiple accounts, and would slightly favor users who consolidated their voting into a single account - which would also help to reduce vote spam.

A third thing to consider (which would be best to discuss in a separate issue if it is accepted) is to remove the limit on being able to power down an account, so that users who have lots of small dust accounts can power them down and consolidate.

iamsmooth:
Another suggestion that is related is to impose a fixed deduction/cost per vote, even for votes above the threshold. This will slightly disadvantage the smallest SP accounts with a degree of sub-linearity, but also discourage spam/dust voting and encourage consolidating stake into a smaller number of accounts/votes, at least to the extent it is reward-motivated (also may serve to increase incentives for the smallest SP accounts to power up). I still basically agree with the original comment here that users should be able to make zero rshares votes for social and UX purposes (subject to other bandwidth/spam limits), but maybe these should be implemented in some non-consensus way?

mvandeberg:
> This will slightly disadvantage the smallest SP accounts with a degree of sub-linearity

It is a slight degree of super-linearity. Not mathematically speaking, of course. If you were to graph the SP to rshare ratio, then it ends up being slightly super-linear with a linear limit.

mvandeberg:
[...]

Issue #248, opened 2016-08-05
Block all votes below a min rshare threshold


bytemaster:
This should prevent spam of meaningless votes.

jl777:
this does increase the cost by requiring each bot to be funded with more, so it filters out bots from unfinanced creators.

However, many bot writers have already made enough they would just fund their bots, so I was thinking of an additonal filter using a minimum threshold of posting awards. The logic being that bot posts tend to not get many upvotes. Probably need to also require having received upvotes from some minimum number of decent reputation score accounts to prevent the bot writer from upvoting his own bots.

This would prevent a person who hasnt posted any decent content from having any voting power also, but maybe that is not a bad thing

mvandeberg:
There are a couple problems with requiring posting history as a threshold for upvoting.

1. The barrier to upvote is already stake. Creating another barrier means that stake is no longer consensus metric for value of an account. The community can essentially preemptively ostracize new members by deciding to not upvote their content, effectively nullifying their stake. This barrier to entry may turn away potential speculators.

2. Having a hard number for a post payout is not an effective barrier against bots. It may actually hurt the platform in the long run. It creates a formula that if an account gets this many votes, it can then vote on other indefinitely. Create the barrier too low, and it does nothing. Create it too high and it excludes real people. Regardless of where it is, bot operators will know the exact amount of rshares needed to enable a new account for voting and be able to farm voting eligible accounts. While not voting on legitimate accounts. It will likely make the bot problem worse as there will be a higher overall percentage of rshares coming from bots rather than people.

Requiring min rshares is the best way to do this. Bots can get around by funding their accounts. But the more the fund, the more they are locked up and the benefits of short term gaming vs long term success of the platform begin to diminish. The goal of the change is not to kill botting, but to reduce blockchain spam.

jl777:
makes sense
so the aim is not a "captcha" but just to reduce spam

mvandeberg:
Correct. There are other measures, such as the super linear reward curve for content, that are designed to minimize the effects of botting. We have known from the very beginning that bots would use Steem, so we have been designing systems to reward human behavior that would be difficult to mimic using bots.

bartcant:
why not use a captcha  for voting ?

abitmore:
@bartcant you can run your own steemit.com server and remove the captcha. Everyone can. https://steemit.com/steemit/@teego/run-steemit-com-server-on-your-computer

Auxon:
[...]

PR #3034, opened 2018-10-05
Fix a bug on minimum voting mana requirement

sydneyitguy:
image

A new user will have 0 SP and 0 VP, but votings should go through even though it has zero effect on the rewards.

Currently the code change was missing, so new users are getting
`{“error”:“server_error”,“error_description”:“voter.voting_manabar.current_mana > 0: Account does not have enough mana to vote.“}` when they vote.

Temporary work-around: Delegate 0.001 SP to the new account

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -1794,7 +1794,7 @@ void hf20_vote_evaluator( const vote_operation& o, database& _db )
 
    if( !_db.has_hardfork( STEEM_HARDFORK_0_21__3034 ) )
    {
-      FC_ASSERT( voter.voting_manabar.current_mana > 0, "Account does not have enough mana to vote." );
+      FC_ASSERT( voter.voting_manabar.current_mana >= 0, "Account does not have enough mana to vote." );
    }
 
    int16_t abs_weight = abs( o.weight );
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -1791,7 +1791,7 @@ void hf20_vote_evaluator( const vote_operation& o, database& _db )
       util::manabar_params params( util::get_effective_vesting_shares( a ), STEEM_VOTING_MANA_REGENERATION_SECONDS );
       a.voting_manabar.regenerate_mana( params, now );
    });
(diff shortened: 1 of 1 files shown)


iamsmooth:
Is it even possible for voting mana to be negative? Perhaps a simpler change is to remove the test altogether, which seems more consistent with the current rules about 'dust votes'. Agree this needs to be a hard fork.

mvandeberg:
Voting mana cannot be negative be the amount consumed is a percentage of the remaining mana. In fact, I think it is impossible for voting mana to actually go to 0 if the user non-zero SP. This change is needed for the corner case of an account with SP, but does need to be gated by a hardfork check.

youkaicountry:
@sydneyitguy Please make the requested changes from code review, otherwise this PR will be closed in two weeks. We're trying to tidy up the open PRs, thanks for your understanding!

sydneyitguy:
@youkaicountry 

I'm not a blockchain developer, so I don't know how I can do "hardfork check".
However, I think this is a critical issue because new users cannot even vote once at the moment, so someone has to fix.. 

Can anyone help? (or should I re-create this as an issue ticket?)

mvandeberg:
A hardfork check is:

``` c++
if( db.has_hardfork( STEEM_HARDFORK_0_21 ) )
{
   // new logic
}
else
{
   // old logic
}
```

If you make such a change without this check then an upgraded node might have different acceptance criteria for an operation or might even have a different result in a calculation. In either case, this leads to a blockchain fork between new nodes and old nodes.

You can see this throughout the codebase. Ideally, this would also be filed as a bug report and the hardfork constant would include the issue number as well.

An example:

https://github.com/steemit/steem/blob/3730a839d88ead9e28da1642359f7054c479496c/libraries/chain/steem_evaluator.cpp#L1417-L1429

This contains all previous iterations of logic surrounding the STEEM_VOTE_DUST_THRESHOLD over several hardforks (13, 14, and 20). You can go back to the issues (#248, #259, and #1764) to see the motivation behind each change.

sydneyitguy:
@mvandeberg Thank you for your kind explanation 🙂
I have added a hard-fork check, and just removed the assertion on HF21 as you pointed out (voting mana cannot be negative value anyway).

mvandeberg:
Related to #3004

mvandeberg:
Fixed in #3004

PR #3046, opened 2018-10-09
Initial design of delegation pool data structures #2696

### LINKED ISSUE #2696: Implement Delegation Pool Objects
For delegation pools, implement the pool objects.

theoreticalbts:
Closes #2696.

### DIFF
--- a/libraries/plugins/rc/include/steem/plugins/rc/rc_objects.hpp
+++ b/libraries/plugins/rc/include/steem/plugins/rc/rc_objects.hpp
@@ -10,6 +10,10 @@
 
 #include <boost/multi_index/composite_key.hpp>
 
+namespace steem { namespace chain {
+struct by_account;
+} }
+
 namespace steem { namespace plugins { namespace rc {
 
 using namespace std;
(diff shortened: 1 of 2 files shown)


libraries/plugins/rc/include/steem/plugins/rc/rc_operations.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+   void validate()const;
+   void get_required_active_authorities( flat_set<account_name_type>& a )const { a.insert( from_account ); }
+};
+
+struct delegate_drc_from_pool_operation : base_operation
+{
+   account_name_type      from_pool;
+   account_name_type      to_account;
+   int64_t                mana_change = 0;
```
mvandeberg: Why int64_t instead of share_type?

libraries/plugins/rc/include/steem/plugins/rc/rc_objects.hpp
```diff
@@ -74,10 +84,79 @@ class rc_account_object : public object< rc_account_object_type, rc_account_obje
...
+      template< typename Constructor, typename Allocator >
+      rc_delegation_pool_object( Constructor&& c, allocator< Allocator > a )
+      {
+         c( *this );
+      }
+
+      id_type                       id;
+
+      account_name_type             account;
+      steem::chain::util::manabar   rc_pool_manabar;
```
mvandeberg: Thinking ahead, the pool object should cache the max rc_pool_mana to avoid needing to walk the indel objects on every regeneration.

Likewise, each user should probably have a cached total outdel and indel on the rc account.

theoreticalbts:
> Why `int64_t` instead of `share_type`?

I wanted to use the same type as the `current_mana` field of `manabar`.

> Thinking ahead, the pool object should cache the max rc_pool_mana to avoid needing to walk the indel objects on every regeneration.
>
> Likewise, each user should probably have a cached total outdel and indel on the rc account.

This is why it's hard to break this down into "design the objects, then design the algorithms" -- the two are closely inter-related.

I'd say put that in a different PR.  Any not-immediately-necessary changes to existing RC objects should probably wait until after `0.20.x` has been stable for a bit.

PR #3048, opened 2018-10-10
Increase resource limits

ajayyy:
This PR multiplies resource budgets by 10.

### DIFF
--- a/libraries/jsonball/data/resource_parameters.json
+++ b/libraries/jsonball/data/resource_parameters.json
@@ -3,7 +3,7 @@
   {
    "time_unit" : "blocks",
    "budget_time" : {"days" : 30},
-   "budget" : "300000000000",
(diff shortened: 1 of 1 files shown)


mvandeberg:
The problem we have currently is not with the budget but with price inelasticity.

You can think of the curves as a section of price elasticity where the quantity can change wildly without significantly impacting price and a section of price inelasticity where a small change in quantity can change the price significantly.

Price inelasticity is desirable for creating a consistent user experience. Price elasticity is important for throttling usage and protecting the blockchain against spam. Our curve implementation allows us to manipulate the crossover point from price inelasticity to elasticity. Moving it left or right will allow less or more utilization of the resource budget before throttling begins to really kick in.

I am going to close this in favor of our changes which should pushed soon. I invite you to take a look at our changes when the PR is made.

PR #3050, opened 2018-10-10
Enforce minimum RC

mvandeberg:


theoreticalbts:
I wasn't planning on doing it this way, but this works.

PR #3060, opened 2018-10-11
Transaction status API

### LINKED ISSUE #2458: Transaction polling API

sgerbino:
Closes #2458.


libraries/plugins/transaction_status/transaction_status_plugin.cpp
```diff
@@ -0,0 +1,110 @@
...
+      {
+         obj.block_num = note.block_num;
+      } );
+   }
+
+   // Remove elements from the index that are deemed too old for tracking
+   if ( note.block_num > block_depth )
+   {
+      uint32_t earliest_tracked_block = note.block_num - block_depth;
+      const auto& idx = _db.get_index< transaction_status_index >().indices().get< by_block_num >();
```
mvandeberg: Removing in a loop like this can be tricky. We haven't had any problems with concurrent modification, but the boost multi-index container documentation doesn't explicitly state support for concurrent modification. Because of this, I would restructure the loop as follows:

``` c++
auto itr = idx.begin();
while( itr != idx.end() && itr->block_num < earliest_tracked_block )
{
   _db.remove( *itr );
   itr = idx.begin();
}
```

This would also require changing the behavior of the block num field slightly to support a non-zero sentinel value such as `#define TX_IN_MEMPOOL ~0`

libraries/plugins/transaction_status/transaction_status_plugin.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+transaction_status_plugin::transaction_status_plugin() {}
+transaction_status_plugin::~transaction_status_plugin() {}
+
+void transaction_status_plugin::set_program_options( boost::program_options::options_description& cli, boost::program_options::options_description& cfg )
+{
+   cfg.add_options()
+      ( TRANSACTION_STATUS_BLOCK_DEPTH_KEY, boost::program_options::value<uint32_t>()->default_value( TRANSACTION_STATUS_DEFAULT_BLOCK_DEPTH ), "Defines the number of blocks from the head block that transaction statuses will be tracked." )
+      ;
+   cli.add_options()
```
mvandeberg: You don't need to add it twice. Appbase automatically copies all non-colliding keys from the config options to the cli options.

mvandeberg:
If you want, you can register the API plugin in the test as well and call the API methods as though they were normal methods to test the API at the same time.

libraries/plugins/apis/transaction_status_api/transaction_status_api.cpp
```diff
@@ -68,13 +68,16 @@ DEFINE_API_IMPL( transaction_status_api_impl, find_transaction )
...
+      {
+         auto last_irreversible_block = _db.fetch_block_by_number( last_irreversible_block_num );
+         if ( expiration <= last_irreversible_block->timestamp )
+            return {
+               .status = transaction_status::expired_irreversible
+            };
+      }
       // If our expiration is in the past
-      else if ( expiration < fc::time_point::now() )
+      if ( expiration < fc::time_point::now() )
```
mvandeberg: Don't check against fc::time_point::now(). Check against the head block time. What if the node gets behind by a couple blocks? The transaction could be contained in a block that has a head time in between the head block time and now.

libraries/plugins/apis/transaction_status_api/transaction_status_api.cpp
```diff
@@ -77,7 +74,8 @@ DEFINE_API_IMPL( transaction_status_api_impl, find_transaction )
             };
       }
       // If our expiration is in the past
-      if ( expiration < fc::time_point::now() )
+      auto head_block = _db.fetch_block_by_number( head_block_num );
```
mvandeberg: Use `_db.head_block_time()`. It returns the cached value in the dgpo.

tests/plugin_tests/transaction_status.cpp
```diff
@@ -206,7 +206,7 @@ BOOST_AUTO_TEST_CASE( transaction_status_test )
       BOOST_REQUIRE( api_return.block_num.valid() == false );
 
       api_return = tx_status_api->api->find_transaction( { .transaction_id = tx1.id(), tx1_expiration } );
-      BOOST_REQUIRE( api_return.status == expired_reversible );
+      BOOST_REQUIRE( api_return.status == unknown );
```
mvandeberg: I don't see any expired reversible tests now...

sgerbino: There is expiration time related tests at the bottom, it covers `expired_irreversible`, `expired_reversible`, and `too_old`.

theoreticalbts:
Just finished reviewing this.  Here's my feedback.  For each section, I've detailed whether it's a blocker or not.  When this PR is merged, any work mentioned here that is not satisfied by the PR should be filed as separate issue(s) / PR(s).

### too_old check bug

This PR is not mergeable until this bug is addressed.

The `too_old` [check](https://github.com/steemit/steem/blob/195c1cfbe37c31dd02dfdeb14f9f4c8b82ec78b5/libraries/plugins/apis/transaction_status_api/transaction_status_api.cpp#L62-L65) is incorrect.

- Alice makes a transaction, it expires at `05:01`.
- Alice's transaction gets into the chain at `04:59`.
- The blockchain continues with no forks or missed blocks, until the last block tracked by the plugin is `05:00`.
- Someone checks Alice's transaction status with `find_transaction`.
- It is reported that Alice's transaction is `expired_irreversible`.
- However, Alice's transaction did get into the chain.

### Other bugs

This PR is not mergeable until this bug is addressed.

- Potential overflow in [this line](https://github.com/steemit/steem/blob/195c1cfbe37c31dd02dfdeb14f9f4c8b82ec78b5/libraries/plugins/apis/transaction_status_api/transaction_status_api.cpp#L60).

### Code style

This PR is not mergeable until the top three items in this section are addressed.

- All indexes in `transaction_status_index` should be `ordered_unique`.  Using any index type other than `ordered_unique` is a common source of difficult bugs.
- The `uint32_t` / `bool` fields of `transaction_status_impl` should be default initialized (it is preferred to use `= 0` or `= false` when declaring integer / boolean fields)
- Use `STEEM_BLOCKS_PER_HOUR` from `config.hpp` instead of defining a different `BLOCKS_PER_HOUR` constant somewhere else.
- [This loop](https://github.com/steemit/steem/blob/195c1cfbe37c31dd02dfdeb14f9f4c8b82ec78b5/libraries/plugins/transaction_status/transaction_status_plugin.cpp#L68-L79) is simpler to understand (and less likely to have an off-by-one error lurking in `lower_bound()`) if you just remove the first block until enough blocks have been removed [1].  (This is a common pattern in queues.)

[1] Here is example code:

```
uint32_t obsolete_block = note.block_num - block_depth;
const auto& idx = _db.get_index< transaction_status_index >().indices().get< by_block_num >();
while( true )
{
   auto it = idx.begin();
   if( (it == idx.end()) || (it->block_num > obsolete_block) )
      break;
   _db.remove(*it);
}
```

### Handling state issues

This section may be best addressed in a separate PR, but I consider it a blocker for including this API in a numbered release.

At a minimum, the plugin should fail fast (i.e., refuse to load or provide an API) if the state file does not contain the right information.  This is a safer way to fail than having the API potentially report incorrect results if it's started from the wrong state file.

Alternatively, you could simply blow away all the objects if they don't exist or are out of date.  Then rebuild them by examining recent blocks.

### Additional return information

This section may be addressed in a separate PR.  It does not block merging or release.

- It may be useful to report the block ID / timestamp for the head block, LIB, and end-of-tracking-window block.  It is trivial to add this information to the API return struct.

### Performance

These optional performance improvements may be addressed in separate PR(s).  It does not block merging or release.

- Only read block header instead of entire block
- Optimize performance when replaying
- Calling `id()` is expensive, in `on_post_apply_block()` [this call](https://github.com/steemit/steem/blob/195c1cfbe37c31dd02dfdeb14f9f4c8b82ec78b5/libraries/plugins/transaction_status/transaction_status_plugin.cpp#L62) can be eliminated.

mvandeberg:
> - Only read block header instead of entire block

We don't actually have a way to read just a block header from the block log... (New issue #3095)

> - Optimize performance when replaying

What ideas do you have beyond what the `transaction-status-track-after-block` config option does?

libraries/plugins/transaction_status/include/steem/plugins/transaction_status/transaction_status_objects.hpp
```diff
@@ -55,8 +55,8 @@ typedef multi_index_container<
    transaction_status_object,
    indexed_by<
       ordered_unique< tag< by_id >, member< transaction_status_object, transaction_status_object_id_type, &transaction_status_object::id > >,
-      hashed_unique< tag< by_trx_id >, BOOST_MULTI_INDEX_MEMBER(transaction_status_object, transaction_id_type, transaction_id), std::hash<transaction_id_type> >,
-      ordered_non_unique< tag< by_block_num >, member< transaction_status_object, uint32_t, &transaction_status_object::block_num > >
+      ordered_unique< tag< by_trx_id >, member< transaction_status_object, transaction_id_type, &transaction_status_object::transaction_id > >,
+      ordered_unique< tag< by_block_num >, member< transaction_status_object, uint32_t, &transaction_status_object::block_num > >
```
mvandeberg: This needs to be a composite key on `(block_num, id)`.

libraries/plugins/transaction_status/transaction_status_plugin.cpp
```diff
@@ -48,43 +81,152 @@ void transaction_status_impl::on_post_apply_block( const block_notification& not
...
+   bool lower_bound_is_valid = true;
+   bool upper_bound_is_valid = true;
+
+   if ( earliest_tx.valid() )
+      lower_bound_is_valid = _db.find< transaction_status_object, by_trx_id >( *earliest_tx ) != nullptr;
+
+   if ( latest_tx.valid() )
+      upper_bound_is_valid = _db.find< transaction_status_object, by_trx_id >( *latest_tx ) != nullptr;
+   
+   return lower_bound_is_valid && upper_bound_is_valid ? true : false;
```
mvandeberg: Why not `return lower_bound_is_valid && upper_bound_is_valid;`

libraries/plugins/transaction_status/transaction_status_plugin.cpp
```diff
@@ -48,43 +81,152 @@ void transaction_status_impl::on_post_apply_block( const block_notification& not
...
+ * the necessary data to track transaction unless the state is rebuilt. We must clear out the
+ * existing objects and recreate the objects as if it would have during run-time.
+ */
+void transaction_status_impl::rebuild_state()
+{
+   ilog( "Rebuilding transaction status state" );
+
+   // Clear out the transaction status index
+   const auto& tx_status_idx = _db.get_index< transaction_status_index >().indices().get< by_trx_id >();
+   for ( auto& e : tx_status_idx )
```
mvandeberg: This does concurrent modification.

Alternate pattern:

``` c++
auto itr = tx_status_idx.begin();
while( itr != tx_status_idx.end() )
{
   _db.remove( *itr );
   itr = tx_status_idx.begin();
}
```

Or:

``` c++
while( true )
{
   auto itr = tx_status_idx.begin();

   if( itr == tx_status_idx.end() )
      break;

   _db.remove( *itr );
}
```

libraries/plugins/transaction_status/transaction_status_plugin.cpp
```diff
@@ -48,43 +81,152 @@ void transaction_status_impl::on_post_apply_block( const block_notification& not
...
+      _db.remove( e );
+
+   // Re-build the index from scratch
+   const auto head_block_num = _db.head_block_num();
+   uint32_t earliest_tracked_block_num = get_earliest_tracked_block_num();
+
+   for ( uint32_t block_num = earliest_tracked_block_num; block_num <= head_block_num; block_num++ )
+   {
+      const auto block = _db.fetch_block_by_number( block_num );
+
```
mvandeberg: Add `FC_ASSERT( block.valid(), "Could not read block ${n}", ("n", block_num) );`

libraries/plugins/transaction_status/transaction_status_plugin.cpp
```diff
@@ -93,8 +235,9 @@ transaction_status_plugin::~transaction_status_plugin() {}
...
 {
    cfg.add_options()
-      ( TRANSACTION_STATUS_BLOCK_DEPTH_KEY, boost::program_options::value<uint32_t>()->default_value( TRANSACTION_STATUS_DEFAULT_BLOCK_DEPTH ), "Defines the number of blocks from the head block that transaction statuses will be tracked." )
-      ( TRANSACTION_STATUS_TRACK_AFTER_KEY, boost::program_options::value<uint32_t>()->default_value( TRANSACTION_STATUS_DEFAULT_TRACK_AFTER ), "Defines the block number the transaction status plugin will begin tracking." )
+      ( TRANSACTION_STATUS_BLOCK_DEPTH_KEY,   boost::program_options::value<uint32_t>()->default_value( TRANSACTION_STATUS_DEFAULT_BLOCK_DEPTH ), "Defines the number of blocks from the head block that transaction statuses will be tracked." )
+      ( TRANSACTION_STATUS_TRACK_AFTER_KEY,   boost::program_options::value<uint32_t>()->default_value( TRANSACTION_STATUS_DEFAULT_TRACK_AFTER ), "Defines the block number the transaction status plugin will begin tracking." )
+      ( TRANSACTION_STATUS_REBUILD_STATE_KEY, boost::program_options::bool_switch()->default_value( false ),                                      "Indicates that the transaction status plugin must re-build its state upon startup." )
```
mvandeberg: Should follow pattern as `replay_blockchain` being only a CLI arg (https://github.com/steemit/steem/blob/295aa99d2bbdc9dd1ca35e7b59e2d192307c43ad/libraries/plugins/chain/chain_plugin.cpp#L317).

Or `enable-stale-production` (https://github.com/steemit/steem/blob/295aa99d2bbdc9dd1ca35e7b59e2d192307c43ad/libraries/plugins/witness/witness_plugin.cpp#L437-L448).

libraries/plugins/transaction_status/transaction_status_plugin.cpp
```diff
@@ -25,11 +48,21 @@ class transaction_status_impl
...
-   uint32_t                      track_after_block;
+   uint32_t                      nominal_block_depth = 0;       //!< User provided block-depth
+   uint32_t                      nominal_track_after_block = 0; //!< User provided track-after-block
+   uint32_t                      actual_block_depth = 0;        //!< Calculated block-depth
+   uint32_t                      actual_track_after_block = 0;  //!< Calculated track-after-block
    bool                          tracking = false;
+   bool                          rebuild_state_flag = false;
    boost::signals2::connection   post_apply_transaction_connection;
    boost::signals2::connection   post_apply_block_connection;
+   bool                          state_is_valid();
```
mvandeberg: Do `state_is_valid` and `rebuild_state` need to be made public? What use case is there for them being accessible here? Are they needed on the `transaction_status_plugin` struct at all?

mvandeberg: If it is only for tests, surround it in `#ifdef IS_TEST_NET`

Public/private specifiers in C++ are somewhat meaningless/easy to refactor. Putting private methods on an impl that lives only is source is a much more effective way of enforcing privacy of members.

sgerbino: Yes they are used just for testing. I wrapped the functions in `transaction_status_plugin` with `#ifdef IS_TEST_NET`, and left them  alone in the `transaction_status_impl` class.

theoreticalbts:
The nominal / actual variables are confusingly named.  I suggest the following:

- Rename`actual_track_after_block` -> `track_after_block`
- Rename `actual_block_depth` -> `block_depth`
- Make `nominal_` variables local variables, to be clear that it's just a config interface, the actual behavior's completely determined by the (renamed) `actual_` variables

This, and the remaining items from my previous review, can be addressed in separate PR(s).

This is ready to merge.

PR #3080, opened 2018-10-17
fix typo in operation_tests.cpp

mvandeberg:
Local PR for #2933

> Not sure if this changes anything, but a non-existing key maximim_block_size was removed from the props, so maximum_block_size was still set for all following tests.


theoreticalbts:
LGTM.

PR #3094, opened 2018-10-19
Revert the NAI reversion

### LINKED ISSUE #2886: NAI Generation

sgerbino:
Closes #2886.


theoreticalbts:
LGTM.

PR #3114, opened 2018-10-26
Event Driven Automated Actions

### LINKED ISSUE #2718: Event Driven Automated Action Push
Automated actions should be pushed to the schedule queue in an event driven manner.

mvandeberg:
Closes #2718

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -918,7 +918,7 @@ struct action_validate_visitor
    }
 };
 
-void database::push_required_action( const required_automated_action& a, const time_point_sec execution_time )
+void database::push_required_action( const required_automated_action& a, time_point_sec execution_time )
 {
    FC_ASSERT( execution_time >= head_block_time(), "Cannot push required action to execute in the past. head_block_time: ${h} execution_time: ${e}",
       ("h", head_block_time())("e", execution_time) );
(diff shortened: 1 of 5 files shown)


theoreticalbts:
Checking `FC_ASSERT( execution_time <= head_block_time())` is wrong direction of inequality.  This occurs in two places.

Between the implementation and the specification, there are no less than three different ideas about when objects are cleared from the queue of pending optional actions:

- (a) Actions are cleared at the end of a round, i.e. all pending actions should be cleared when the witness schedule is updated.
- (b) Actions are cleared when they are more than 63 seconds, i.e. the timestamp is checked against `head_block_time` every block.
- (c) Actions are cleared when they are earlier than the LIB timestamp.

You might be able to prove that some of these conditions result in equivalent observable behavior, but this is not obvious.  I think the spec needs to be clarified.

mvandeberg:
Regarding the expiration of optional actions:

I am not sure where you are getting (a) from. I cannot find it in the code or the issue.

(b) This was my original thought but during implementation decided to go with (c) to account for case that 63 seconds can actually include only half the witnesses if the same 11 witness that produced in the second half of the last round produce in the first half of the current round. This is the exact problem that our irreversibility logic already solves, so it should be used here as well. I will update the issue to specify this.

theoreticalbts:
> `while( pending_optional_itr != pending_optional_action_idx.end() && pending_required_itr->execution_time <= when )`

Looks like a copy-paste error here.

> `const time_point_sec execution_time`

I'm not sure our coding style permits `const` by-value parameters, perhaps you intended `const time_point_sec& execution_time` ?

mvandeberg:
It probably was `const time_point_sec& execution_time` and I removed the reference because it doesn't make sense to implicitly dereference a stack value to pass a 32bit value with a 64 bit reference. It is such a small thing. It should just be `time_point_sec execution_time` because there is no need for it to be a reference and a const value (while valid c++) makes no sense as a parameter.

Issue #3084, opened 2018-10-18
Platform Independent State Files


youkaicountry:
Write a platform independent state save / load system.

Acceptance Criteria:
- Command-line switch to launch a node into "saving mode" in which node processes to a given block, terminates, and writes state file
- State files are binary with JSON header
- Header contains list of active plugins, and hash of state affecting settings for each plugin
- Loading a state file checks the header against its own plugins / settings and halts if there is a difference
- State file writing code should append only, never seek

mvandeberg:
Instead of having plugins add a hash of their options impacting state, I am having them upload the json of the state. This matches the methodology with the schema checking. It will aid in debugging and will provide a technical error message when state files are rejected.

For the flags, I am implementing the following.

`--stop-replay-at-block` is being renamed to `--stop-at-block` and will be expanded to terminate steemd at a particular block height in all modes of operation (instead of just steemd).

`--to-state` will save state whenever steemd exits (primarily used in conjunction with `--stop-at-block`, but could be used with a monitor script to send SIGINT at a particular block height or when a certain set of conditions are met.

`--replay` can be combined with the previous two in order to reindex to a particular block height and then generate a state file. 

We should add a feature that if the block height is less than the head of the block log, the node will go in to a resumed reindex mode. Some infrastructure is needed to be added to the reindex code so that if SIGINT is send during a reindex, it can cleanly shut down and resume. (#3519)

PR #3122, opened 2018-10-29
Separate source file for core indexes, misc. cleanup

theoreticalbts:
In order to get this to compile:

- I did some minor cleanup of the `chainbase` code (adding a missing `#include`, fixing whitespace, changing from C style to C++ style include guards)
- Add a header specifically for forward declaration of custom FC serializations implemented by Steem, and move all serializers to it.  FC serialization is very picky about declaration order.
- Implement custom serializers for `bip::flat_map`, `bip::deque` and `shared_string`.  (Not strictly necessary, but this PR is a subset of my integration branch for #3084 which does need them.  Extracting these new serializers into a separate PR would be more trouble than it's worth).

### DIFF
--- a/libraries/chain/include/steem/chain/steem_fwd.hpp
+++ b/libraries/chain/include/steem/chain/steem_fwd.hpp
@@ -1,6 +1,6 @@
 #pragma once
 
-// This header forward-declare spack/unpack and to/from variant functions for Steem types.
+// This header forward-declares pack/unpack and to/from variant functions for Steem types.
 // These declarations need to be as early as possible to prevent compiler errors.
 
 #include <chainbase/allocators.hpp>
@@ -31,17 +31,5 @@ template<typename Stream, typename K, typename V, typename C, typename A>
 void pack( Stream& s, const boost::interprocess::flat_map< K, V, C, A >& value );
 template<typename Stream, typename K, typename V, typename C, typename A>
 void unpack( Stream& s, boost::interprocess::flat_map< K, V, C, A >& value );
-
-/*
-template<typename Stream, typename K, typename V, typename C, typename A>
-void pack( Stream& s, const boost::interprocess::map< K, V, C, A >& value );
-template<typename Stream, typename K, typename V, typename C, typename A>
-void unpack( Stream& s, boost::interprocess::map< K, V, C, A >& value );
-
-template<typename Stream, typename E, typename C, typename A>
-void pack( Stream& s, const boost::interprocess::set< E, C, A >& value );
-template<typename Stream, typename E, typename C, typename A>
-void unpack( Stream& s, boost::interprocess::set< E, C, A >& value );
-*/
 
 } }
--- a/libraries/chain/include/steem/chain/steem_fwd.hpp
+++ b/libraries/chain/include/steem/chain/steem_fwd.hpp
@@ -1,0 +1,47 @@
+#pragma once
+
+// This header forward-declare spack/unpack and to/from variant functions for Steem types.
+// These declarations need to be as early as possible to prevent compiler errors.
+
+#include <chainbase/allocators.hpp>
+
+namespace chainbase {
+template< typename T >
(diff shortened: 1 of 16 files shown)


theoreticalbts:
Added a one-liner (well, two-liner) to trivially implement the heretofore unimplemented `abstract_object` constructor/destructor.  The implementations live in the newly created `index.cpp`.

libraries/chain/include/steem/chain/steem_fwd.hpp
```diff
@@ -0,0 +1,47 @@
...
+template<typename Stream, typename K, typename V, typename C, typename A>
+void pack( Stream& s, const boost::interprocess::map< K, V, C, A >& value );
+template<typename Stream, typename K, typename V, typename C, typename A>
+void unpack( Stream& s, boost::interprocess::map< K, V, C, A >& value );
+
+template<typename Stream, typename E, typename C, typename A>
+void pack( Stream& s, const boost::interprocess::set< E, C, A >& value );
+template<typename Stream, typename E, typename C, typename A>
+void unpack( Stream& s, boost::interprocess::set< E, C, A >& value );
+*/
```
sgerbino: Should this commented code just be removed?

libraries/chain/include/steem/chain/steem_object_types.hpp
```diff
@@ -203,6 +203,69 @@ namespace fc
...
+
+      template< typename Stream, typename K, typename V, typename C, typename A >
+      inline void pack( Stream& s, const boost::interprocess::flat_map< K, V, C, A >& value )
+      {
+         fc::raw::pack( s, unsigned_int((uint32_t)value.size()) );
+         auto itr = value.begin();
+         auto end = value.end();
+         while( itr != end )
+         {
+           fc::raw::pack( s, *itr );
```
mvandeberg: This looks copied directly from FC's map pack, which is great (fewer opportunities for error), but it copied over the inconsistent indentation of FC as well... The internals of the while loop are only indented twice.

libraries/chain/include/steem/chain/steem_object_types.hpp
```diff
@@ -203,6 +203,69 @@ namespace fc
...
+      inline void unpack( Stream& s, boost::interprocess::flat_map< K, V, C, A >& value )
+      {
+         unsigned_int size;
+         unpack( s, size );
+         value.clear();
+         FC_ASSERT( size.value*(sizeof(K)+sizeof(V)) < MAX_ARRAY_ALLOC_SIZE );
+         value.reserve(size.value);
+         for( uint32_t i = 0; i < size.value; ++i )
+         {
+             std::pair<K,V> tmp;
```
mvandeberg: Indented 4 times

libraries/chainbase/include/chainbase/allocators.hpp
```diff
@@ -11,14 +11,15 @@
 #include <boost/interprocess/sync/interprocess_sharable_mutex.hpp>
 #include <boost/interprocess/sync/sharable_lock.hpp>
 #include <boost/interprocess/sync/file_lock.hpp>
+#include <boost/thread/locks.hpp>
```
mvandeberg: I do not see any changes in chainbase that would require adding this include.

theoreticalbts: Everything that currently includes `allocators.hpp` includes `locks.hpp` first (possibly indirectly).  This PR creates `steem_fwd.hpp`, which is supposed to always be included first.  And `steem_fwd.hpp` includes `allocators.hpp`.

So compilation will fail if you remove this `include`.

mvandeberg: I hate how lazy we have been in the past with implicit inclusions. This is not the first time a refactor has revealed that a header needs to include code that it doesn't and only works because every source file including it happens to include the file it needs.

Issue #3123, opened 2018-10-29
int_array::from_variant assumes integer type


theoreticalbts:
I implemented the `int_array` class for `rc_plugin`.  It is basically the same as `fc::array`, but with default initialization for integer values, and better JSON serialization.

The JSON deserialization compares to integer bounds.  This causes `int_array` to fail compilation when generating the code to deserialize non-integer types, as is actually done in `rc_plugin`.  (`int_array` should probably be renamed to something else since it now works with non-integer types.  I'm open to suggestions.)

mvandeberg:
Closed in #3124

PR #3124, opened 2018-10-29
Fix int_array from_variant

theoreticalbts:
PR for #3123.

This also includes a couple tiny commits from the #3084 integration branch, namely, fixing a couple objects that didn't use the correct paradigm for multi-index container tag structs.

Specifically:

- `hardfork_property_object` had an anonymous index rather than `by_id`
- `steem::plugins::reputation` created its own `by_id` and `by_account` structs, rather than using the ones in `steem::chain`

### DIFF
--- a/libraries/chain/include/steem/chain/hardfork_property_object.hpp
+++ b/libraries/chain/include/steem/chain/hardfork_property_object.hpp
@@ -32,7 +32,7 @@ namespace steem { namespace chain {
    typedef multi_index_container<
       hardfork_property_object,
       indexed_by<
-         ordered_unique< member< hardfork_property_object, hardfork_property_object::id_type, &hardfork_property_object::id > >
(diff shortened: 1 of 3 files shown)


mvandeberg:
Let's plan on moving the changes from `fc::int_array` to `fc::array`. Hopefully none of the code that currently relies on `fc::array` relies on uninitialized fields to work correctly. For now, this looks good.

mvandeberg:
Where this belongs in fc is an interesting conundrum. Default initializing for integer types is great, but if you throw structs in this with their own default values, it 0s them out, which could lead to some unintended side effects.

Maybe the correct course of action is to refactor non integer instances of int_array into fc::array to utilize default object initialization and modify the constructor of `int_array` to only work with integer types so that code fails compilation if a non integer type is templated into `int_array`.

mvandeberg:
FC's array types need some cleaning up, but it is beyond the scope of the original issue. I have created #3128 so we can address it later.

PR #3125, opened 2018-10-30
Remove vesting liquid less class

### LINKED ISSUE #3100: Remove vesting_liquid_less class
[This struct](https://github.com/steemit/steem/blob/791e76d1f193c336b075fcd0ef118f8874f85479/libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp#L165-L179) is used to implement an operator which compares `asset_symbol_type` to `asset_symbol_type` and also accepts a `uint32_t`.

This is a very unclean, poorly defined interface.  I'm not sure why it was implemented this way.

I propose the following rules:

- In general, fields should only be eligible to contain *either* vesting *or* non-vesting assets.  Not both.  (Perhaps we could come up with a way to enforce this at compile time.)
- If you compare vesting to non-vesting symbols, the symbols should be ordered in order of their symbols.
- What if you need to query `smt_token_index` for a liquid symbol, but you could have either a liquid or a vesting symbol?  This is a "code smell" and probably means your code needs to be redesigned.  Nonetheless, if the design is acceptable, you could use an appropriate `asset_symbol_type` method or arithmetic to explicitly normalize the symbol before querying.
- What if you need to query `smt_token_index` using an integer?  This is a "code smell" and probably means your code needs to be redesigned.  Nonetheless, if the design is acceptable, you should explicitly convert it to `asset_symbol_type`.  There shouldn't be automagic acceptance of it by the comparator.

I recommend the following implementation strategy:

- Delete `vesting_liquid_less` from the code
- Let any code which previously used this comparator use the default comparison operators of `asset_symbol_type`
- Investigate any resulting compilation or test errors.
- For each error, consider carefully whether it is an indication that some part of the code needs to be redesigned.
- If it does not need to be redesigned, add explicit conversion code as appropriate.

sgerbino:
Closes #3100.

### DIFF
--- a/libraries/chain/include/steem/chain/util/smt_token.hpp
+++ b/libraries/chain/include/steem/chain/util/smt_token.hpp
@@ -1,0 +1,13 @@
+#pragma once
+#include <stdint.h>
+#include <fc/optional.hpp>
+#include <steem/chain/database.hpp>
+#include <steem/chain/smt_objects/smt_token_object.hpp>
+#include <steem/protocol/asset_symbol.hpp>
+
+namespace steem { namespace chain { namespace util {
+
+const smt_token_object* smt_token_lookup( database& db, uint32_t nai );
+const smt_token_object* smt_token_lookup( database& db, asset_symbol_type symbol, bool precision_agnostic = false );
+
+} } } // steem::chain::util
--- a/libraries/chain/include/steem/chain/util/smt_token.hpp
+++ b/libraries/chain/include/steem/chain/util/smt_token.hpp
@@ -1,6 +1,7 @@
 #pragma once
-#include <stdint.h>
-#include <fc/optional.hpp>
+
+#ifdef STEEM_ENABLE_SMT
+
 #include <steem/chain/database.hpp>
 #include <steem/chain/smt_objects/smt_token_object.hpp>
 #include <steem/protocol/asset_symbol.hpp>
@@ -11,3 +12,5 @@ const smt_token_object* smt_token_lookup( database& db, uint32_t nai );
 const smt_token_object* smt_token_lookup( database& db, asset_symbol_type symbol, bool precision_agnostic = false );
 
 } } } // steem::chain::util
+
+#endif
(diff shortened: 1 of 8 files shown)


libraries/chain/util/smt_token.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#ifdef STEEM_ENABLE_SMT
+
+namespace steem { namespace chain { namespace util {
+
+const smt_token_object* smt_token_lookup( database& db, uint32_t nai )
+{
+   const auto& idx = db.get_index< smt_token_index >().indices().get< by_symbol >();
+
+   auto range = idx.range(
```
mvandeberg: The lower 4 bits are the precision which means `lower_bound( asset_symbol_type::from_nai( nai, 0 ) );` will go straight to the correct value. I think this makes a lot more sense than creating a range that you know only has one element and returning it in a for loop.

libraries/chain/util/smt_token.cpp
```diff
@@ -0,0 +1,35 @@
...
+   return nullptr;
+}
+
+const smt_token_object* smt_token_lookup( database& db, asset_symbol_type symbol, bool precision_agnostic )
+{
+   // Only liquid symbols are stored in the smt token index
+   auto s = symbol.is_vesting() ? symbol.get_paired_symbol() : symbol;
+
+   if ( precision_agnostic )
+      return smt_token_lookup( db, s.to_nai() );
```
mvandeberg: I would not do a conversion to nai to got back to `asset_symbol_type` when all you really want to do is 0 out the lower 4 bits.

mvandeberg:
Never mind my previous feedback. Your implementation hides all of the bit magic well. You would have to reimplement the helper functions in an effort to appear better documented, but ultimately create a maintenance nightmare.

libraries/chain/include/steem/chain/util/smt_token.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#ifdef STEEM_ENABLE_SMT
+
+#include <steem/chain/database.hpp>
+#include <steem/chain/smt_objects/smt_token_object.hpp>
+#include <steem/protocol/asset_symbol.hpp>
+
+namespace steem { namespace chain { namespace util {
+
+const smt_token_object* smt_token_lookup( database& db, uint32_t nai );
```
mvandeberg: Change these to `find_smt_token` to match naming convention of other database helper methods.

PR #3127, opened 2018-10-30
Automated Actions RC Scheduling

### LINKED ISSUE #2719: Event Driven Automated Action RC scheduling
Event Driven Automated Action RC scheduling

mvandeberg:
Closes #2719


theoreticalbts:
Looks like `on_post_apply_optional_action()` has a lot of copy-pasted code.  I wonder if it's possible to refactor this into a function that can be used multiple places?

Other than that, LGTM.

mvandeberg:
There are bits and pieces, but the largest part comes from `on_post_apply_transaction`. Everywhere that we loop over operations gets replaces with a single call for the action. There isn't a good way to template the code.

PR #3129, opened 2018-10-30
Implement state saving #3084

theoreticalbts:
Progress toward #3084.  Should be mergeable, but so far only implements saving to JSON.  It does not yet implement binary format or loading.


libraries/plugins/chain/chain_plugin.cpp
```diff
@@ -315,6 +318,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
          ("checkpoint,c", bpo::value<vector<string>>()->composing(), "Pairs of [BLOCK_NUM,BLOCK_ID] that should be enforced as checkpoints.")
          ("flush-state-interval", bpo::value<uint32_t>(),
             "flush shared memory changes to disk every N blocks")
+         ("from-state", bpo::value<string>()->default_value(""), "Load from state, then replay subsequent blocks (EXPERIMENTAL)")
+         ("to-state", bpo::value<string>()->default_value(""), "File to save state after --stop-replay-at-block" )
```
mvandeberg: Maybe we should simply have steemd save the state on shutdown when this is specified. It will work the same when `--stop-replay-at-block` is set, but will also make it easier to generate independent state files from the head block without needing to reindex.

theoreticalbts: So the use-case is, "I want to generate a state file when I send the process SIGINT, and I don't know ahead of time when that will be."  Sounds like a separate feature to me (should be a fairly trivial patch though).

theoreticalbts:
Needs to add code for the statefile versioning of plugins.

mvandeberg:
@sgerbino This is ready for review

PR #3134, opened 2018-10-31
Set is_producing flag when applying transactions from mempool

### LINKED ISSUE #3133: is_producing not set when re-applying pending transaction state.
We noticed in our logs some logging like the following:

```
rc_plugin.cpp:261             operator()           ] Accepting transaction by shan96, has -75995440 RC, needs 419358673 RC, block 27296438, witness cervantes.
rc_plugin.cpp:261             operator()           ] Accepting transaction by shan96, has -495354113 RC, needs 417995011 RC, block 27296438, witness cervantes.
```

It appears that the transaction is being applied multiple times in the same block with different RC values. Upon investigation we realized that the `is_producing` flag is not being set when reapplying transactions from the mempool to the pending transaction state after applying a block. Transactions that were valid are becoming invalid after block application and sticking around in the pending transaction queue until expiration. We suspect this was probably happening with bandwidth, but because this event was not being logged, we did not notice it. The solution is to set the `is_producing` flag when re-applying pending transactions to remove newly rejected transactions from the mempool.

mvandeberg:
Closes #3133

### DIFF
--- a/libraries/chain/include/steem/chain/db_with.hpp
+++ b/libraries/chain/include/steem/chain/db_with.hpp
@@ -53,6 +53,8 @@ struct pending_transactions_restorer
 
    ~pending_transactions_restorer()
    {
+      bool producing = _db.is_producing();
+      _db.set_producing( true );
       for( const auto& tx : _db._popped_tx )
       {
          try {
@@ -93,6 +95,7 @@ struct pending_transactions_restorer
             */
          }
       }
+      _db.set_producing( producing );
    }
 
    database& _db;


abitmore:
This doesn't seem right to me. IIRC `is_producing` was disabled by default, but here you assume that all nodes should enable it when processing pending transactions.

mvandeberg:
All nodes already set `is_producing` during `push_transaction`.

mvandeberg:
I checked git blame and it looks like the behavior was first introduced May 25, 2016

https://github.com/steemit/steem/blob/dc60a047135c2f42a0ec2d3a566775f680ad8b34/libraries/chain/database.cpp#L578

theoreticalbts:
> I checked git blame and it looks like the behavior was first introduced May 25, 2016

I'm pretty sure this behavior's been in Steem since forever.  Also, I don't see how that commit introduces the problem?

Regardless, this commit should fix it.

mvandeberg:
> I'm pretty sure this behavior's been in Steem since forever. Also, I don't see how that commit introduces the problem?

The commit is when we first set `is_producing` on `push_transaction` in reference to abitmore's concern over the precedence for this change.

PR #3138, opened 2018-11-01
Grab head_block_time and head_block_number in visitor constructor

mvandeberg:
In #2719 I had done a few refactors during development that got squashed in the PR. One change was only partially reverted. This adds back the correct behavior.


theoreticalbts:
Approved.  It might be a slight performance win (avoids a GPO lookup) to put this initialization where it originally was.  But it's not worth holding up fixing the CI issues caused by the huge amount of log messages caused by the bug this patch fixes.

Issue #2729, opened 2018-08-06
As a token creator, I want to be able to initiate my ICO.


youkaicountry:
Entails:
smt_setup_operation
Generation policies
Extended Fund Destinations
pre_contribution_pause

PR #3139, opened 2018-11-02
SMT setup emissions operation

### LINKED ISSUE #2726: smt_setup_inflation_operation
Create the `smt_setup_inflation_operation`.

sgerbino:
Closes #2726.


libraries/chain/util/smt_token.cpp
```diff
@@ -23,17 +23,43 @@ const smt_token_object* find_smt_token( database& db, uint32_t nai )
...
+
+   for ( const auto& r : boost::make_iterator_range( range ) )
+   {
+      time_point_sec schedule_end;
+
+      // A maximum interval count indicates we should emit indefinitely
+      if ( r.interval_count == std::numeric_limits< uint32_t >::max() )
+         return time_point_sec::maximum();
+      else
+         schedule_end = r.schedule_time + fc::seconds( r.interval_seconds * r.interval_count );
```
mvandeberg: interval_seconds * interval_count could overflow.

libraries/protocol/smt_operations.cpp
```diff
@@ -198,23 +198,37 @@ void smt_setup_emissions_operation::validate()const
...
+
+   // If time modulation is enabled
+   if ( lep_time != rep_time )
+   {
+      FC_ASSERT( lep_time >= schedule_time, "Left endpoint time cannot be before the schedule time" );
+
+      // If we don't emit indefinitely
+      if ( interval_count != std::numeric_limits<uint32_t>::max() )
+      {
+         FC_ASSERT( rep_time <= schedule_time + fc::seconds( interval_seconds * interval_count ), "Right endpoint time cannot be after the schedule end time" );
```
mvandeberg: Potential overflow

libraries/protocol/smt_operations.cpp
```diff
@@ -198,23 +198,38 @@ void smt_setup_emissions_operation::validate()const
...
+
+   // If time modulation is enabled
+   if ( lep_time != rep_time )
+   {
+      FC_ASSERT( lep_time >= schedule_time, "Left endpoint time cannot be before the schedule time" );
+
+      // If we don't emit indefinitely
+      if ( interval_count != SMT_EMIT_INDEFINITELY )
+      {
+         FC_ASSERT( rep_time <= schedule_time + fc::seconds( uint64_t( interval_seconds ) * uint64_t( interval_count ) ),
```
mvandeberg: You overflow checks don't solve the real problem that `fc::time_point_sec` is a 32 bit timestamp.

You need to ensure that `uint64_t( interval_seconds ) * uint64_t( interval_count ) + schedule_time.sec_since_epoch() <= std::numeric_limits< int32_t >::max` or else the time might wrap. We are pretty lose with using fc::time_point_sec vs fc::time_point_sec::sec_since_epoch and I am not convinced there isn't behavior that would fail in one way or work in the other. Ensuring the end time does not wrap is safest way to know that regardless of how the end time is calculated that it does not overflow.

libraries/protocol/smt_operations.cpp
```diff
@@ -198,23 +198,38 @@ void smt_setup_emissions_operation::validate()const
...
+   FC_ASSERT( symbol.is_vesting() == false, "Use liquid variant of SMT symbol to specify emission amounts" );
+   FC_ASSERT( symbol == lep_abs_amount.symbol, "Left endpoint symbol mismatch" );
+   FC_ASSERT( symbol == rep_abs_amount.symbol, "Right endpoint symbol mismatch" );
+   FC_ASSERT( lep_abs_amount.amount >= 0, "Left endpoint cannot have negative emission" );
+   FC_ASSERT( rep_abs_amount.amount >= 0, "Right endpoint cannot have negative emission" );
 
-   // lep_rel_amount_numerator <- any value of unsigned int is OK
-   // rep_rel_amount_numerator <- any value of unsigned int is OK
+   FC_ASSERT( lep_abs_amount.amount > 0 || lep_rel_amount_numerator > 0 || rep_abs_amount.amount > 0 || rep_rel_amount_numerator > 0,
+      "An emission operation must emit" );
```
mvandeberg: Potential reword, "An emission operation must have positive non-zero emission"

libraries/chain/util/smt_token.cpp
```diff
@@ -23,17 +23,39 @@ const smt_token_object* find_smt_token( database& db, uint32_t nai )
...
-      return find_smt_token( db, s.to_nai() );
+      return find_token( db, s.to_nai() );
    else
       return db.find< smt_token_object, by_symbol >( s );
 }
 
-} } } // steem::chain::util
+fc::optional< time_point_sec > last_emission_time( const database& db, const asset_symbol_type& symbol )
+{
+   const auto& idx = db.get_index< smt_token_emissions_index >().indices().get< by_symbol >();
```
mvandeberg: This can be shortened and made more readable by using the chainbase `get_index` templates.

`db.get_index< smt_token_emissions_index, by_symbol >()`

theoreticalbts:
I reviewed this myself, the only thing I found was the same bug in the review above:

> The evaluator should ensure that the emissions cannot start prior to accounced_launch_time.

Specifically, this check needs to be in `smt_setup_evaluator`, because that where `announced_launch_time` enters the system.

mvandeberg:
> Specifically, this check needs to be in smt_setup_evaluator, because that where announced_launch_time enters the system.

Agreed. This will be addressed in #2729

PR #3162, opened 2018-11-15
witness_plugin:  Implement multiple custom checking in post_apply_transaction #3161

### LINKED ISSUE #3161: Move custom op restriction from pre-operation to post-transaction evaluator
The witness plugin's restriction of custom ops makes it difficult to test custom ops (e.g. #3103 ).

Currently the restriction is applied pre-operation (due to a bug, although it is named as if it is a post-operation handler, it is actually registered as a pre-operation handler).

I'm going to move the check to post-transaction processing.

theoreticalbts:
Closes #3161.

### DIFF
--- a/libraries/plugins/witness/witness_plugin.cpp
+++ b/libraries/plugins/witness/witness_plugin.cpp
@@ -1,4 +1,5 @@
 #include <steem/plugins/witness/witness_plugin.hpp>
+#include <steem/plugins/witness/witness_plugin_objects.hpp>
 
 #include <steem/chain/database_exceptions.hpp>
 #include <steem/chain/account_object.hpp>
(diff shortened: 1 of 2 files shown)


theoreticalbts:
Updated PR, please re-review.

libraries/plugins/witness/witness_plugin.cpp
```diff
@@ -229,7 +228,6 @@ namespace detail {
 
    void witness_plugin_impl::on_pre_apply_block( const chain::block_notification& b )
```
mvandeberg: If pre_apply_block notification is not used, let's remove it.

theoreticalbts:
Removed.

PR #3158, opened 2018-11-14
Implement smt_set_runtime_parameters

mvandeberg:
Closes #2728
Closes #1410

WP Updates: steemit/smt-whitepaper#203

PR #3163, opened 2018-11-15
SMT set setup parameters operation

### LINKED ISSUE #2727: Smt_set_setup_parameters_operation
Create `smt_set_setup_parameters_operation`.

sgerbino:
Closes #2727.

White paper updated: https://github.com/steemit/smt-whitepaper/pull/207


libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
```diff
@@ -90,7 +90,6 @@ class smt_token_object : public object< smt_token_object_type, smt_token_object
 
    /// set_setup_parameters
    bool              allow_voting = false;
```
mvandeberg: Should this default value be true or false? In #3158 I tried to make all of the default values match STEEM or at least be a common scenario. I think `allow_voting` makes more sense being initialized to true.

PR #3167, opened 2018-11-17
Closes #3149 - Update Exchange Example Config File

### LINKED ISSUE #3149: Exchange sample config.ini is missing plugins
In the https://github.com/steemit/steem/blob/master/doc/example_config.ini file, which is linked to from exchangequickstart.md, the list of plugins on the plugin line is not sufficient for running an exchange node.

Update it to:
```
plugin = witness account_history_rocksdb account_history_api
```

TimCliff:
This PR replaces the exchange example config file with an updated version.
- It uses the default config file that is created from a fresh default steemd installation as the starting point.
- It sets the enabled plugins to: `witness account_history_rocksdb account_history_api`
- It adds an example `account-history-track-account-range`
- It enables `webserver-http-endpoint` and `webserver-ws-endpoint`

### DIFF
--- a/doc/example_config.ini
+++ b/doc/example_config.ini
@@ -108,7 +108,7 @@ market-history-buckets-per-size = 5760
 # seed-node =
 
 # The IP address and port of a remote peer to sync with.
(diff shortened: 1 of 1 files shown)


TimCliff:
changes made

PR #3185, opened 2018-11-29
SMT contribution operation

### LINKED ISSUE #2730: User participating in SMT ICO
As a user interested in a token, I want to be able to contribute to the ICO, so that the user owns interest in the SMT.

**Note:**
Entails:
smt_contribute_operation
contribution_phase

```c++
smt_contribute_operation
{
   account_name_type contributor;
   account_symbol_type symbol;
   uint32_t contribution_id; // (contributor, symbol, contribution_id) must be a unique tuple
   asset amount;

   extensions_type extensions;
};
```
**AC:**
1. This creates a contribution object that needs to track individual contributions on an SMT during the ICO phase. These contributions need to be sorted chronologically. `by_id` is sufficient.
2. Only STEEM can be contributed.
3. User must possess the STEEM to be contributed.
4. Only valid during ICO phase.
5. If the contribution would cause the total contributions to pass a revealed hard cap, the ICO phase is completed and the SMT is now waiting for launch.
6. Refactor ICO fields from `smt_token_object` in to a secondary token object that contains fields specific to ICO phase that can be removed after the ICO succeeds or fails.
7. When returning a token object in an API, add an optional subfield that contains the ICO specific object if the SMT is in the ICO phase.

**Tasks:**
- [x] Update `smt_token_object` API calls to return optional secondary ICO fields.
- [x] Validation/testing of `smt_contribute_operation`.
- [x] Evaluation/testing of `smt_contribute_operation`.
- [x] Refactor ICO fields into secondary object.
- [x] Add API calls to retrieve ICO Contribution Objects
- [ ] Update SMT Whitepaper

sgerbino:
Closes #2730.


libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
```diff
@@ -110,6 +110,21 @@ class smt_token_object : public object< smt_token_object_type, smt_token_object
 
    ///parameters for 'smt_setup_operation'
    int64_t                       max_supply = 0;
+};
+
+class smt_ico_object : public object< smt_ico_object_type, smt_ico_object >
```
mvandeberg: What is the difference between this object and the `smt_event_token_object`? They look like they contain, by and large, the same fields. Trying to keep that up to date in multiple places seems like a bad idea.

libraries/chain/smt_evaluator.cpp
```diff
@@ -300,5 +307,31 @@ void smt_set_runtime_parameters_evaluator::do_apply( const smt_set_runtime_param
...
+      FC_ASSERT( token->phase < smt_phase::contribution_end_time_completed, "SMT is no longer in the contribution phase" );
+      FC_ASSERT( _db.get_balance( o.contributor, o.contribution.symbol ) >= o.contribution, "Account does not have sufficient funds for contribution" );
+
+      _db.adjust_balance( o.contributor, -o.contribution );
+
+      _db.create< smt_contribution_object >( [&] ( smt_contribution_object& obj )
+      {
+         obj.contributor = o.contributor;
+         obj.symbol = o.symbol;
+         obj.contribution_id = o.contribution_id;
```
mvandeberg: There is no check for uniqueness of contribution id other than letting the creation of the object fail and a horrible exception thrown to the user.

libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
```diff
@@ -172,6 +187,42 @@ class smt_event_token_object : public object< smt_event_token_object_type, smt_e
...
+};
+
+struct by_symbol_contributor;
+
+typedef multi_index_container <
+   smt_contribution_object,
+   indexed_by <
+      ordered_unique< tag< by_id >,
+         member< smt_contribution_object, smt_contribution_object_id_type, &smt_contribution_object::id > >,
+      ordered_unique< tag< by_symbol_contributor >,
```
mvandeberg: Add a non-consensus order for a full node that sorts by `{contributor, symbol, constribution_id}` so that UIs can quickly show users all the ICOs they have contributed to. Also add this to order to the API.

libraries/protocol/smt_operations.cpp
```diff
@@ -382,5 +382,13 @@ void smt_set_setup_parameters_operation::validate() const
    FC_ASSERT( setup_parameters.empty() == false );
 }
 
+void smt_contribute_operation::validate() const
+{
+   validate_account_name( contributor );
+   common_symbol_validation( symbol );
+   FC_ASSERT( contribution.symbol == STEEM_SYMBOL );
```
mvandeberg: Add helpful error messages for these two assertions

tests/tests/smt_tests.cpp
```diff
@@ -680,6 +680,7 @@ BOOST_AUTO_TEST_CASE( setup_authorities )
    FC_LOG_AND_RETHROW()
 }
 
+#if 0
```
mvandeberg: What is now broken about this test?

tests/tests/smt_tests.cpp
```diff
@@ -826,36 +828,38 @@ void setup_smt_and_reveal_caps( const account_name_type& control_account, const
...
    // Reveal min cap.
    op.cap.amount = min_cap_val;
    PUSH_OP( op, private_key );
-   FC_ASSERT( smt_object.steem_units_min_cap == min_cap_val && smt_object.steem_units_hard_cap == max_cap_val );
+   FC_ASSERT( ico_object.steem_units_min_cap == min_cap_val && ico_object.steem_units_hard_cap == max_cap_val );
    // Try to reveal min cap again.
    FAIL_WITH_OP( op, private_key, fc::assert_exception );
 }
 
+#if 0
```
mvandeberg: Again, what is now broken about this test?

libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
```diff
@@ -206,6 +182,7 @@ class smt_contribution_object : public object< smt_contribution_object_type, smt
 };
 
 struct by_symbol_contributor;
+struct by_contributor;
 
 typedef multi_index_container <
```
mvandeberg: We need to do a little thinking ahead here. To complete the ICO we need to be able to iterate through contributions in the order in which they came in. The contribution id is important from a UI perspective, but doesn't actually give us what we need.

Thankfully, we can leverage the fact that IDs are monotonically increasing. So if we create a composite key on `(symbol,id)`, you will have an in order index of contributions for an SMT.

PR #3201, opened 2018-12-10
FC unpack fixes (master)

sgerbino:


abitmore:
Doesn't the removal of memory pre-allocation harm performance?

mvandeberg:
`resize` pre-allocates and initializes. `reserve` is the best, but still consumes memory. It appears that initialization cancels out any gains the pre-allocation may have.

In our testing, we reindexed the first 5 millions blocks and usually saw a slight increase in performance with this patch (10ish seconds out of 600ish), but it appears to not be statistically significant.

abitmore:
So it makes sense to keep the `reserve()` calls, and perhaps change the `resize()` calls to `reserve()` as well?

PR #3204, opened 2018-12-11
Unlinkable block p2p

### LINKED ISSUE #3180: 0.20.6 Unlinkable Block P2P Issue
Many witnesses have noticed recently a lot of seemingly random missed blocks with little explanation. 

I tracked the issue to an exception that wasn't being caught properly here: 

https://github.com/steemit/steem/blob/75300afff53d850f777f46dab116591d9a1764c5/libraries/plugins/p2p/p2p_plugin.cpp#L216

Instead of catching the correct exception, it throws a generic one. If, due to a latency event, you get block N+2 ahead of block N+1, your client _/should/_ go back and request N+1. However when this exception is caught in a generic way it instead disconnects from that peer, and then one by one cascades to kill all your peers, then proceeds to find new peers to sync. (See this code path: https://github.com/steemit/steem/blob/75300afff53d850f777f46dab116591d9a1764c5/libraries/net/node.cpp#L3573)
This resulting time delay re-syncing is the likely source of these random missed blocks.

For the life of me I can't figure out why the derived exception isn't being caught correctly, but as a 'hack' it seems you can compare the FC code and perform the right action.
You can find a gist of this patch here:
https://gist.github.com/ScottSallinen/7ec83d3fcc4edca650137df875326ced

sgerbino:
Resolves #3180.

### DIFF
--- a/libraries/plugins/p2p/p2p_plugin.cpp
+++ b/libraries/plugins/p2p/p2p_plugin.cpp
@@ -227,7 +227,12 @@ bool p2p_plugin_impl::handle_block( const graphene::net::block_message& blk_msg,
                ("e", e.to_detail_string())
                ("head", head_block_num));
          elog("Error when pushing block:\n${e}", ("e", e.to_detail_string()));
-         throw;
+         if (e.code() == 4080000) {
+           elog("Rethrowing as graphene::net exception");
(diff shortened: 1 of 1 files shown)


abitmore:
This hasn't been merged into `stable` branch?

PR #3209, opened 2018-12-14
Fix version.hpp for newer glibc by adding sysmacros

Someguy123:
When building on Ubuntu 18.04, errors are raised about the use of `minor` and `major` without including `sys/sysmacros.h`

This one line addition of `#include <sys/sysmacros.h>` fixes the steemd build on newer versions of glibc.


Someguy123:
It was observed when building v0.20.7, but as far as I could tell, this wasn't fixed on master.

It's a 1 line fix which allows it to build on 18.04 and anything else that uses a newer glibc.

PR #3263, opened 2019-03-19
Closes #3258 Fix update_median_feed when total sbd is 0

### LINKED ISSUE #3258: Blockchain stops when SBD supply is 0
This issue is related with the function `update_median_feed()` in `database.cpp`.

#### Expected behavior
If the SBD supply is 0 the blockchain takes the median feed price from witnesses. 

#### Actual behavior
When the SBD supply is 0 everyone stop producing blocks.

#### How to reproduce
- Start a ~~testnet~~ a new chain with 21 witnesses. 
- Set a feed price on at least 7 witnesses.
- Wait 1 hour for the update of the median feed price.
- All witnesses stop producing blocks with this error:

```
Failed to push new block:
base.amount > share_type(0): 
    {}
    asset.cpp:377 validate

    {"base":{"amount":"0","precision":3,"nai":"@@000000013"},"quote":{"amount":"250037129","precision":3,"nai":"@@000000021"}}
    asset.cpp:380 validate
```

```
Got exception while generating block:
base.amount > share_type(0): 
    {}
    asset.cpp:377 validate

    {"base":{"amount":"0","precision":3,"nai":"@@000000013"},"quote":{"amount":"250037129","precision":3,"nai":"@@000000021"}}
    asset.cpp:380 validate

    {}
    database.cpp:3128 update_median_feed
```

#### Reason
The `update_median_feed()` function updates the feed price taking into account two things: 
- The median price given by the witnesses, and 
- The haircut price to force SBD to remain below 10% of the virtual supply. 

However, if the SBD supply is zero, **the haircut price throws an exception because both quote and base have to be greater than zero** (to prevent division by 0). No witness can create the new block and therefore the **blockchain freezes**.

Line in question:
```
price min_price( asset( 9 * gpo.current_sbd_supply.amount, SBD_SYMBOL ), gpo.current_supply );
```

#### Low vulnerability
The unique way to reach 0 SBD in the mainnet is stopping printing SBD and converting all of them into STEEM, which is very unlikely. However, it becomes more relevant on testnets or private blockchains.

#### Solution
Only calculate the haircut price if the SBD supply is greater than zero, because in this case there is no risk of debt, the ratio is 0%.

joticajulian:
Closes #3258.
This issue can be solved with a simple IF, since if for some reason the SBD goes to zero, then there is no debt and there is no need to calculate the haircut price.
A hardfork is not needed.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -3266,10 +3266,14 @@ try {
                // so the combined market cap is $900 + $100 = $1000.
 
                const auto& gpo = get_dynamic_global_properties();
-               price min_price( asset( 9 * gpo.current_sbd_supply.amount, SBD_SYMBOL ), gpo.current_supply );
 
-               if( min_price > fho.current_median_history )
-                  fho.current_median_history = min_price;
+               if( gpo.current_sbd_supply.amount > 0 )
+               {
+                  price min_price( asset( 9 * gpo.current_sbd_supply.amount, SBD_SYMBOL ), gpo.current_supply );
+
+                  if( min_price > fho.current_median_history )
+                     fho.current_median_history = min_price;
+               }
             }
          }
       });


joticajulian:
Also closes #3268

Issue #3270, opened 2019-03-31
Assert Exception:is_valid_account_name( name ): Account name null is invalid


nnnarvaez:
When trying to transfer funds to `null` account from steem-python or cli_wallet the node replies:

from python:
```
assert_exception from 127.0.0.1 (Assert Exception:is_valid_account_name( name ): Account name null is invalid) in broadcast_transaction
```

from cli_wallet (`null` with no quotes)
```
10 assert_exception: Assert Exception
is_valid_account_name( name ): Account name  is invalid
    {"n":""}
    validation.hpp:19 validate_account_name

    {"*this":{"from":"initdreamer","to":"","amount":{"amount":"10000000","precision":3,"nai":"@@000000021"},"memo":"no Memo"}}
    steem_operations.cpp:191 validate

    {"from":"initdreamer","to":"","amount":"10000.000 DREAM","memo":"no Memo","broadcast":true}
    wallet.cpp:1864 transfer
``` 

from cli_wallet (`"null"` with quotes) 
```
10 assert_exception: Assert Exception
is_valid_account_name( name ): Account name null is invalid
    {"n":"null"}
    validation.hpp:19 validate_account_name

    {"*this":{"from":"initdreamer","to":"null","amount":{"amount":"10000000","pr                                                 ecision":3,"nai":"@@000000021"},"memo":"no Memo"}}
    steem_operations.cpp:191 validate

    {"from":"initdreamer","to":"null","amount":"10000.000 DREAM","memo":"no Memo                                                 ","broadcast":true}
    wallet.cpp:1864 transfer
unlocked >>> transfer initdreamer null "10000.000 DREAM" "no Memo" true
transfer initdreamer null "10000.000 DREAM" "no Memo" true

```

mvandeberg:
The null account is successfully receiving transfer on the Steem network and I double checked the validity of the null account in a unit test. This all seems to be working.

I also tested on a cli-wallet built for our testnet.

`unlocked >>> transfer "initminer" "null" "1000.000 TESTS" "" false`

```
{
  "ref_block_num": 30814,
  "ref_block_prefix": 3853958081,
  "expiration": "2019-04-01T15:35:21",
  "operations": [[
      "transfer",{
        "from": "initminer",
        "to": "null",
        "amount": "1000.000 TESTS",
        "memo": ""
      }
    ]
  ],
  "extensions": [],
  "signatures": [],
[...]

PR #3271, opened 2019-04-03
added note to document #3270 behavior on `null` transfers

nnnarvaez:
Assert Exception:is_valid_account_name( name ): Account name null is invalid #3270


mvandeberg:
Changing any of the config values can, and probably will, change the consensus rules of the Steem blockchain, which may lead to "buggy" behavior. I do not believe this needs to be documented. If you want to fork Steem and make modifications on your own, you have the freedom to do so, but we do not need to support you in that endeavor.

nnnarvaez:
Thanks for your answer.
I guess if in the future the steem block chain decides to change that setting this discussion will be useful to troubleshoot why transfers to null aren't working.

I am not forking steembi am running a live chain because the testnet has its own limitations and I don't like testing stuff in a production environment.

You don't need to be rude but you are free to be

PR #3272, opened 2019-04-16
Implemented Steem Proposal System and also C++/python based tests suite

vogel76:


joticajulian:
I was wondering if it is possible to get control of steem.dao after the hardfork using the recovery account operation. Maybe it is better to remove the owner history from the `owner_authority_history_index` when processing the hardfork.

mvandeberg:
Changing the owner auth in the hardfork bypasses the old auth from being added to the `owner_authority_history_index`. Blocktrades is the current custodian and so long as he does not change the owner auth between now and the hardfork, it would not be necessary. I trust that Blocktrades will not change the keys, but if he were hacked some how and the key was stolen, it could throw the efficacy of the hardfork in question and would require another hardfork to fix it.

We can quickly implement this in #3276

dnotestein:
I was just planning to change the keys to null and set the recovery account to null via the cli before the release (most likely tomorrow), but I'm fine if you guys want to code the change instead.

joticajulian:
@dnotestein it is better if you do not change the keys. Just leave it like that.
Because by changing the keys the blockchain adds a new record in the `owner_authority_history_index` and opens the door to the recovery.

dnotestein:
> @dnotestein it is better if you do not change the keys. Just leave it like that.
> Because by changing the keys the blockchain adds a new record in the `owner_authority_history_index` and opens the door to the recovery.

No, not if the recovery account is first set to null. It's not possible to do a recovery after this.

dnotestein:
Someone told me that setting to null still allows recovery via top #1 witness. So instead, we can set the recovery agent to steem.dao, then set the keys to null. This should solve the issue, since steem.dao won't be able to recover itself since it can't sign a transaction after key is set to null.

mvandeberg:
That is correct, but you can only recover using an owner auth that was active within the last 30 days. If there are none, then the account cannot be recovered regardless of what recovery account is specified.

The recovery partner is also only changed after 30 days.

Regarding setting the recovery partner to null, if you set it to the null account, then the account is not recoverable. If you set it to the empty string then it is recoverable by the top witness.

I still want to enforce this account state in code as a final stop gap. It 100% prevents any sort of unforeseen circumstance from causing problems in the fork for minimal effort.

mvandeberg:
@vogel76 Can we get a PR based on `master`? I believe that was the original base branch for development.

dnotestein:
It was, but subsequent changes to master broke our ability to compile the software, so they based it off something that compiled for us.

libraries/chain/sps_evaluator.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+using steem::chain::create_proposal_evaluator;
+
+void create_proposal_evaluator::do_apply( const create_proposal_operation& o )
+{
+   try
+   {
+      FC_ASSERT( _db.has_hardfork( STEEM_PROPOSALS_HARDFORK ), "Proposals functionality not enabled until hardfork ${hf}", ("hf", STEEM_PROPOSALS_HARDFORK) );
+
+      ilog("creating proposal: ${op}", ("op", o));
```
mvandeberg: Logging should be removed.

libraries/chain/sps_evaluator.cpp
```diff
@@ -0,0 +1,157 @@
...
+   FC_CAPTURE_AND_RETHROW( (o) )
+}
+
+void update_proposal_votes_evaluator::do_apply( const update_proposal_votes_operation& o )
+{
+   try
+   {
+      FC_ASSERT( _db.has_hardfork( STEEM_PROPOSALS_HARDFORK ), "Proposals functionality not enabled until hardfork ${hf}", ("hf", STEEM_PROPOSALS_HARDFORK) );
+
+      ilog("voting proposal: ${op}", ("op", o));
```
mvandeberg: Logging should be removed.

libraries/chain/sps_evaluator.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+      if( o.proposal_ids.empty() )
```
mvandeberg: This is not needed. In `update_proposal_votes_operation::validate()` `proposal_ids.empty()` is asserted false.

libraries/chain/sps_evaluator.cpp
```diff
@@ -0,0 +1,157 @@
...
+   FC_CAPTURE_AND_RETHROW( (o) )
+}
+
+void remove_proposal_evaluator::do_apply(const remove_proposal_operation& op)
+{
+   try
+   {
+      FC_ASSERT( _db.has_hardfork( STEEM_PROPOSALS_HARDFORK ), "Proposals functionality not enabled until hardfork ${hf}", ("hf", STEEM_PROPOSALS_HARDFORK) );
+
+      ilog("Attempting to evaluate remove_proposal_operation: ${o}", ("o", op));
```
mvandeberg: Remove logging

libraries/chain/sps_objects/sps_processor.cpp
```diff
@@ -0,0 +1,286 @@
...
+   auto found = pvidx.find( id );
+
+   while( found != pvidx.end() && found->proposal_id == id )
+   {
+      const auto& _voter = db.get_account( found->voter );
+
+      //If _voter has set proxy, then his votes aren't taken into consideration
+      if( _voter.proxy == STEEM_PROXY_TO_SELF_ACCOUNT )
+      {
+         auto sum = std::accumulate(   _voter.proxied_vsf_votes.begin(),
```
mvandeberg: Should probably just call `_voter.proxied_vsf_votes_total()` to avoid repeated code.

libraries/chain/sps_objects/sps_processor.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+   return asset( daily_budget_limit, treasury_fund.symbol );
+}
+
+void sps_processor::transfer_daily_inflation_to_treasury( const asset& daily_inflation )
+{
+   if( daily_inflation.amount.value > 0 )
+   {
+      const auto& treasury_account = db.get_account( STEEM_TREASURY_ACCOUNT );
+      db.adjust_balance( treasury_account, daily_inflation );
```
mvandeberg: Is this printing new SBD?

That's not how we have handled inflation in the past. Either inflation should be paid, per block, in to the treasury account in `database::process_funds` or added to a temp fund in the dgpo that is then transferred in to the treasury account during maintenance.

libraries/chain/sps_objects/sps_processor.cpp
```diff
@@ -0,0 +1,286 @@
...
+{
+   auto& treasury_account = db.get_account( STEEM_TREASURY_ACCOUNT );
+
+   return treasury_account.sbd_balance;
+}
+
+asset sps_processor::get_daily_inflation()
+{
+   //temporary
+   return asset( 0, SBD_SYMBOL );
```
mvandeberg: We should have the proper logic here, even though the inflation rate is 0 so that when it is changed later, this method does not get overlooked. At a minimum, this needs an FCTODO as a reminder.

libraries/chain/sps_objects/sps_processor.cpp
```diff
@@ -0,0 +1,286 @@
...
+   //Get daily proposal inflation ( daily_proposal_inflation )
+   asset daily_inflation = get_daily_inflation();
+
+   FC_ASSERT( treasury_fund.symbol == daily_inflation.symbol, "symbols must be the same" );
+
+   //Calculate budget for given maintenance period
+   auto passed_time_seconds = ( head_time - db.get_dynamic_global_properties().last_budget_time ).to_seconds();
+
+   //Calculate daily_budget_limit
+   auto daily_budget_limit = treasury_fund.amount.value / 100 + daily_inflation.amount.value;
```
mvandeberg: 100 should be replaced with some sort of a constant. At first glance it appears that we are dealing with a percentage, but that is not the case.

libraries/chain/sps_objects/sps_processor.cpp
```diff
@@ -0,0 +1,286 @@
...
+   daily_budget_limit = daily_budget_limit * ( static_cast< double >( passed_time_seconds ) / daily_seconds );
```
mvandeberg: casting to a double is a *very* bad idea. Floating point math is notoriously non-deterministic which could lead to a fork in the chain. Please replace all numerical declarations of `auto` with the correct integer type to ensure we do not release non-deterministic code.

mvandeberg: To get this as a fraction, we use `STEEM_1_PERCENT` and `STEEM_100_PERCENT` constants.

This should be rewritten as:

```
daily_budget_limit = ( daily_budget_limit * passed_time_seconds ) / ( daily_seconds * STEEM_100_PERCENT );
```

Developer analysis needs to be done to ensure that 64 bit fields cannot overflow in all cases. To be safe, it should probably be cast to fc's 128 bit integer type because assets can be up to 64 bits and a timestamp can be up to 32 bits. The denominator needs to be cast to a 64 bit integer to accommodate the multiplications between 32 bit and 16 bit integers.

libraries/chain/sps_objects/sps_processor.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+void sps_processor::transfer_payments( const time_point_sec& head_time, asset& maintenance_budget_limit, const t_proposals& proposals )
+{
+   if( maintenance_budget_limit.amount.value == 0 )
+      return;
+
+   const auto& treasury_account = db.get_account(STEEM_TREASURY_ACCOUNT);
+
+   auto passed_time_seconds = ( head_time - db.get_dynamic_global_properties().last_budget_time ).to_seconds();
+   double ratio = static_cast< double >( passed_time_seconds ) / daily_seconds;
```
mvandeberg: Again, do not use doubles.

This should be:

```
uint64_t ratio_percent = ( passed_time_seconds * STEEM_100_PERCENT ) / daily_seconds;
```

libraries/chain/sps_objects/sps_processor.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+   for( auto& item : proposals )
+   {
+      const proposal_object& _item = item;
+
+      //Proposals without any votes shouldn't be treated as active
+      if( _item.total_votes == 0 )
+         break;
+
+      auto period_pay = asset( ratio * _item.daily_pay.amount.value, _item.daily_pay.symbol );
```
mvandeberg: Then this becomes

```
asset period_pay = asset( ( ratio * _item.daily_pay.amount.value ) / STEEM_100_PERCENT, _item.daily_pay.symbol );
```

libraries/protocol/include/steem/protocol/sps_operations.hpp
```diff
@@ -0,0 +1,129 @@
...
+   proposal_pay_operation(const account_name_type r, const asset& p, transaction_id_type _trx_id, uint16_t _op_in_trx )
+                     : receiver(r), payment(p), trx_id(_trx_id), op_in_trx(_op_in_trx) {}
+
+   /// Name of the account which is paid for
+   account_name_type receiver;
+   /// Amount of SBDs paid.
+   asset             payment;
+
+   /// Transaction id + position of operation where appeared a proposal being a source of given operation.
+   transaction_id_type trx_id;
```
mvandeberg: These virtual ops don't belong to transactions. In fact, they always happen during per block processing, which is external to transactions. Besides, this information is already handled by the account history plugin.

vogel76:
Most of requested changes are done here:
https://github.com/steemit/steem/pull/3278
Also new PR is made against current version of master.

libraries/protocol/include/steem/protocol/sps_operations.hpp
vogel76: We have added `Transaction id + position of operation` fields to point operation which created proposal related to given payment. This way AH can always hold information pointing to data describing proposal. Holding just proposal ID  is insufficient, since `expired` proposals are removed from state, so AH could hold dangling reference.

mvandeberg:
Closing in favor of #3278

PR #3285, opened 2019-05-01
Update MIRA configuration

### LINKED ISSUE #3284: Update MIRA configuration for sustainable memory growth

sgerbino:
Resolves #3284.


libraries/plugins/chain/chain_plugin.cpp
```diff
@@ -476,9 +476,9 @@ void chain_plugin::plugin_startup()
    {
       database_config = fc::json::from_file( my->database_cfg, fc::json::strict_parser );
    }
-   catch ( const std::exception& e )
+   catch ( ... )
```
mvandeberg: Let's add catching an fc::assert_exception so that

1) We can log it
2) We don't suppress unexpected exception types (i.e. boost exceptions) as a database configuration error.

PR #3294, opened 2019-05-08
Update account metadata via posting key

### LINKED ISSUE #3274: SIP - Update Account Metadata via Posting Key
An account's JSON metadata is currently updated via the owner or active authority. The field is non-consensus and is used for many applications that do not need/want access to an account's active authority. To require the active authority for updating such information is a security weakness as it requires users to sign with an active key when they would otherwise not need to.

## New Operation

``` C++
struct account_update2_operation : public base_operation
{
   account_name_type             account;
   optional< authority >         owner;
   optional< authority >         active;
   optional< authority >         posting;
   optional< public_key_type >   memo_key;
   string                        json_metadata;
   string                        posting_json_metadata;

   extensions_type               extensions;
};
```

The operation would behave identically to the existing `account_update_operation` except that it would allow the posting authority to sign when the owner and active authorities and memo key are not present (i.e. only valid for posting and JSON metadata).

Prior to HF 21, `json_metadata` and `posting_json_metadata` will be updated at the same time by the previous operation. After HF 21, the old account update will only change `json_metadata`, and `posting_json_metadata` can only be updated by this operation. This way any App that is relying on security backed by the active authority can and will continue to operate after the hardfork without needing to update.

## Alternate Solutions

The biggest drawback here is the creation of a new (and mostly redundant) operation. There are two alternative solutions. Both involve using existing portions to the account update operation to signal old use of the operation and remove the soft fork during a hard fork. This would allow us to get around requiring a new operation, but would require apps to update their use of this operation to continue operating. Creating a new operation would allow for perfect backwards compatibility with existing apps as well as adding the extraordinarily useful extensions field.

sgerbino:
Resolves #3274.

### DIFF
--- a/libraries/chain/steem_evaluator.cpp
+++ b/libraries/chain/steem_evaluator.cpp
@@ -565,6 +565,10 @@ void account_update_evaluator::do_apply( const account_update_operation& o )
       _db.modify( _db.get< account_metadata_object, by_account >( account.id ), [&]( account_metadata_object& meta )
       {
          from_string( meta.json_metadata, o.json_metadata );
+         if ( !_db.has_hardfork( STEEM_HARDFORK_0_21__3274 ) )
+         {
+            from_string( meta.posting_json_metadata, o.json_metadata );
+         }
       });
    }
    #endif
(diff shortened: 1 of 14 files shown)


libraries/chain/steem_evaluator.cpp
```diff
@@ -580,6 +584,70 @@ void account_update_evaluator::do_apply( const account_update_operation& o )
 
 }
 
+void account_update2_evaluator::do_apply( const account_update2_operation& o )
+{
```
mvandeberg: The beginning of the evaluator needs a hardfork check so the new op is not accepted prior to the hardfork time.

tests/tests/operation_tests.cpp
```diff
@@ -7607,5 +7607,291 @@ BOOST_AUTO_TEST_CASE( account_auth_tests )
...
+      op1.validate();
+
+      validate_database();
+   }
+   FC_LOG_AND_RETHROW()
+}
+
+BOOST_AUTO_TEST_CASE( account_update2_authorities )
+{
+   try
```
mvandeberg: Auth tests are a mixed bag. I wrote most of them during the initial development of Steem while I was still learning about blockchain. Look at a newer operation test, such as `claim_account_authorities` for a better example on how these tests should be structured to be as clear as possible and not test redundant scenarios such as duplicate keys.

PR #3295, opened 2019-05-08
Remove magic numbers

### LINKED ISSUE #3277: Incorrect value in max auth membership error message
image

Upon further investigation, the error message in steemd is wrong.

https://github.com/steemit/steem/blob/4dd53f780aff99eb48e191169be245265986069a/libraries/protocol/steem_operations.cpp#L14

There are a few other "magic numbers" in op error messages that should be changed to config constants to avoid such problems in the future.

sgerbino:
Resolves #3277.


libraries/chain/steem_evaluator.cpp
```diff
@@ -2027,11 +2027,13 @@ void custom_evaluator::do_apply( const custom_operation& o )
 {
    database& d = db();
    if( d.is_producing() )
-      FC_ASSERT( o.data.size() <= 8192, "custom_operation must be less than 8k" );
+      FC_ASSERT( o.data.size() <= STEEM_CUSTOM_OP_DATA_MAX_LENGTH,
+         "custom_operation must be less than ${bytes}", ("bytes", STEEM_CUSTOM_OP_DATA_MAX_LENGTH) );
```
mvandeberg: provide unit of measurement in message.

libraries/chain/steem_evaluator.cpp
```diff
@@ -2040,12 +2042,14 @@ void custom_json_evaluator::do_apply( const custom_json_operation& o )
    database& d = db();
 
    if( d.is_producing() )
-      FC_ASSERT( o.json.length() <= 8192, "custom_json_operation json must be less than 8k" );
+      FC_ASSERT( o.json.length() <= STEEM_CUSTOM_OP_DATA_MAX_LENGTH,
+         "custom_json_operation json must be less than ${bytes}", ("bytes", STEEM_CUSTOM_OP_DATA_MAX_LENGTH) );
```
mvandeberg: Unit of measurement

libraries/protocol/steem_operations.cpp
```diff
@@ -69,7 +70,7 @@ namespace steem { namespace protocol {
 
    void comment_operation::validate() const
    {
-      FC_ASSERT( title.size() < 256, "Title larger than size limit" );
+      FC_ASSERT( title.size() < STEEM_COMMENT_TITLE_LIMIT, "Title larger than size limit" );
```
mvandeberg: It wasn't there previously, but we should include max size in error message.

libraries/protocol/steem_operations.cpp
```diff
@@ -327,14 +329,14 @@ namespace steem { namespace protocol {
    void custom_json_operation::validate() const {
       /// required auth accounts are the ones whose bandwidth is consumed
       FC_ASSERT( (required_auths.size() + required_posting_auths.size()) > 0, "at least one account must be specified" );
-      FC_ASSERT( id.size() <= 32, "id is too long" );
+      FC_ASSERT( id.size() <= STEEM_CUSTOM_OP_ID_MAX_LENGTH, "id is too long" );
```
mvandeberg: Include max value in error message.

libraries/protocol/steem_operations.cpp
```diff
@@ -327,14 +329,14 @@ namespace steem { namespace protocol {
...
       FC_ASSERT( fc::is_utf8(json), "JSON Metadata not formatted in UTF8" );
       FC_ASSERT( fc::json::is_valid(json), "JSON Metadata not valid JSON" );
    }
    void custom_binary_operation::validate() const {
       /// required auth accounts are the ones whose bandwidth is consumed
       FC_ASSERT( (required_owner_auths.size() + required_active_auths.size() + required_posting_auths.size()) > 0, "at least one account must be specified" );
-      FC_ASSERT( id.size() <= 32, "id is too long" );
+      FC_ASSERT( id.size() <= STEEM_CUSTOM_OP_ID_MAX_LENGTH, "id is too long" );
```
mvandeberg: Include max value in error message.

Issue #2545, opened 2018-06-13
`vote2_operation` - specify rshares instead of percent weight


roadscape:
A downside of specifying only percent `weight` in a `vote` op is that it's not possible to calculate `rshares` outside of steemd (you would have to know the exact VP and VESTS balance at the time of the vote). Vote `rshares` data is only available via steemd API and it only tracks latest state.

Since `rshares` has a lot of utility it would be ideal if we could eventually migrate to a sort of `vote2` operation which has you specify `rshares` instead of `weight`. This would allow outside apps easy and direct access to the raw value.

Steemd *could* maintain the current vote power logic and simply enforce that the specified rshares don't exceed what a 100% vote would be able to generate at that moment. Or this restriction could be loosened to give users more flexibility over how they distribute their influence.

Likewise, frontends could implement the same vote power logic to submit equivalent ops without changing UX. Or they could easily implement constant-weight voting.

mvandeberg:
This is related to the stateless discussion we were having this week.

mvandeberg:
Will be implemented in #2748

PR #3298, opened 2019-05-09
Allow votes when voting mana is zero

### LINKED ISSUE #3004: accounts with 0 voting mana can't vote

sgerbino:
Resolves #3004.


roadscape:
I left a comment on #3004, not sure if anyone saw it:

> Should "advisory"/no-op votes be allowed at all? Given that SP delegation seems to be a core concern for UIs, faucets, and communities for the foreseeable future, how many legit users would ever have 0SP?
> 
> Other than it seems strange to allow no-op reward distribution core consensus actions, one practical concern is that there is no way for external services to efficiently filter spammy no-op votes (until we have something like https://github.com/steemit/steem/issues/2545)
> 
> If the goal is to allow 0SP users to interact with content, there are alternative solutions:
>  - off-chain, up to each UI
>  - on-chain in the form of some custom_json-based "advisory vote" standard

mvandeberg:
What is the difference between a 1 rshare vote and a 0 rshare vote for UX?

We used to have the 0.02 SBD rule. But again, what is the difference between 0.02 SBD and 0.019 SBD for UX?

#2545 is not a solution to a display problem. Comment votes are already returned with their rshare value. From a UI perspective, this change now allows votes with 0 rshares to be returned via APIs. Votes with 1 rshares can already be returned. This change will not increase the burden on UIs.

roadscape:
Ok, I guess it boils down to: accounts with `0 VESTS` will now be able to vote until their RC is drained. That's roughly 3 votes per day per 1 SP of RC "adjustment". Not significant I suppose.

PR #3299, opened 2019-05-09
String comparison for parent and permlink use AND logic instead of OR logic

### LINKED ISSUE #2203: implementation of strcmp_equal looks wrong
I think the `||` should be changed to `&&`

https://github.com/steemit/steem/blob/42e2d95ec09d1695ec1b392d47a2e44612815cf0/libraries/chain/steem_evaluator.cpp#L69

sgerbino:
Resolves #2203.


libraries/chain/steem_evaluator.cpp
```diff
@@ -944,7 +936,12 @@ void comment_evaluator::do_apply( const comment_operation& o )
       {
          com.last_update   = _db.head_block_time();
          com.active        = com.last_update;
-         strcmp_equal equal;
+         std::function< bool( const shared_string& a, const string& b ) > equal;
+
```
mvandeberg: Add an FC_TODO to test if things aren't broken after HF 21 so we can simplify this.

PR #3303, opened 2019-05-10
Remove *.cmake files from the .gitignore, remove RocksDB Makefile

### LINKED ISSUE #3302: Remove RocksDB Makefile and cleanup .gitignore

sgerbino:
Resolves #3302.


.gitignore
```diff
@@ -2,7 +2,6 @@
 *.sw*
 
 .vscode
-*.cmake
```
mvandeberg: CMake creates files named `cmake_install.cmake` that we don't want included. Perhaps change the ignore to that file specifically?

PR #3311, opened 2019-05-14
Added index_delegate functionality to access indices by name during runtime

### LINKED ISSUE #3307: Add configuration option to place some or all indices in memory during reindex

sgerbino:
Resolves #3307.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -185,20 +185,22 @@ void database::open( const open_args& args )
 #ifdef ENABLE_MIRA
 void reindex_set_index_helper( database& db, mira::index_type type, const boost::filesystem::path& p, const boost::any& cfg )
 {
+   std::string type_str = type == mira::index_type::mira ? "mira" : "bmic";
    for ( auto const& delegate : db.index_delegates() )
    {
-      ilog( "Setting index '${name}'.", ("name", delegate.first) );
+      ilog( "Converting index '${name}' to ${type} type.", ("name", delegate.first)("type", type_str) );
       delegate.second.set_index_type( db, type, p, cfg );
    }
 }
 
 void reindex_set_index_helper( database& db, mira::index_type type, const boost::filesystem::path& p, const boost::any& cfg, std::vector< std::string > indices )
 {
+   std::string type_str = type == mira::index_type::mira ? "mira" : "bmic";
    for ( auto& index_name : indices )
    {
       if ( db.has_index_delegate( index_name ) )
       {
-         ilog( "Setting index '${name}'.", ("name", index_name) );
+         ilog( "Converting index '${name}' to ${type} type.", ("name", index_name)("type", type_str) );
          db.get_index_delegate( index_name ).set_index_type( db, type, p, cfg );
       }
       else
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -183,30 +183,30 @@ void database::open( const open_args& args )
 }
 
 #ifdef ENABLE_MIRA
-void reindex_set_index_helper( database& db, mira::index_type type, const boost::filesystem::path& p, const boost::any& cfg )
+void reindex_set_index_helper( database& db, mira::index_type type, const boost::filesystem::path& p, const boost::any& cfg, std::vector< std::string > indices )
 {
+   index_delegate_map delegates;
+
+   if ( indices.size() > 0 )
+   {
+      for ( auto& index_name : indices )
+      {
+         if ( db.has_index_delegate( index_name ) )
+            delegates[ index_name ] = db.get_index_delegate( index_name );
+         else
+            wlog( "Encountered an unknown index name '${name}'.", ("name", index_name) );
+      }
+   }
+   else
+   {
+      delegates = db.index_delegates();
+   }
+
    std::string type_str = type == mira::index_type::mira ? "mira" : "bmic";
-   for ( auto const& delegate : db.index_delegates() )
+   for ( auto const& delegate : delegates )
    {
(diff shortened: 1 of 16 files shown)


libraries/chain/database.cpp
```diff
@@ -244,7 +236,11 @@ uint32_t database::reindex( const open_args& args )
       if( args.replay_in_memory )
       {
          ilog( "Configuring replay to use memory..." );
-         reindex_set_index_helper( *this, mira::index_type::bmic, args.shared_mem_dir, args.database_cfg );
+
+         if ( args.replay_memory_indices.size() > 0 )
```
mvandeberg: Maybe this branching logic should be moved to `reindex_set_index_helper`?

libraries/chain/include/steem/chain/database.hpp
```diff
@@ -33,6 +34,15 @@ namespace steem { namespace chain {
    using steem::protocol::price;
    using abstract_plugin = appbase::abstract_plugin;
 
+   class database;
+   using set_index_type_func = std::function< void(database&, mira::index_type, const boost::filesystem::path&, const boost::any&) >;
+
+   struct index_delegate {
+      set_index_type_func set_index_type;
```
mvandeberg: Why not simply pass around `std::function`s instead of a struct containing the `std::function`? This would allow (if wanted) for the index delegates to be lambdas.

libraries/chain/include/steem/chain/index.hpp
```diff
@@ -58,3 +58,23 @@ void add_plugin_index( database& db )
...
+
+#define STEEM_ADD_CORE_INDEX(db, index_name)                                                                 \
+   do {                                                                                                      \
+      steem::chain::add_core_index< index_name >( db );                                                      \
+      steem::chain::index_delegate delegate;                                                                 \
+      delegate.set_index_type =                                                                              \
+         []( database& _db, mira::index_type type, const boost::filesystem::path& p, const boost::any& cfg ) \
+            { _db.get_mutable_index< index_name >().mutable_indices().set_index_type( type, p, cfg ); };     \
+      db.set_index_delegate( #index_name, std::move( delegate ) );                                           \
+   } while( false );
```
mvandeberg: Why is this a do/while? Because of the repeated symbol name `delegate`?

First of all, you can encapsulate any arbitrary code block in braces without needed the do/while. Second, passing around `std::function` directly as lambdas allows you do set the index delegate without needing any in scope declarations and cleans up this macro a fair bit.

libraries/plugins/chain/chain_plugin.cpp
```diff
@@ -327,6 +328,9 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
          ("checkpoint,c", bpo::value<vector<string>>()->composing(), "Pairs of [BLOCK_NUM,BLOCK_ID] that should be enforced as checkpoints.")
          ("flush-state-interval", bpo::value<uint32_t>(),
             "flush shared memory changes to disk every N blocks")
+#ifdef ENABLE_MIRA
+         ("indices-memory-replay", bpo::value<vector<string>>()->multitoken()->composing(), "Specify which indices should be in memory during replay")
```
mvandeberg: If this option exists only in the config file (which I agree it should) then you can prefix it with `memory`. I think `memory-replay-indices` sounds a bit better. A little less "yoda"-ish.

libraries/chain/include/steem/chain/database.hpp
sgerbino: The purpose of this was extensibility, if you needed any functionality by index name you could add a function to the struct and modify the macro and every index would have the new feature.

libraries/chain/include/steem/chain/index.hpp
sgerbino: The purpose of the do/while is to ensure correct behavior in conditionals. Consider the following:
```
#include <iostream>
#include <string>

#define MY_MACRO(x)                 \
    do {                            \
        std::cout << "multi\n";     \
        std::cout << "statement\n"; \
        std::cout << x << std::endl;\
    } while ( false )

#define MY_OTHER_MACRO(x)           \
    {                               \
        std::cout << "multi\n";     \
        std::cout << "statement\n"; \
        std::cout << x << std::endl;\
    }

int main()
{
  bool condition = true;
  
  if ( condition )
    MY_MACRO("test");
  else
    std::cout << "fail" << std::endl;

  return 0;
}
```

In this case `MY_MACRO` would behave as expected and `MY_OTHER_MACRO` would cause a compilation error, or worse yet erroneously cause the end of the if statement in other cases.

However, I did accidentally add `;` to the macro, and this should be removed to preserve this behavior in my opinion.

libraries/plugins/chain/chain_plugin.cpp
sgerbino: I believe appbase will copy any configuration options to command line options. As it stands this option is available in both as per `steemd -h` and local testing. I'm not aware of a way to avoid this, please advise.

mvandeberg: You are correct. Config options are copied to command line arguments but not the other way.

Go ahead and name it `memory-replay-indices` and change `memory-replay` to `memory-replay,m`. I believe that will force the shorthand as `-m`, which is arguably no more arbitrary than `--memory` compared to `--memory-replay`.

libraries/chain/include/steem/chain/index.hpp
mvandeberg: I think this is a bit overkill because of the role of this macro being something internal.

I agree that if the macro were being used in a library, then creating it in such a manner would be preferred. As it stands, it is not incorrect.

PR #3329, opened 2019-05-23
Implementation of convergent linear rewards curve

### LINKED ISSUE #3325: Implement Convergent Linear Rewards Curve
Extend the reward logic in `reward.hpp` and `reward.cpp` with the new curve `n^2/(n/C + K)`.

sgerbino:
Resolves #3325.

### DIFF
--- a/libraries/chain/util/reward.cpp
+++ b/libraries/chain/util/reward.cpp
@@ -65,7 +65,7 @@ uint64_t get_rshare_reward( const comment_reward_context& ctx )
    } FC_CAPTURE_AND_RETHROW( (ctx) )
 }
 
-uint128_t evaluate_reward_curve( const uint128_t& rshares, const protocol::curve_id& curve, const uint128_t& var1, const uint128_t& var2 )
+uint128_t evaluate_reward_curve( const uint128_t& rshares, const protocol::curve_id& curve, const uint128_t& var1 )
 {
    uint128_t result = 0;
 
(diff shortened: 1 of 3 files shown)


eonwarped:
Out of curiosity, how come you need two constants? Is the reward calculation not independent of multiplicative scaling of the curve?

libraries/chain/util/reward.cpp
```diff
@@ -89,6 +91,18 @@ uint128_t evaluate_reward_curve( const uint128_t& rshares, const protocol::curve
       case protocol::square_root:
          result = approx_sqrt( rshares );
          break;
+      case protocol::convergent_linear:
+         {
+            const uint128_t& s = var1;
+            result = ( rshares + s ) * ( rshares + s ) - s * s / ( rshares + 4 * s );
```
mvandeberg: Check your order of operations. This is (r+2)^2 - ( s^2 / (r + 4s) )

mvandeberg:
@eonwarped The more generic form of this equation is `((r + a)^2 - a^2)/(br + y)`. The curation curve is derived from this and core principles that govern how curation should work. Based on the curation curve, these values are then fixed.

From this generic curve, you can see that our linear curve is the special case when b=1, y=0. Our quadratic curve is the special case when b=0.

Due to the restrictions put in place during the derivation of the curation curve, `a`, `b,` and `y` all become related, and thus we can simplify the equation to a single variable,`s`.

A mathematically complete derivation of these curves is coming.

PR #3330, opened 2019-05-24
Parallelize Tests in CI

mvandeberg:
Includes a few changes to build and run tests without MIRA


libraries/chainbase/src/chainbase.cpp
```diff
@@ -165,6 +165,7 @@ namespace chainbase {
 
    void database::wipe( const bfs::path& dir )
    {
+      std::cout << "WIPE!!!" << std::endl;
```
sgerbino: Probably meant to remove this.

tests/CMakeLists.txt
```diff
@@ -7,12 +7,273 @@ if( GPERFTOOLS_FOUND )
...
+   endforeach()
+
+   set(${RESULT} ${tests} PARENT_SCOPE)
+endfunction()
+
+parse_unit_tests(TEST_CASES ${UNIT_TESTS})
+
+#message(STATUS "${TEST_CASES}")
+
+add_boost_test( chain_test
```
sgerbino: It looks like work was done to parse unit test suites and names automagically, but they are still listed manually.

tests/tests/operation_time_tests.cpp
```diff
@@ -235,7 +235,7 @@ BOOST_AUTO_TEST_CASE( comment_payout_dust )
 }
 
 /*
-BOOST_AUTO_TEST_CASE( reward_funds )
+OOST_AUTO_TEST_CASE( reward_funds )
```
sgerbino: Looks like a typo, should we just remove these?

tests/CMakeLists.txt
mvandeberg: Using a regex to pull out `BOOST_AUTO_TEST_CASE` gets some tests that have been commented out or in the case of SMTS, not compiled in on certain builds. When a test is ran and not found, it returns a non-zero exit code, which would result in CI failing. Some more fine tuning of the build system needs to be done. But for now, I parsed and logged the tests cases and then manually removed the ones we don't want to run. I am viewing this as a first step, but it accomplishes the goal of reducing CI time.

PR #3344, opened 2019-05-31
Downvote Pool

### LINKED ISSUE #3336: Down Vote Pool
Implement a separate downvote pool that is a percentage of the max size of the voting mana. Downvotes will be taken from the downvote pool first, and then the upvote pool once the downvote pool has been consumed. The downvote pool will follow the same rules as the upvote pool, regenerating over five days and filling instantly and proportionally to new Steem Power and delegations.

mvandeberg:
Closes #3336


libraries/chain/steem_evaluator.cpp
```diff
@@ -1936,7 +1956,30 @@ void hf20_vote_evaluator( const vote_operation& o, database& _db )
 
       _db.modify( voter, [&]( account_object& a )
       {
-         a.voting_manabar.use_mana( used_mana.to_uint64() );
+         if( _db.has_hardfork( STEEM_HARDFORK_0_21__3336 ) && dgpo.downvote_pool_percent > 0 && o.weight < 0 )
+         {
+            if( used_mana.to_uint64() > a.downvote_manabar.current_mana )
+            {
+               /* used mana is always less than downvote_mana + voting_mana because the amount used
+                * is a fraction of max( downvote_mana, voting_mana ). If more mana is consumed than
```
eonwarped: This comment doesn't appear to be true since used mana is taking a fraction of max(4*dvmana, mana) under current parameters. Is there an issue when someone drains their regular mana and then starts draining the DV mana?  Or is that fraction guaranteed to be < 1/4? (Or dvpoolfraction).

roadscape:
Does this keep the total mana (upvote+downvote) constant, or do users get an increased allowance of rshares/mana through the new downvote pool?

libraries/chain/steem_evaluator.cpp
mvandeberg: Currently, used mana ends up being 1/50th of the remaining voting mana.

The downvote logic uses up 1/50th of the remaining voting mana or 4/50th of the remaining downvote mana, whichever is greater (from downvote mana, then upvote mana). If 1/50th of the remaining voting mana is larger than the remaining downvote mana, the the remainder is taken from the voting mana. In all cases, the amount of used mana must always be less than the remaining voting mana because it is 1/50th of the remaining mana.

In the other scenario, where downvote mana is high and voting mana is low, the downvote mana will not be over consumed because the max that can be used is 4/50th. Once downvote mana is consumed low enough that upvote mana is being used, then 1/50th of the voting mana is used.

Under no scenario can the amount of used mana be more than voting mana + downvote mana because used mana is always a fraction of remaining mana.

mvandeberg:
@roadscape It doesn't matter because all users get a stake proportional bump. The long term economics would be identical if we reduced max mana to 80% and created a 20% downvote pool.

roadscape:
The decision will affect UI; currently the vote mana lines up nicely to total effective vests. E.g. if you have 15 SP, you have 15 SP max of voting mana. For such a complex system this simplicity is welcome.

If the approach is to add downvote mana, there remains 15 SP of voting mana plus 3 SP of downvote mana. Relevant UIs would need to decide how to represent this logically. Simplest would probably be just a separate downvote bar, leaving the original untouched.

If the approach is to reduce to 80% and split 20% to a new pool, the total voting mana remains at 15 SP. If this was represented graphically as a single bar, users may be confused that the last 3 SP of mana cannot be used unless downvoting. Otherwise the display could be split into two bars, one being 12 SP max and the other 3 SP.

The former seems simpler to explain and display. Your comment seems to imply that is the case.

libraries/chain/steem_evaluator.cpp
```diff
@@ -2046,7 +2089,31 @@ void hf20_vote_evaluator( const vote_operation& o, database& _db )
 
       _db.modify( voter, [&]( account_object& a )
       {
-         a.voting_manabar.use_mana( used_mana.to_uint64() );
+         if( _db.has_hardfork( STEEM_HARDFORK_0_21__3336 ) && dgpo.downvote_pool_percent > 0 && o.weight < 0 )
+         {
+            if( used_mana.to_int64() > a.downvote_manabar.current_mana )
+            {
+               idump( (used_mana)(a.downvote_manabar)(a.voting_manabar) );
```
sgerbino: Extraneous logging

sgerbino:
LGTM other than `idump`

mvandeberg:
@roadscape The current implementation is the 15 + 3 model. Both the voting manabar and the downvote manabar units are in satoshi VESTS. Even if UIs do not immediately update to show downvote mana, voting mana will continue to work exactly as it does now.

PR #3345, opened 2019-05-31
Adjust HF21 reward fund and SPS inflation

### LINKED ISSUE #3328: Update Reward Fund to 50% Curation
When the hardfork is applied, set the reward fund object to 50% curation.

### LINKED ISSUE #3346: Set SPS Inflation
Branch `20190522-sps-inflation` already adds the inflation code to send SBD to the SPS.

The inflation values need to be set at the hardfork and a virtual op needs to be created once per maintenance interval that documents how much SBD was added to the SPS from inflation in that maintenance interval.

sgerbino:
Resolves #3328.
Resolves #3346.


libraries/chain/database.cpp
```diff
@@ -5235,6 +5235,12 @@ void database::apply_hardfork( uint32_t hardfork )
          auto change_request = find< change_recovery_account_request_object, by_account >( STEEM_TREASURY_ACCOUNT );
          if( change_request )
             remove( *change_request );
+
+         modify( get< reward_fund_object, by_name >( STEEM_POST_REWARD_FUND_NAME ), [&]( reward_fund_object& rfo )
+         {
+            rfo.percent_content_rewards = 50 * STEEM_1_PERCENT;
```
mvandeberg: I was mistaken when we talked. This needs to be 100%.

This is the percent of all content rewards that this reward fund should get. When we implemented reward funds we were planning to have separate reward funds for root posts and comments.

libraries/chain/database.cpp
```diff
@@ -2190,15 +2192,25 @@ void database::process_funds()
...
       modify( props, [&]( dynamic_global_property_object& p )
       {
          p.total_vesting_fund_steem += asset( vesting_reward, STEEM_SYMBOL );
          if( !has_hardfork( STEEM_HARDFORK_0_17__774 ) )
             p.total_reward_fund_steem  += asset( content_reward, STEEM_SYMBOL );
-         p.current_supply           += asset( new_steem, STEEM_SYMBOL );
-         p.virtual_supply           += asset( new_steem, STEEM_SYMBOL );
+         p.current_supply      += asset( new_steem, STEEM_SYMBOL );
+         p.current_sbd_supply  += new_sbd;
+         p.virtual_supply      += asset( new_steem, STEEM_SYMBOL );
```
mvandeberg: The virtual supply should be updated with the sps_fund value as well `asset( new_steem + sps_fund, STEEM_VALUE)`. It will be automatically updated later in the block, but we should be proactive here to prevent any inaccuracies that might have unintended side effects in other parts of the code.

libraries/chain/util/sps_processor.cpp
```diff
@@ -296,6 +296,26 @@ void sps_processor::run( const block_notification& note )
 {
    remove_old_proposals( note );
    make_payments( note );
+   record_funding( note );
```
mvandeberg: I think this should be done before making payments so that the inflation is recorded in account history before it is paid out to proposals.

mvandeberg:
Merge conflicts.

PR #3347, opened 2019-05-31
Prevent STEEM transfer to SPS

### LINKED ISSUE #3343: Prevent transfer of STEEM/SP to steem.dao
The Steem SPS only pays out in SBD. The system supports a donation system via transfers to `steem.dao`. STEEM and SP transfers should be prevented to stop users from attempting to fund the SPS with STEEM or SP.

The ops that need to check for this are:

`transfer_operation`
`transfer_to_vesting_operation`
`set_withdraw_vesting_route_operation`
`transfer_to_savings_operation`
`transfer_from_savings_operation`

Comment payout beneficiaries will be able to set `steem.dao` as a beneficiary and in that case all STEEM and SP rewards will be converted to SBD and deposited in `steem.dao`.

mvandeberg:
Closes #3343


tests/tests/operation_tests.cpp
```diff
@@ -6637,6 +6709,19 @@ BOOST_AUTO_TEST_CASE( comment_beneficiaries_apply )
       ACTORS( (alice)(bob)(sam)(dave) )
       generate_block();
 
+      db_plugin->debug_update( [=]( database& db )
+      {
+         db.modify( db.get_dynamic_global_properties(), [=]( dynamic_global_property_object& gpo )
+         {
+            // Update to turn off inflation
```
sgerbino: Let's merge inflation branch and merge master to this branch. Then fix this test.

Smittyvb:
Escrow transfers also need to be prevented.

mvandeberg:
Escrow does not need to be prevented because an escrow transfer need to be approved by the sender, receiver, and escrow agent before the ratification deadline. If it is not (which it won't be because the SPS cannot cannot sign those transactions) the funds will be returned to the sender. Funds cannot be permanently locked in this manner and it is no more dangerous to transfer via escrow to the SPS as it is an account with whom you have not previously made arrangements with.

PR #3388, opened 2019-07-01
Fix URI to testing.md

Smittyvb:
It got broke when it was moved to the ``devs`` directory.


sgerbino:
Thank you for proposing these corrections. If there are more, please try to include more of them in a single PR. There is just a bit of overhead for each community PR (we have to make a branch and merge your changes in each time).

PR #3390, opened 2019-07-02
merging

steemdevelopment:


steemdevelopment:
Sorry It auto went to the main rep instead of the fork rep master.

Issue #1519, opened 2017-09-06
Implement extended fund destination syntax


theoreticalbts:
SMT syntax has some specials like `smt.rewards` for the rewards pool, `alice.vesting` for a user's vesting balance, etc.  Also `$from`.

Precise syntax specification, parsing, and serialization of these "destinations" needs to be developed.

Operations which potentially take a destination should have a different field type.

theoreticalbts:
Moving to SMT project and unassigning myself.  This isn't necessary for MVP.  Maybe we should add this feature once the base functionality of SMT's is a little more developed.

youkaicountry:
Closing as outdated.

PR #3407, opened 2019-07-18
SMT ICO

### LINKED ISSUE #2731: As a contributor to a failed ICO, I want to be able to recall my contribution.
As a contributor to a failed ICO, or perceived failed ICO, I want to recall my contribution.

Entails:
smt_refund_operation
post_contribution_pause

```c++
struct smt_refund_operation
{
   account_name_type executor;
   asset_symbol_type symbol;
   extensions_type   extensions;
};
```

```c++
struct smt_refund_action
{
   account_name_type  contributor;
   asset_symbol_type  symbol;
   uint32_t           contribution_id;
   asset              contribution;
   extensions_type    extensions;
};
```

AC:

1. After an SMT's announced launch time if all hidden caps have been revealed and the ICO has failed anyone can submit an `smt_refund_operation`.
2. After an SMT's announced launch time plus 24 hours (known as the post contribution pause) regardless of whether hidden caps have not been revealed anyone can submit an `smt_refund_operation`.
3. The `smt_refund_operation` begins a series of cascading automated actions that refund contributions to an ICO.
4. This moves the SMT in to the failed state.
5. The cascading `smt_refund_action`s can be of O(n). However, it corresponds 1:1 with contributions. `smt_contribution_operation` should cover the cost of the refund. 

Tasks:
-[ ] Validation/testing of `smt_refund_operation`
-[ ] Evaluation/testing of `smt_refund_operation`
-[ ] Evaluation/testing of `smt_refund_action`
-[ ] Update `smt_contribution_operation` RC cost to include costs of refunds.
-[ ] Update SMT Whitepaper

Related #2251, #2056

### LINKED ISSUE #2732: ICO Contributor Payout
As a contributor to an ICO, I want to receive my SMT.

```c++
struct smt_contributor_payout_action
{
   account_name_type contributor;
   uint32_t          contribution_id;
   asset             payout;
};
```

**AC:**
1. At the SMT launch time, if all caps are revealed, trigger cascading `smt_contributor_payout_action`s.
2. If the SMT launch time is passed and caps are revealed, trigger cascading `smt_contributor_payout_action`s.

In whitepaper see sections `Token units` and `Unit Ratios`

### LINKED ISSUE #2733: Route ICO tokens to founders
After a successful ICO, tokens need to be routed to founders as specified during the token setup.

```c++
struct smt_founder_payout_action
{
   account_name_type destination;
   asset_symbol_type symbol;
   asset             payout;
};
```

**AC:**
1. Following ICO contributor payout, founders need to receive their tokens via cascading `smt_founder_payout_action`s.
2. Support extended fund destinations

Note:
This work is strongly related to #2734 #2735

In whitepaper see sections `Token units` and `Unit Ratios`

### LINKED ISSUE #2734: Implement ICO Hard Cap
Hard cap for ICOs

In whitepaper see sections `Token units` and `Unit Ratios`

### LINKED ISSUE #2735: Implement ICO Soft Cap
ICO soft cap.

In whitepaper see sections `Token units` and `Unit Ratios`

### LINKED ISSUE #2736: ICO Failed Launch
Handle the case of a failed ICO launch.

### LINKED ISSUE #2767: As a stakeholder in an ICO, I want to get my tokens.
- [x] #2732 
- [x] #2733 
- [x] #2734 
- [x] #2735 
- [x] #2736

sgerbino:
Resolves #2767 
Resolves #2732 
Resolves #2735 
Resolves #2734 
Resolves #2733 
Resolves #2736 
Resolves #2731


sgerbino:
### Items for discussion:
* `smt_capped_generation_policy::soft_cap_percent` purpose is not explicitly clear in the SMT whitepaper. It does mention that setting it to `STEEM_100_PERCENT` disables the soft cap, but it doesn't exist in the example ICO calculation spreadsheet, [ICO Parameters](https://github.com/steemit/smt-whitepaper/blob/master/smt-manual/ico-parameters.ods).
> We disable the soft cap by setting soft_cap_percent to STEEM_100_PERCENT = 10000.
* Based on the existing validator code, we only allow SMT target units to be `SMT_DESTINATION_FROM` and `SMT_DESTIONATION_FROM_VESTING` so I have implemented it that way.
### Todo:
* Implement a more complex SMT ICO payout test utilizing soft_cap/hard_cap `smt_generation_unit`, `smt_capped_generation_policy::soft_cap_percent`, and possibly other special account names if it is deemed required.

sgerbino:
## Updated Todo:
* Add `steem_units_min` field to `smt_setup_operation` to use as the minimum required STEEM to be considered an ICO success
* Calculate state transitions outside of `required_automated_actions` to allow for optimized replay times
* Remove `soft_cap_percent` as this is no longer necessary
* Reimplement ICO queue processing as `required_automated_actions`

libraries/chain/smt_evaluator.cpp
```diff
@@ -162,10 +164,18 @@ void smt_setup_evaluator::do_apply( const smt_setup_operation& o )
       token_ico_obj.launch_time = o.launch_time;
       token_ico_obj.steem_units_soft_cap = o.steem_units_soft_cap;
       token_ico_obj.steem_units_hard_cap = o.steem_units_hard_cap;
+      token_ico_obj.steem_units_min = o.steem_units_min;
    } );
 
    smt_setup_evaluator_visitor visitor( token_ico, _db );
    o.initial_generation_policy.visit( visitor );
+
+   if ( !_db.is_pending_tx() )
```
mvandeberg: This is going to be a required check for any action that is pushed during as a result of a transaction. Let's move this check in to `push_required_action`.

sgerbino:
## Todo

* Change `smt_phase` to something more sensible:
```
setup,
setup_completed,
ico,
ico_completed,
launch_failed,
launch_success 
```
* Move `_db.is_pending_tx()` inside `push_required_action` and `push_optional_action`

libraries/protocol/include/steem/protocol/smt_operations.hpp
```diff
@@ -13,8 +13,13 @@
 #define SMT_MIN_SATURATION_STEEM_UNITS 1000
 #define SMT_MIN_SOFT_CAP_STEEM_UNITS   1000
 
-namespace steem { namespace protocol {
+#define SMT_DESTINATION_FROM          account_name_type( "$from" )
+#define SMT_DESTINATION_FROM_VESTING  account_name_type( "$from.vesting" )
+#define SMT_DESTINATION_MARKET_MAKER  account_name_type( "$market_maker" )
```
mvandeberg: I don't see that market maker is used anywhere. While we are not launching with the market maker enabled, ICOs should still be able to route tokens (both STEEM and SMTs) to the market maker for future use.

mvandeberg:
There seems to be a glaring hole in the documentation of SMT Extended Fund Destinations. It was brought up in #1519 to clarify, but was seemingly closed without resolution. Or at the very least closed without documenting the resolution.

It looks like all of the `$` prefixed destinations are handled, but there is one outlier documented in #1519 that isn't handled, `alice.vesting`. `.` is not a valid character in an account name, so this can easily be parsed by splitting on `.`. The use for this is to power up founder STEEM/SMTs automatically. While it could be powered up by the user, doing this automatically in the smart contract does not require contributor to need to trust the SMT creator to do so.

Additionally, because this can cause token destinations to overflow the 16 character account name type, fund destination types should be expanded to a 32 bit fixed length string field.

libraries/chain/util/smt_token.cpp
```diff
@@ -68,6 +69,253 @@ fc::optional< time_point_sec > last_emission_time( const database& db, const ass
...
+         } );
+      }
+   }
+}
+
+bool schedule_next_refund( database& db, const asset_symbol_type& a )
+{
+   bool action_scheduled = false;
+   auto& idx = db.get_index< smt_contribution_index, by_symbol_id >();
+   auto itr = idx.lower_bound( boost::make_tuple( a, 0 ) );
```
mvandeberg: This doesn't need to change, but in the future, you shouldn't need to create a tuple with an id of 0. Passing in just the asset symbol type should do the same thing.

libraries/chain/util/smt_token.cpp
```diff
@@ -68,6 +69,253 @@ fc::optional< time_point_sec > last_emission_time( const database& db, const ass
...
+
+   if ( itr != idx.end() && itr->symbol == a )
+   {
+      smt_refund_action refund_action;
+      refund_action.symbol = itr->symbol;
+      refund_action.contributor = itr->contributor;
+      refund_action.contribution_id = itr->contribution_id;
+      refund_action.refund = itr->contribution;
+
+      db.push_required_action( refund_action, db.head_block_time() + STEEM_BLOCK_INTERVAL );
```
mvandeberg: Let's add an override 

``` c++
database::push_required_action( const required_automated_action& a )
{
   push_required_action( a, head_block_time() + STEEM_BLOCK_INTERVAL )
}
```

Seems like another common use case.

libraries/protocol/include/steem/protocol/steem_required_actions.hpp
```diff
@@ -16,8 +18,76 @@ namespace steem { namespace protocol {
    };
 #endif
 
+#ifdef STEEM_ENABLE_SMT
+   struct smt_refund_action : public base_operation
+   {
+      asset_symbol_type symbol;
```
mvandeberg: This is minor, but let's put contributor first. Our json output will be in field order and I think it makes more sense to most people to their the account name first.

libraries/protocol/include/steem/protocol/steem_required_actions.hpp
```diff
@@ -16,8 +18,76 @@ namespace steem { namespace protocol {
...
+      asset             refund;
+
+      void validate()const;
+
+      friend bool operator==( const smt_refund_action& lhs, const smt_refund_action& rhs );
+   };
+
+   struct smt_contributor_payout_action : public base_operation
+   {
+      asset_symbol_type    symbol;
```
mvandeberg: Contributor first

libraries/protocol/include/steem/protocol/steem_required_actions.hpp
```diff
@@ -16,8 +18,76 @@ namespace steem { namespace protocol {
...
+   {
+      asset_symbol_type                                   symbol;
+      std::map< account_name_type, std::vector< asset > > payouts;
+
+      void validate()const;
+
+      friend bool operator==( const smt_founder_payout_action& lhs, const smt_founder_payout_action& rhs );
+   };
+
+   struct smt_ico_launch_action : public base_operation
```
mvandeberg: Add the control account. It is not needed for the action, but we need to statelessly get the control account so this can go in the account history.

libraries/protocol/include/steem/protocol/steem_required_actions.hpp
```diff
@@ -16,8 +18,76 @@ namespace steem { namespace protocol {
...
+   {
+      asset_symbol_type symbol;
+
+      void validate()const;
+      friend bool operator==( const smt_ico_launch_action& lhs, const smt_ico_launch_action& rhs );
+   };
+
+   struct smt_ico_evaluation_action : public base_operation
```
mvandeberg: Add control account

libraries/protocol/include/steem/protocol/steem_required_actions.hpp
```diff
@@ -16,8 +18,76 @@ namespace steem { namespace protocol {
...
+   {
+      asset_symbol_type symbol;
+
+      void validate()const;
+      friend bool operator==( const smt_ico_evaluation_action& lhs, const smt_ico_evaluation_action& rhs );
+   };
+
+   struct smt_token_launch_action : public base_operation
```
mvandeberg: Add control account

mvandeberg:
Because actions are operations in the class hierarchy, we can use `get_required_active_authorities` or we can add them to the `impacted` visitor. I don't care which one is used. I believe I used `get_required_active_authorites` for the example actions in the tests.

libraries/chain/util/smt_token.cpp
```diff
@@ -68,6 +69,262 @@ fc::optional< time_point_sec > last_emission_time( const database& db, const ass
...
+      return true;
+   if ( name == SMT_DESTINATION_VESTING )
+      return true;
+
+   return false;
+}
+
+account_name_type get_account( const unit_target_type& unit_target )
+{
+   FC_ASSERT( is_valid_account_name( unit_target ) );
```
mvandeberg: Won't this fail on routes prefixed with `$`, which are not valid in account names?

I would also not name this `get_account` because it looks similar to `database::get_account`.

libraries/protocol/smt_operations.cpp
```diff
@@ -35,18 +22,20 @@ void smt_create_operation::validate()const
       ("prec1",symbol.decimals())("prec2",precision) );
 }
 
-bool is_valid_unit_target( const account_name_type& name )
+bool is_valid_unit_target( const unit_target_type& name )
```
mvandeberg: This needs to return true for `$alice.vesting`, and currently does not.

libraries/chain/util/smt_token.cpp
sgerbino: This is not implemented yet, it is there so I can add `unit_target_type` and keep it compiling.

libraries/protocol/smt_operations.cpp
sgerbino: This is not implemented yet.

libraries/chain/util/smt_token.cpp
sgerbino: This will (when implemented), either return `unit_target_type` as an `account_name_type`, e.g. "alice". Or extract the `account_name_type` from a special destination that is an account, e.g "$alice.vesting".

Issue #3184, opened 2018-11-28
1% SBD print rate despite the haircut


economicstudio:
Despite the haircut, SBD is still being printed at 1% rate.

I made a detailed post here: https://steemit.com/utopian-io/@blockchainstudio/why-sbd-print-rate-is-still-1-despite-the-haircut-bug-report-explanation-and-suggestions
, where I explained in detail with suggestions.

TimCliff:
It is related to the discussion here: https://github.com/steemit/steem/issues/3176

oflyhigh:
* [Why is median_history_price no longer the median of feed price? My investigation.](https://steemit.com/witness-category/@oflyhigh/why-is-medianhistoryprice-no-longer-the-median-of-feed-price-my-investigation)
*

ajayyy:
SBD from posts cannot end at 10%, it must end earlier or else SBD will continue to be printed FOREVER. SBD printing must revert back to the previous schedule.

iamsmooth:
Reverting back to the previous schedule is certainly not needed. Other than changing the blockchain logic on how the cap is implemented, probably a change to the top of the print phase out range from 10% to 9.9% would be sufficient to stop it.

As it is, this is entirely irrelevant economically, although confusing and messy. A total of about 60-70 SBD per day are currently being printed (due to the bug). This is literally an effect of 0.000065% per day to the debt ratio and will never add up to anything significant, while at the same time, a minimum hundreds and usually many more (tens of thousands) are being destroyed daily by conversions, meaning the SBD supply is still _declining_ despite the extra 60-70 being printed (even when SBD was way over $1 and it made no sense to convert, even mistaken conversions might have been over 60-70 per day).

In no sense is this important enough for an emergency hard fork, if it is even worth fixing at all.

economicstudio:
I agree with smooth most of the part. It's too late to change the current feed schedule back. Also, while this is a very important issue, It's not urgent at all in terms of impact, especially given that the resources that Steemit, Inc. currently have.

But I strongly believe that feed api information should be self-contained, i.e., provide 3 information: base, quote, haircut (or debt) ratio. (not just 2, which is again very inconsistent, current value is haircut, historical value is non-haircut) Then everything will be much clearer including the code.

[...]

PR #3410, opened 2019-07-19
Fix sbd print rate

### LINKED ISSUE #3409: `sbd_print_rate` is 1% in the haircut
When the haircut price applies, the virtual supply is:
`virtual_supply = current_supply + current_supply/9`

What is the `sbd_print_rate`?
https://github.com/steemit/steem/blob/a470bc283260e582b56897bd5c68107aa0289f18/libraries/chain/database.cpp#L3901-L3909

```percent_sbd = ( current_supply/9 * 10000 ) / ( current_supply + current_supply/9 )```

The problem with this formula is that all operations are made with integers. Meaning that the **percent_sbd will be 1000 only if the `current_supply` is divisible by 9. Otherwise it always be 999**.

If `percent_sbd == 999` then `sbd_print_rate=1%`.

joticajulian:
A previous fix was submitted in #3291
However that solution creates new problems: The sbd print rate jumps from 5% to 0%.

In this PR I propose calculate `percent_sbd` by **rounding** the number instead of getting the floor (which is a intrinsic characteristic of integers). To do that I use the function described [here](https://stackoverflow.com/a/2422723/5938661).

This closes #3409 and #3184


mvandeberg:
I prefer this solution as the "more correct" solution. However, we are closing in on a release of 0.21.0 and this change would require a reindex because it change the dynamic global property object. We have a couple of options...

1. Postpone this change until the SMT hardfork and then clean up as needed, potentially removing the stop adjust code altogether if it is not triggered prior to the SMT hardfork.

2. Merge this PR in two parts. The first removing the stop adjust code behavior, but not the field from the dgpo, to remove the need for a replay and not delaying the release. The second part would be done after the release and clean up the dgpo for the SMT release, which would require a replay anyway.

@TimCliff @eonwarped Thoughts on the above propositions?

mvandeberg:
There was an unmerged change from testing that requires a reindex, so merging this change will not add additional delay to tagging a release.

TimCliff:
I’m happy with the change. The version that I added in was somewhat of a hack to just make the problem go away. My main concern was adding additional work to get the change merged in prior to the hardfork, but that seems like a non issue. All good from my end 👍

PR #3414, opened 2019-07-23
Assert account has sufficient funds in adjust_balance

### LINKED ISSUE #1629: Move sufficient funds checks to adjust_balance
A lot of operations (and a single `database` method) have checks of the form:

```
FC_ASSERT( get_balance( account, sym ) <= op.field );
db.adjust_balance( account, -op.field );
```

Let's have these checks occur just once, in `adjust_balance()`, which will make it less likely that new operations will omit such checks by mistake.  It will also be an optimization because it only requires the balance to be fetched once, which will make a difference for SMT's.

sgerbino:
Resolves #1629.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -4537,21 +4537,45 @@ void database::adjust_balance( const account_object& a, const asset& delta )
    bool check_balance = has_hardfork( STEEM_HARDFORK_0_20__1811 );
 
 #ifdef STEEM_ENABLE_SMT
-   // No account object modification for SMT balance, hence separate handling here.
-   // Note that SMT related code, being post-20-hf needs no hf-guard to do balance checks.
    if( delta.symbol.space() == asset_symbol_type::smt_nai_space )
    {
+      // No account object modification for SMT balance, hence separate handling here.
+      // Note that SMT related code, being post-20-hf needs no hf-guard to do balance checks.
       smt_regular_balance_operator balance_operator( delta );
       adjust_smt_balance< account_regular_balance_object >( a.name, delta, false/*check_account*/, balance_operator );
-      return;
    }
+   else
 #endif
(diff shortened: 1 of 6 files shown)


libraries/chain/database.cpp
```diff
@@ -4543,20 +4551,7 @@ void database::adjust_balance( const account_object& a, const asset& delta )
...
-   if( delta.symbol.space() == asset_symbol_type::smt_nai_space )
-   {
-      smt_regular_balance_operator balance_operator( delta );
-      adjust_smt_balance< account_regular_balance_object >( name, delta, true/*check_account*/, balance_operator );
-      return;
-   }
-#endif
-   const auto& a = get_account( name );
-   modify_balance( a, delta, check_balance );
+   adjust_balance( get_account( name ), delta );
```
mvandeberg: We should overload `get_balance` as well to accept an account name so when we get the balance of an SMT we don't need to first look up the account and then the SMT balance object.

sgerbino: It looks like we already have that, the implementation is just inside the header. For consistency I will move it into the implementation file.

mvandeberg: The implementation in the header is wrong. The point of having get_balance overloaded to take an account name is that when we are looking up information about an SMT, there is a good chance that we haven't looked up the account object, nor do we need to, because an SMT balance is not stored on the account object. The current implementation still looks up an account object to pass to the account object version that, if the symbol is an SMT, grabs the account name and looks up the SMT balance object which results in an unnecessary database read.

This might not be the best decomposition, but I think it might make sense...

For each version of `get_balance/adjust_balance` first place an SMT/core asset check. Place the SMT/core asset implementation in the appropriately optimized overload and as an else condition to the above check/convert from account name/account object and call the overload if the suboptimal overload is being called.

PR #3415, opened 2019-07-23
Fix sbd print rate

### LINKED ISSUE #3409: `sbd_print_rate` is 1% in the haircut

mvandeberg:
Closes #3409

Local PR for #3410

> A previous fix was submitted in #3291
> However that solution creates new problems: The sbd print rate jumps from 5% to 0%.
> 
> In this PR I propose calculate percent_sbd by rounding the number instead of getting the floor (which is a intrinsic characteristic of integers). To do that I use the function described here.
> 
> This closes #3409 and #3184


libraries/chain/database.cpp
```diff
@@ -3901,9 +3900,15 @@ void database::update_virtual_supply()
          auto percent_sbd = uint16_t( ( ( fc::uint128_t( ( dgp.current_sbd_supply * get_feed_history().current_median_history ).amount.value ) * STEEM_100_PERCENT )
```
sgerbino: It looks like this statement would be unnecessary after hardfork 21 and should probably go in an `else` statement.

libraries/chain/database.cpp
```diff
@@ -3898,12 +3897,22 @@ void database::update_virtual_supply()
...
+         uint16_t percent_sbd = 0;
+
+         if( has_hardfork( STEEM_HARDFORK_0_21 ) )
+         {
+            percent_sbd = uint16_t( ( ( fc::uint128_t( ( dgp.current_sbd_supply * get_feed_history().current_median_history ).amount.value ) * STEEM_100_PERCENT + dgp.virtual_supply.amount.value/2 )
+               / dgp.virtual_supply.amount.value ).to_uint64() );
+         }
+         else
+         {
+            percent_sbd = uint16_t( ( ( fc::uint128_t( ( dgp.current_sbd_supply * get_feed_history().current_median_history ).amount.value ) * STEEM_100_PERCENT )
```
eonwarped: What about the HF20 behavior? It looks like it isn't happening here?

mvandeberg: The else block is the HF 20 behavior. What about this looks wrong to you?

eonwarped: The buffer of 5 for the stop print rate gap that was specific to HF20. And then HF19 had that parameter set to zero. Isn't this going to behave differently for the HF20 blocks?

joticajulian: This buffer of 5 was only for the HF21. Not present in HF20 or HF19.

eonwarped: Ah that's what I was missing. Thanks!

PR #3440, opened 2019-07-31
Update default configs and exchange documentation

jredbeard:
This updates the default example config file as well as the default docker config file. Exchange documentation has been brought up-to-date including new instructions for config files and new instructions for upgrading.

There are a few minor differences between the two config files. The non-docker example config has `account_history` and `account_history_api` listed along with `account-history-track-account-range`, and it also includes `shared-file-full-threshold` and `shared-file-scale-rate`. These are all things that are set by the docker image when using the docker config file.

### DIFF
--- a/doc/exchangequickstart.md
+++ b/doc/exchangequickstart.md
@@ -1,9 +1,9 @@
 Exchange Quickstart
 -------------------
 
-System Requirements: A dedicated server or virtual machine with a minimum of 64GB of RAM, and at least 220GB of fast **local** SSD storage. STEEM is one of the most active blockchains in the world and handles an incredibly large amount of transactions per second, as such, it requires fast storage to run efficiently.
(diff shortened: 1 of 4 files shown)


doc/exchangequickstart.md
```diff
@@ -90,7 +100,7 @@ You can see that the container is running with the `docker ps` command.
 
 To follow along with the logs, use `docker logs -f`.
 
-Initial syncing will take between 6 and 48 hours depending on your equipment, faster storage devices will take less time and be more efficient. Subsequent restarts will not take as long.
+Initial syncing will take between 6 and 72 hours depending on your equipment, faster storage devices will take less time and be more efficient. Subsequent restarts will not take as long.
```
mvandeberg: Is it even worth listing this as a lower bound?

PR #3448, opened 2019-08-01
Add db env skip

### LINKED ISSUE #3446: database created by a different compiler, build, or operating system
I build v0.20.12-no-mira from the source, and then replaced the steemd (v0.20.10).

But when I try to run it, i got the following error messages:
>Error opening database. If the binary or configuration has changed, replay the blockchain explicitly. Error: {"code":13,"name":"N5boost16exception_detail10clone_implINS0_19error_info_injectorISt13runtime_errorEEEE","message":"database created by a different compiler, build, or operating system","stack":[{"context":{"level":"warn","file":"database.cpp","line":184,"method":"open","hostname":"","timestamp":"2019-08-01T15:37:50"}
....

But in the v0.20.12-no-mira release note. it shows that:
>Steem 0.20.12 does require reindexing from all previous versions.

So, What's wrong?

mvandeberg:
Closes #3446

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -115,7 +115,7 @@ void database::open( const open_args& args )
    try
    {
       init_schema();
-      chainbase::database::open( args.shared_mem_dir, args.chainbase_flags, args.shared_file_size, args.database_cfg );\
+      chainbase::database::open( args.shared_mem_dir, args.chainbase_flags, args.shared_file_size, args.database_cfg );
 
       initialize_indexes();
       initialize_evaluators();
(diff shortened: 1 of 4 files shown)


libraries/chain/database.cpp
```diff
@@ -115,7 +115,7 @@ void database::open( const open_args& args )
    try
    {
       init_schema();
-      chainbase::database::open( args.shared_mem_dir, args.chainbase_flags, args.shared_file_size, args.database_cfg );
+      chainbase::database::open( args.shared_mem_dir, args.chainbase_flags, args.shared_file_size, args.database_cfg );\
```
sgerbino: Rogue backslash

PR #3457, opened 2019-08-08
SMT Setup Input & Tests

### LINKED ISSUE #3379: Cleanup smt setup tests
The tests for smt_setup_operation are messy and do not follow our testing paradigm. They should be cleaned up.

### LINKED ISSUE #3455: Review smt_setup_operation destination inputs
The full extended fund destination code was not written until #3407. This means that extended fund destinations were not properly tested during SMT setup. The following fund destinations need to be tested during SMT setup.

| Special Destinations | Valid ICO STEEM Dest | Valid ICO SMT Dest | Valid Emission Dest
|---------------------:|:--------------------:|:------------------:|:-------------------:|
| `$from` |   | X |   |
| `$from.vesting` |   | X |   |
| `$rewards` |   | X | X |
| `$vesting` |   |   | X |
| `$market_maker` | X | X | X |
| `${account}.vesting` | X | X |   |
| `{account}` | X | X |   |

#3379 Has related work to clean up SMT setup tests. This work will have some overlap and can be done together.

sgerbino:
Resolves #3379.
Resolves #3455.

### DIFF
--- a/libraries/chain/util/smt_token.cpp
+++ b/libraries/chain/util/smt_token.cpp
@@ -70,23 +70,6 @@ fc::optional< time_point_sec > last_emission_time( const database& db, const ass
    return {};
 }
 
-namespace unit_target {
-
-void verify_accounts( database& db, const flat_map< unit_target_type, uint16_t >& units )
-{
-   for ( auto& unit : units )
-   {
-      if ( !protocol::smt::unit_target::is_account_name_type( unit.first ) )
-         continue;
-
-      auto account_name = protocol::smt::unit_target::get_unit_target_account( unit.first );
-      const auto* account = db.find_account( account_name );
-      FC_ASSERT( account != nullptr, "The provided account unit target ${target} does not exist.", ("target", unit.first) );
-   }
-}
-
-} // steem::chain::util::smt::unit_target
-
 namespace ico {
 
 share_type payout( database& db, const asset_symbol_type& symbol, const account_object& account, const std::vector< asset >& assets )
--- a/libraries/chain/util/smt_token.cpp
+++ b/libraries/chain/util/smt_token.cpp
@@ -70,6 +70,23 @@ fc::optional< time_point_sec > last_emission_time( const database& db, const ass
    return {};
 }
 
+namespace unit_target {
+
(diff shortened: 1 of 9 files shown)


libraries/chain/util/smt_token.cpp
```diff
@@ -70,6 +70,23 @@ fc::optional< time_point_sec > last_emission_time( const database& db, const ass
    return {};
 }
 
+namespace unit_target {
+
+void verify_accounts( database& db, const flat_map< unit_target_type, uint16_t >& units )
```
mvandeberg: This can be placed in `smt_evaluator.cpp` directly. It is only used in the evaluator. No need to declare in a header. If is also directly tied to evaluation of the setup operation.

tests/tests/smt_tests.cpp
```diff
@@ -486,7 +535,7 @@ BOOST_AUTO_TEST_CASE( setup_apply )
 {
    try
    {
-      ACTORS( (alice)(bob) )
+      ACTORS( (alice)(bob)(xyz)(xyz2) )
```
mvandeberg: Is this needed?

tests/tests/smt_tests.cpp
```diff
@@ -542,14 +591,6 @@ BOOST_AUTO_TEST_CASE( setup_apply )
    FC_LOG_AND_RETHROW()
 }
 
-/*
- * SMT legacy tests
- *
- * The logic tests in legacy tests *should* be entirely duplicated in smt_operation_tests.cpp
- * We are keeping these tests around to provide an additional layer re-assurance for the time being
- */
-FC_TODO( "Remove SMT legacy tests and ensure code coverage is not reduced" );
```
mvandeberg: Do we no longer need to remove these tests or is there a compelling reason to keep them?

tests/tests/smt_tests.cpp
sgerbino: Yes this makes the old test pass. We now check that accounts actually exist before allowing the operation to go through. I re-wrote this test, but left the old one in there for now to ensure we do not lose code coverage.

PR #3462, opened 2019-08-13
Refactor vote_power_reserve_rate

### LINKED ISSUE #1411: Change from target votes per day to target votes per period
According to the comment, currently `vote_power_reserve_rate` is the number of votes regenerated per day.  We want to change this to the number of votes regenerated per regeneration period.

I believe this refactor can be done without affecting past semantics, we simply change [this line](https://github.com/steemit/steem/blob/ce2a2c41a321fd9f437de5bed8b62a42314073f9/libraries/chain/steem_evaluator.cpp#L1175) to simply:

```
int64_t max_vote_denom = dgpo.votes_per_regeneration_period;
```

Additionally, remove the following `FC_ASSERT` or change it to a `static_assert`.

Related: #1252

mvandeberg:
Closes #1411

### DIFF
--- a/libraries/plugins/apis/condenser_api/include/steem/plugins/condenser_api/condenser_api.hpp
+++ b/libraries/plugins/apis/condenser_api/include/steem/plugins/condenser_api/condenser_api.hpp
@@ -375,7 +375,8 @@ struct extended_dynamic_global_properties
       recent_slots_filled( o.recent_slots_filled ),
       participation_count( o.participation_count ),
       last_irreversible_block_num( o.last_irreversible_block_num ),
-      vote_power_reserve_rate( o.vote_power_reserve_rate ),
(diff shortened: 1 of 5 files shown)


libraries/plugins/apis/condenser_api/include/steem/plugins/condenser_api/condenser_api.hpp
```diff
@@ -375,7 +375,8 @@ struct extended_dynamic_global_properties
       recent_slots_filled( o.recent_slots_filled ),
       participation_count( o.participation_count ),
       last_irreversible_block_num( o.last_irreversible_block_num ),
-      vote_power_reserve_rate( o.vote_power_reserve_rate ),
+      target_votes_per_period( o.target_votes_per_period),
```
sgerbino: Whitespace

PR #3467, opened 2019-08-15
Fix optional automated action mira iterator logic

sgerbino:
#3464 fixes only required automated actions but optional automated actions requires the same fix.


mvandeberg:
Optional actions should not be cascading, but this fix is not harmful to add.

Issue #3451, opened 2019-08-05
building.md is missing build instructions for Ubuntu 18.04


iamsmooth:
Ubuntu 18.04 is apparently a supported or even recommended platform now and should be addressed in the build instructions.

I was looking at 0.21.0 and did not check any of the other releases but I would guess they are the same.

iamsmooth:
To follow up, as far as I can tell the existing 16.04 instructions work fine on 18.04 as well, so this may be as simple as changing the header to include both.

iamsmooth:
Also, CONFIG_MIRA is missing

iamsmooth:
Also @Gandalf-the-Grey notes:

> also no info about nofiles limits, libboost-signals-dev might not be needed, submodules are no longer used, and I'm not sure about cmake options (I was providing them explicitly just in case) at least some of them are matching defaults

PR #3469, opened 2019-08-19
20190819 doc updates

Gandalf-the-Grey:
Fixes https://github.com/steemit/steem/issues/3451
+ some other minor changes.


Dockerfile
```diff
@@ -41,7 +41,6 @@ RUN \
         jq \
         wget \
         virtualenv \
-        gdb \
```
Gandalf-the-Grey: duplicate

Jolly-Pirate:
They kept `seed-node =` in the config.ini for backward compatibility.

Gandalf-the-Grey:
> They kept `seed-node =` in the config.ini for backward compatibility.

No. Most likely you are mixing that with `doc/seednodes.txt` which is still there for backward compatibility as some of the scripts are relying on it as the source of seed nodes

Dockerfile
BuHeP: 444

PR #3470, opened 2019-08-21
Votable Assets

### LINKED ISSUE #2747: As a content creator, I want to be able to specify which token(s) in which rewards are to be paid.
As part of comment options, authors can specify up to three (consider less) SMTs that are valid votable assets on the content.

Comments on posts inherit the votable assets of the root post.

Entails:
votable_assets
SMT and comment options
Decline payouts
active_phase

mvandeberg:
Closes #2747


libraries/chain/database.cpp
```diff
@@ -5476,10 +5476,7 @@ void database::validate_invariants()const
 
       for ( auto itr = smt_ico_idx.begin(); itr != smt_ico_idx.end(); ++itr )
       {
-         if ( get< smt_token_object, by_symbol >( itr->symbol ).phase <= smt_phase::launch_failed )
-            total_supply += asset( itr->contributed.amount - itr->processed_contributions, STEEM_SYMBOL );
-         else
-            total_supply += asset( itr->contributed.amount, STEEM_SYMBOL );
+         total_supply += asset( itr->contributed.amount - itr->processed_contributions, STEEM_SYMBOL );
```
sgerbino: This was probably removed by mistake. I think this change causes `smt_ico_payouts` and `smt_ico_payouts_special_destinations` to fail.

PR #3472, opened 2019-08-22
SMT Vesting

### LINKED ISSUE #2746: As a token holder, I wish to be able to vest / power down my tokens.
Implement vesting / power down for SMTs. 

Issue #2455 contains information regarding stability of vesting.

> To ensure numerical stability, we want to ensure that small `p` (pot) situations do not occur. To do this we are going to initialize the pot with ballast claims and tokens `c_0` and `p_0`, as described in the [SMT vesting wiki page](https://github.com/steemit/steem/wiki/SMT-Vesting).
> 
> - Let the worst case exhange rate `r_min` be 1000, and ballast tokens `p_0' be 100,000 satoshis.
> - Claim storage fields should be 128-bit
> 

There may be problems/bit-rot from a previous implementation discussed in #2253. It is not a requirement to salvage the existing implementation if making the changes complicates the implementation.

sgerbino:
Resolves #2746.


libraries/chain/database.cpp
```diff
@@ -1244,20 +1244,26 @@ asset create_vesting2( database& db, const account_object& to_account, asset liq
...
          FC_ASSERT( liquid.symbol.is_vesting() == false );
-         // Get share price.
-         const auto& smt = db.get< smt_token_object, by_symbol >( liquid.symbol );
-         FC_ASSERT( smt.allow_voting == to_reward_balance, "No voting - no rewards" );
-         price vesting_share_price = to_reward_balance ? smt.get_reward_vesting_share_price() : smt.get_vesting_share_price();
+
+         const auto& token = db.get< smt_token_object, by_symbol >( liquid.symbol );
+
+         if ( to_reward_balance )
+            FC_ASSERT( token.allow_voting == true, "Cannot create vests for the reward balance when voting is disabled." );
```
mvandeberg: This assertion should never fail as voting with a token that doesn't allow voting should not be allowed further up in the chain. As it turns out, it seems I forgot that assertion in my work on #2747

Can you add to the message that this assertion should not be triggered, so that if it is, it raises an alarm much faster?

libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
```diff
@@ -44,30 +44,34 @@ class smt_token_object : public object< smt_token_object_type, smt_token_object
...
 
-      return price( asset( total_vesting_shares, liquid_symbol.get_paired_symbol() ), asset( total_vesting_fund_smt, liquid_symbol ) );
+      share_type effective_total_vesting_shares = total_vesting_shares_ballast + total_vesting_shares;
+      share_type effective_total_vesting_fund   = total_vesting_fund_ballast + total_vesting_fund_smt;
+
+      return price( asset( effective_total_vesting_shares, liquid_symbol.get_paired_symbol() ), asset( effective_total_vesting_fund, liquid_symbol ) );
    }
 
-   price    get_reward_vesting_share_price() const
+   price get_reward_vesting_share_price() const
```
mvandeberg: The formula for reward vesting price for STEEM is:

```c++
return price( total_vesting_shares + pending_rewarded_vesting_shares,
   total_vesting_fund_steem + pending_rewarded_vesting_steem );
```

For an SMT it should be:

``` c++
return price( total_vesting_shares_ballast + total_vesting_shares + reward_vesting_shares,
   total_vesting_fund_ballast + total_vesting_fund_smt + reward_vesting_smt );
```

libraries/chain/include/steem/chain/smt_objects/smt_token_object.hpp
```diff
@@ -44,30 +44,34 @@ class smt_token_object : public object< smt_token_object_type, smt_token_object
...
    {
       share_type reward_vesting_shares = total_vesting_shares + pending_rewarded_vesting_shares;
       share_type reward_vesting_smt = total_vesting_fund_smt + pending_rewarded_vesting_smt;
 
       if( reward_vesting_shares == 0 || reward_vesting_smt == 0 )
```
mvandeberg: This check should be if the ballast are empty again. If for some reason there are no pending rewards but there are vesting SMT, then you'll return the wrong price.

libraries/chain/steem_evaluator.cpp
```diff
@@ -1222,61 +1223,109 @@ void withdraw_vesting_evaluator::do_apply( const withdraw_vesting_operation& o )
...
-      _db.modify( account, [&]( account_object& a ) {
-         a.vesting_withdraw_rate = asset( 0, VESTS_SYMBOL );
-         a.next_vesting_withdrawal = time_point_sec::maximum();
-         a.to_withdraw = 0;
-         a.withdrawn = 0;
-      });
+            if( new_vesting_withdraw_rate.amount * vesting_withdraw_intervals < o.vesting_shares.amount )
+               new_vesting_withdraw_rate.amount += 1;
+
+            FC_ASSERT( a.vesting_withdraw_rate != new_vesting_withdraw_rate, "This operation would not change the vesting withdraw rate." );
```
mvandeberg: Do the math and the assertion outside of the modify, if possible. It minimizes performance impact if the assertion fails.

libraries/chain/util/smt_token.cpp
```diff
@@ -79,16 +79,12 @@ share_type payout( database& db, const asset_symbol_type& symbol, const account_
    for ( auto& _asset : assets )
    {
       if ( _asset.symbol.is_vesting() )
-      {
```
mvandeberg: Is there a reason to remove these brackets? This specific remove make this code less readable and subtly changes the behavior here. It is not immediately clear to me why this should be changed.

libraries/chain/steem_evaluator.cpp
```diff
@@ -1222,61 +1223,109 @@ void withdraw_vesting_evaluator::do_apply( const withdraw_vesting_operation& o )
       return;
    }
 
+#ifdef STEEM_ENABLE_SMT
+   if ( o.vesting_shares.symbol.space() == asset_symbol_type::smt_nai_space )
+   {
+      FC_ASSERT( o.vesting_shares.symbol.is_vesting(), "Vesting shares must be a vesting symbol." );
```
mvandeberg: Add SMT Hardfork Assertion to be sure this code cannot execute before the hardfork. It would appear safe is as an account should not be able to have an SMT token balance prior, but better to be safe. It can be removed after the hardfork.

libraries/chain/steem_evaluator.cpp
```diff
@@ -1222,61 +1223,109 @@ void withdraw_vesting_evaluator::do_apply( const withdraw_vesting_operation& o )
...
+            a.vesting_withdraw_rate = new_vesting_withdraw_rate;
+            a.next_vesting_withdrawal = _db.head_block_time() + fc::seconds( SMT_VESTING_WITHDRAW_INTERVAL_SECONDS );
+            a.to_withdraw = o.vesting_shares.amount;
+            a.withdrawn = 0;
+         } );
+      }
    }
    else
+#endif
    {
```
mvandeberg: Due to the changes to `withdraw_vesting_operation::validate()`, you need to add an assertion in the else case that `o.vesting_shares.symbol == VESTS_SYMBOL`.

libraries/chain/util/smt_token.cpp
sgerbino: According to `validate_database_invariants` I needed to add vesting to the total supply on the token for it to add up.

mvandeberg: Got it, I understand now. This was probably written in an earlier vesting of the SMT scope prior to the SMT vesting destination.

Curious why this wasn't causing a failure earlier... Even awarding Vesting SMTs with a 1:1 ratio should have caused problems...

PR #3473, opened 2019-08-23
SMT Master Cleanup

### LINKED ISSUE #3461: Prepare master for SMT development
* Remove `STEEM_ENABLE_SMT` from `master` branch, all builds should have SMTs enabled from this point forward
* Add in all SMT tests to CI
* Remove extraneous builds and tests from CI docker image (we should be able to avoid builds without SMT now making CI times much faster)

sgerbino:
Resolves #3461.


Dockerfile
```diff
@@ -93,45 +93,6 @@ RUN \
     rm -rf /usr/local/src/steem/build ; \
     fi
 
-RUN \
```
mvandeberg: We can't get rid of this build. This is the build that makes the testnet binary.

libraries/protocol/include/steem/protocol/required_automated_actions.hpp
```diff
@@ -11,18 +11,14 @@ namespace steem { namespace protocol {
    typedef fc::static_variant<
 #ifdef IS_TEST_NET
             example_required_action
```
mvandeberg: Combine lines to move `,` to line 13.

mvandeberg: Also move the example required_action to the end of the static variant.

PR #3510, opened 2019-09-13
Tick Pricing Rules

### LINKED ISSUE #1573: Rationalizing Market Orders (was: Tick Pricing Rules)

sgerbino:
Resolves #1573.


libraries/protocol/steem_operations.cpp
```diff
@@ -564,20 +564,32 @@ namespace steem { namespace protocol {
...
-               || (
-                     amount_to_sell.symbol.space() == asset_symbol_type::smt_nai_space
-                     && is_asset_type( exchange_rate.quote, STEEM_SYMBOL )
-                  )
-               || (
-                     is_asset_type( amount_to_sell, STEEM_SYMBOL )
-                     && exchange_rate.quote.symbol.space() == asset_symbol_type::smt_nai_space
-                  ),
-               "Limit order must be for the STEEM:SBD or SMT:(STEEM/SBD) market" );
+      FC_ASSERT( amount_to_sell.symbol.is_vesting() == false, "Cannot create a limit order for vesting types." );
```
mvandeberg: These additions might be able to be skipped based on the changed you made in the evaluator...

mvandeberg: I tested locally and removing this entire block does not break the tests.

mvandeberg: I tested this locally, and it can be removed without breaking the tests.

libraries/protocol/steem_operations.cpp
```diff
@@ -564,20 +564,30 @@ namespace steem { namespace protocol {
...
-               || (
-                     amount_to_sell.symbol.space() == asset_symbol_type::smt_nai_space
-                     && is_asset_type( exchange_rate.quote, STEEM_SYMBOL )
-                  )
-               || (
-                     is_asset_type( amount_to_sell, STEEM_SYMBOL )
-                     && exchange_rate.quote.symbol.space() == asset_symbol_type::smt_nai_space
-                  ),
-               "Limit order must be for the STEEM:SBD or SMT:(STEEM/SBD) market" );
+      if ( amount_to_sell.symbol.space() == asset_symbol_type::legacy_space )
```
mvandeberg: It was this entire addition that is not needed.

PR #3516, opened 2019-09-23
Resolve new warnings introduced by Clang 8

sgerbino:
Alternatively we could disable warnings these warnings like so: `-Wno-error=shadow`


sgerbino:
Closing in favor of #3517

PR #3522, opened 2019-10-07
SMT Token Emissions

### LINKED ISSUE #2738: As a token holder, I want token inflation.
Implement token inflation events.

#2430 & #2429 contain discussions regarding emissions bounds.

Token emissions will be handled via optional automated actions. These actions will later be rejected via soft consensus in the RC plugin.

sgerbino:
Resolves #2738.


libraries/chain/optional_action_evaluator.cpp
```diff
@@ -1,11 +1,83 @@
...
+      }
+      else if( is_rewards( e.first ) )
+      {
+         reward_tokens += e.second;
+      }
+      else if( is_vesting( e.first ) )
+      {
+         vesting_tokens += e.second;
+      }
+      else if( is_vesting_type( e.first ) )
```
sgerbino: This should be `is_founder_vesting` and that should probably be renamed to `is_account_vesting` or something similar.

libraries/protocol/smt_util.cpp
```diff
@@ -102,6 +102,19 @@ account_name_type get_unit_target_account( const unit_target_type& unit_target )
    return unit_target_account;
 }
 
+bool is_valid_emissions_destination( const unit_target_type& unit_target )
+{
+   if ( is_valid_account_name( unit_target ) )
```
sgerbino: Should support `is_founder_vesting` or `is_accounting_vesting` if renamed.

libraries/plugins/token_emissions/token_emissions_plugin.cpp
```diff
@@ -0,0 +1,173 @@
...
+   else
+   {
+      _db.remove( emission_obj );
+   }
+}
+
+void token_emissions_impl::on_generate_optional_actions( const generate_optional_actions_notification& note )
+{
+   // We're creating token emissions for the upcoming block.
+   time_point_sec next_emission_time = _db.head_block_time() + STEEM_BLOCK_INTERVAL;
```
mvandeberg: If we are generating on the opt action signal, we shouldn't need to add a block interval.

PR #3555, opened 2019-11-14
Prevent token emissions from emitting passed the max token supply

### LINKED ISSUE #3528: Limit Token Emissions to the tokens max supply
The token emission plugin currently does not respect the tokens `max_supply`, this needs to be fixed.

sgerbino:
Resolves #3528.

### DIFF
--- a/libraries/chain/util/smt_token.cpp
+++ b/libraries/chain/util/smt_token.cpp
@@ -120,8 +120,8 @@ flat_map< unit_target_type, share_type > generate_emissions( const smt_token_obj
       {
          fc::uint128 lep_abs_val{ emission.lep_abs_amount.value },
                      rep_abs_val{ emission.rep_abs_amount.value },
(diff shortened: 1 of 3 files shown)


libraries/chain/util/smt_token.cpp
```diff
@@ -138,10 +138,13 @@ flat_map< unit_target_type, share_type > generate_emissions( const smt_token_obj
       else
          new_token_supply = std::max( abs_amount, rel_amount );
 
-      uint32_t unit_sum = emission.emissions_unit.token_unit_sum();
+      if ( token.current_supply + new_token_supply <= token.max_supply )
```
mvandeberg: This should cap the new tokens to max supply, rather than preventing it if the emissions go above.

For example, a low supply token that constantly replenishes the supply should be able to create the deficit of tokens on emission.

PR #3558, opened 2019-11-19
Clear null account balances

### LINKED ISSUE #3556: Clear null account SMT balance
The null account's balance of SMTs needs to be cleared periodically.

The null account balance for STEEM, SBD, and VESTS is checked every block. That is probably too often for SMTs. The only real impact it has on calculations is for emissions. Because of that, I propose tying in clearing of the null account balance in to emissions. Let's add an extension to the emissions operation that includes how much SMT and SMT power is being removed from the null account and the supply. The emissions operation should be generated to remove that amount, but accept any amount less than the balance of the null account so that the emissions action cannot fail by a user transferring to the null account after the creation of the action.

This creates a slight UX problem for SMTs without emissions as burned supply will not be reflected in current supplies. That is easily enough fixed by client libraries and UIs by looking up the null account balance for the SMT and doing the calculation client side.

sgerbino:
Resolves #3556.

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -1501,10 +1501,16 @@ void database::clear_null_account_balance()
    //////////////////////////////////////////////////////////////
 
    if( total_steem.amount > 0 )
+   {
+      wlog( "NOTIFYALERT! Adjusting total STEEM supply in database::clear_null_account_balance(). Asset: ${a}", ("a", total_steem) );
       adjust_supply( -total_steem );
+   }
 
    if( total_sbd.amount > 0 )
+   {
+      wlog( "NOTIFYALERT! Adjusting total SBD supply in database::clear_null_account_balance(). Asset: ${a}", ("a", total_sbd) );
       adjust_supply( -total_sbd );
+   }
 
    post_push_virtual_operation( vop_op );
 }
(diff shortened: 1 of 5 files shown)


libraries/chain/database.cpp
```diff
@@ -1490,21 +1457,37 @@ void database::clear_null_account_balance()
...
+      a.reward_sbd_balance.amount     = 0;
+      a.reward_vesting_balance.amount = 0;
+      a.reward_vesting_steem.amount   = 0;
+   });
+
    //////////////////////////////////////////////////////////////
 
    if( total_steem.amount > 0 )
+   {
+      wlog( "NOTIFYALERT! Adjusting total STEEM supply in database::clear_null_account_balance(). Asset: ${a}", ("a", total_steem) );
```
mvandeberg: These are going to spam during reindex.

libraries/chain/database.cpp
```diff
@@ -1490,21 +1457,37 @@ void database::clear_null_account_balance()
...
       adjust_supply( -total_steem );
+   }
 
    if( total_sbd.amount > 0 )
+   {
+      wlog( "NOTIFYALERT! Adjusting total SBD supply in database::clear_null_account_balance(). Asset: ${a}", ("a", total_sbd) );
```
mvandeberg: Same here

libraries/chain/database.cpp
```diff
@@ -1490,21 +1457,37 @@ void database::clear_null_account_balance()
...
-      {
-         a.reward_vesting_steem.amount = 0;
-         a.reward_vesting_balance.amount = 0;
-      });
    }
 
+   // We simply set all account balances to zero here because of the newly added
+   // assertions in the `adjust_*` family of functions that prevents negative deltas
+   // to the null account.
+   modify( null_account, [&]( account_object& a )
```
mvandeberg: Put this in a conditional `if( total_steem.amount > 0 || total_sbd.amount > 0 )` so that we do not do the modifications every block.

PR #3563, opened 2019-12-02
Multiphasic ICO

### LINKED ISSUE #3560: SMT Multiphasic ICOs
The SMT specification supports two phase ICO (A soft cap and a hard cap).

The Binance ICO was a three phase ICO.

It seems reasonable to assume that a hard limit of two ICOs may be restrictive for some SMTs. It is relatively straight forward to expand the existing soft cap system to support an arbitrary number of ICO phases.

The minimum required steem will remain and can have any number of phases prior or after and can be in the middle of a phase.

Phases can be configured in any order.

The last ICO phase is the implicit hard cap.

### LINKED ISSUE #3564: Relax SMT Emission Deletion
To delete an emission during setup, all attributes of the emission must match.

Let's relax the requirement to only require the schedule time to match as that is still a sufficient identifier for any emission.

sgerbino:
Resolves #3560.
Resolves #3564.


libraries/chain/include/steem/chain/util/smt_token.hpp
```diff
@@ -22,6 +22,9 @@ bool schedule_next_contributor_payout( database& db, const asset_symbol_type& a
 bool schedule_founder_payout( database& db, const asset_symbol_type& a );
 
 share_type payout( database& db, const asset_symbol_type& symbol, const account_object& account, const std::vector< contribution_payout >& payouts );
+fc::optional< share_type > steem_units_hard_cap( database& db, const asset_symbol_type& a );
+std::size_t tier_size( database& db, const asset_symbol_type& symbol );
```
mvandeberg: Maybe make these names a bit more descriptive? `ico_tier_size`, `remove_ico_tier_objects`?

libraries/chain/util/smt_token.cpp
```diff
@@ -428,6 +443,51 @@ bool schedule_founder_payout( database& db, const asset_symbol_type& a )
...
+   auto itr = ico_tier_idx.lower_bound( symbol );
+   while( itr != ico_tier_idx.end() && itr->symbol == symbol )
+   {
+      const auto& current = *itr;
+      ++itr;
+      db.remove( current );
+   }
+}
+
+std::size_t tier_size( database& db, const asset_symbol_type& symbol )
```
mvandeberg: This implementation is fine for a small ico tier cap, but we may want to make a note to track this in the ico object if we increase the number of ICO tiers allowed.

libraries/chain/smt_evaluator.cpp
```diff
@@ -161,32 +168,91 @@ void smt_setup_evaluator::do_apply( const smt_setup_operation& o )
...
+
+      FC_ASSERT( ito != nullptr,
+         "The specified ICO tier does not exist. Symbol: ${s}, Steem Units Cap: ${c}",
+         ("s", token.liquid_symbol)("c", o.steem_units_cap)
+      );
+
+      FC_ASSERT(
+         ito->symbol == o.symbol &&
+         ito->steem_units_cap == o.steem_units_cap &&
+         ito->capped_generation_policy.generation_unit.steem_unit == generation_policy.generation_unit.steem_unit &&
```
mvandeberg: Do we need to check the generation policy, or is a match on the symbol and cap sufficient enough? The reason to do so would be reduced impact on block size for removal of a generation policy.

libraries/chain/smt_evaluator.cpp
```diff
@@ -132,26 +132,33 @@ static void verify_accounts( database& db, const flat_map< unit_target_type, uin
    }
 }
 
-struct smt_setup_evaluator_visitor
+struct smt_generation_policy_validator
```
mvandeberg: For consistency in naming between validate / evaluate, let's rename this `smt_setup_ico_tier_evaluator_visitor`

libraries/chain/smt_evaluator.cpp
sgerbino: Symbol and cap is sufficient, but I thought it would be more consistent with `smt_token_emissions`, which require an exact match for removal.

libraries/chain/smt_evaluator.cpp
sgerbino: This is already consistent because the evaluator aprt is called `smt_generation_policy_visitor`. I chose to name it with way because the validate/evaluate is not particular to `smt_setup_ico_tier_operation`, it is for any `smt_generation_policy`.

PR #3581, opened 2020-01-16
Correctly clear null account balance only after HF 14

### LINKED ISSUE #3556: Clear null account SMT balance

### LINKED ISSUE #3580: Master does not replay
`master` fails reindex with error:

```
2056686ms database.cpp:3213             _apply_block         ] 10 assert_exception: Assert Exception
available >= -delta: Account anastacia does not have sufficient funds for balance adjustment. Required: {"amount":"-2362304","precision":3,"nai":"@@000000021"}, Available: {"amount":"2362303","precision":3,"nai":"@@000000021"}
    {"acc":"anastacia","r":{"amount":"-2362304","precision":3,"nai":"@@000000021"},"a":{"amount":"2362303","precision":3,"nai":"@@000000021"}}
    database.cpp:4554 adjust_balance

    {"op":{"type":"transfer_operation","value":{"from":"anastacia","to":"poloniex","amount":{"amount":"2362304","precision":3,"nai":"@@000000021"},"memo":"6f9cf0a77c0e5207"}}}
    database.cpp:3449 _apply_transaction

    {"trx":{"ref_block_num":46581,"ref_block_prefix":2983141536,"expiration":"2016-09-03T01:41:18","operations":[{"type":"transfer_operation","value":{"from":"anastacia","to":"poloniex","amount":{"amount":"2362304","precision":3,"nai":"@@000000021"},"memo":"6f9cf0a77c0e5207"}}],"extensions":[],"signatures":["1f5c16f5c75524a7e392c678e945c5f4cf1c61f1197681d516161cd2f028a5a5a31df150814b4c9669ce1fb9f26732868b9a389c3d3e91ec0ff4dc2f56f4e3f0ed"]}}
    database.cpp:3455 _apply_transaction
2056686ms database.cpp:3213             _apply_block         ] next_block.block_num(): 4634102 
```

The problem is from f61ca957. Prior to HF 14, no null account balances are cleared. The changed clear the balance unconditionally. HF 14 checks need to be added.

mvandeberg:
Closes #3580
Closes #3556

### DIFF
--- a/libraries/chain/database.cpp
+++ b/libraries/chain/database.cpp
@@ -4528,7 +4528,7 @@ void database::adjust_balance( const account_object& a, const asset& delta )
 
    bool check_balance = has_hardfork( STEEM_HARDFORK_0_20__1811 );
 
-   if ( a.name == STEEM_NULL_ACCOUNT )
+   if ( has_hardfork( STEEM_HARDFORK_0_14__327 ) && a.name == STEEM_NULL_ACCOUNT )
    {
       FC_ASSERT( delta.amount >= 0, "The null account cannot be withdrawn from." );
       adjust_supply( -delta );
@@ -4558,7 +4558,7 @@ void database::adjust_balance( const account_name_type& name, const asset& delta
 
    bool check_balance = has_hardfork( STEEM_HARDFORK_0_20__1811 );
 
-   if ( name == STEEM_NULL_ACCOUNT )
+   if ( has_hardfork( STEEM_HARDFORK_0_14__327 ) && name == STEEM_NULL_ACCOUNT )
    {
       FC_ASSERT( delta.amount >= 0, "The null account cannot be withdrawn from." );
       adjust_supply( -delta );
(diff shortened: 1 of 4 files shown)


tests/tests/smt_operation_time_tests.cpp
```diff
@@ -1530,6 +1530,7 @@ BOOST_AUTO_TEST_CASE( smt_token_emissions )
 
       BOOST_REQUIRE( approximately_equal( db->get_balance( db->get_account( "george" ), symbol ).amount, george_share ) );
       BOOST_REQUIRE( db->get_balance( STEEM_NULL_ACCOUNT, symbol ).amount == 0 );
+      idump( (token.current_supply)(supply) );
```
sgerbino: We can scrap this `idump`.

libraries/chain/database.cpp
```diff
@@ -5569,10 +5538,6 @@ void database::validate_smt_invariants()const
 
          if ( itr->reward_balance.amount > 0 )
             theMap[ itr->liquid_symbol ].liquid += itr->reward_balance;
-
-         // The @null account should have no balance for any particular Smart Media Token.
-         FC_ASSERT( get_balance( null_acct, itr->liquid_symbol ).amount == 0 );
```
sgerbino: We should keep these if we're immediately clearing null balance for SMTs.

PR #3582, opened 2020-01-19
Prevent emission to account that does not exists

drov0:
Add check on smt_setup_emissions_evaluator to prevent users from submitting emissions to non-existent accounts.

### DIFF
--- a/libraries/chain/smt_evaluator.cpp
+++ b/libraries/chain/smt_evaluator.cpp
@@ -279,15 +279,16 @@ void smt_setup_emissions_evaluator::do_apply( const smt_setup_emissions_operatio
             ("s", SMT_EMISSION_MIN_INTERVAL_SECONDS)("end", *end_time) );
       }
       
-      for ( const auto& e : o.emissions_unit.token_unit )
-      {
-          if (smt::unit_target::is_founder_vesting( e.first ))
-          {
+     for ( const auto& e : o.emissions_unit.token_unit )
+     {
+        if (smt::unit_target::is_account_name_type( e.first ))
+           {
               std::string name = e.first;
+              ilog(name);
               auto acc = _db.find<account_object, by_name>(name);
-              FC_ASSERT(acc != nullptr, "Invalid emission destiname, account ${a} must exist", ("a", name));
-          }
-      }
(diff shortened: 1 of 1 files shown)


libraries/chain/smt_evaluator.cpp
```diff
@@ -278,6 +278,17 @@ void smt_setup_emissions_evaluator::do_apply( const smt_setup_emissions_operatio
             "New token emissions must begin no earlier than ${s} seconds after the last emission. Last emission time: ${end}",
             ("s", SMT_EMISSION_MIN_INTERVAL_SECONDS)("end", *end_time) );
       }
+      
+     for ( const auto& e : o.emissions_unit.token_unit )
+     {
+        if (smt::unit_target::is_account_name_type( e.first ))
+           {
+              std::string name = e.first;
```
sgerbino: If it is a vesting account, we need to use the utility function `get_unit_target_account()` to parse the account name. See branch: `20200205-check-emission-accounts` for the unit test to verify.

libraries/chain/smt_evaluator.cpp
```diff
@@ -278,6 +278,17 @@ void smt_setup_emissions_evaluator::do_apply( const smt_setup_emissions_operatio
...
+              ilog(name);
```
sgerbino: Remove `ilog` message in operation.

libraries/chain/smt_evaluator.cpp
```diff
@@ -278,6 +278,17 @@ void smt_setup_emissions_evaluator::do_apply( const smt_setup_emissions_operatio
             "New token emissions must begin no earlier than ${s} seconds after the last emission. Last emission time: ${end}",
             ("s", SMT_EMISSION_MIN_INTERVAL_SECONDS)("end", *end_time) );
       }
+      
+     for ( const auto& e : o.emissions_unit.token_unit )
+     {
+        if (smt::unit_target::is_account_name_type( e.first ))
```
sgerbino: Try to keep spacing within conditionals consistent with the rest of the application.

sgerbino:
Just adding a note here.

We should add a positive test case for an existing account and vesting version, for instance:

`bob`
`$!bob.vesting`

sgerbino:
Community PR for #3590.

PR #3584, opened 2020-01-27
Revert "Clear null account balances"

mvandeberg:
Reverts steemit/steem#3558

This implementation is problematic. Let's revert it entirely and reimplement.


mvandeberg:
Included in #3581

PR #3586, opened 2020-02-05
Add smt desired_ticker

mvandeberg:


libraries/protocol/include/steem/protocol/asset_symbol.hpp
```diff
@@ -33,12 +33,20 @@
 
 #ifdef IS_TEST_NET
 
+#define VESTS_SYMBOL_STR "VESTS"
```
sgerbino: Looks like these defines are backwards

PR #3607, opened 2020-02-11
Enforce ICO STEEM_MAX_SHARE_SUPPLY

### LINKED ISSUE #3596: Verify ICO tiers do not surpass `STEEM_MAX_SHARE_SUPPLY`
It may be possible with a high enough unit ratio and with the correct generation policy that ICO tiers could create an ICO that would cause an SMT token supply to be greater than `STEEM_MAX_SHARE_SUPPLY`.

To fix this, we need to track a cumulative max token supply from ICO on the ICO object.

On each new ICO tier object, the running max total should be calculated and updated. The operation should be rejected if the total surpasses `STEEM_MAX_SHARE_SUPPLY`.

At most, this calculation should only require reading from the current ICO tier and a single overlapping tier (if there is overlap).

mvandeberg:
Closes #3596

### DIFF
--- a/tests/tests/smt_operation_tests.cpp
+++ b/tests/tests/smt_operation_tests.cpp
@@ -5198,5 +5198,117 @@ BOOST_AUTO_TEST_CASE( smt_setup_ico_tier_apply )
    FC_LOG_AND_RETHROW()
 }
 
+BOOST_AUTO_TEST_CASE( smt_max_share_ico )
+{ try {
+   BOOST_TEST_MESSAGE( "Testing SMT contribution refunds" );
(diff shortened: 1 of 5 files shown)


tests/tests/smt_operation_tests.cpp
```diff
@@ -5236,5 +5236,117 @@ BOOST_AUTO_TEST_CASE( smt_setup_ico_tier_apply )
    FC_LOG_AND_RETHROW()
 }
 
+BOOST_AUTO_TEST_CASE( smt_max_share_ico )
```
sgerbino: This new test needs to be added to the `CMakeLists.txt`.

PR #3612, opened 2020-02-15
Hf 23 Rising Sun

anthonyadavisii:
This is a proactive proposal to prevent the Steemit Inc Stake to be leveraged to facilitate a hostile takeover with the express intent to collapse our chain to be assimilated into the Tron blockchain. The intent is implicitly stated in the following article. It doesn't seem ambiguous in it terms to be perfectly frank. 

> Together, TRON Foundation and Steemit Inc. will look to create further value for their users and to augment their advancements in decentralized technologies, including **moving old STEEM token to a new TRON based STEEM token**, giveaways to the existing TRX users with the new TRON based STEEM token

https://medium.com/@TronFoundation/steemit-joining-tron-ecosystem-6ad66584d0b

I see this as the last ditch effort that the **current witnesses** may employ to preserve our blockchain. Since previous forks, I am not sure if the coding as @jnordberg originally implemented will still achieve the same end. With that in mind, I ask for my fellow Steem developers to please review and adjust accordingly.

I don't we want to wait for an AMA but rather we should act IMMEDIATELY! I understand some may have ethical concerns about this but urge you to consider the probable alternative we will face considering the text from Medium article. It is clear that the intention is to "dissolve" Steem

At a minimum, perhaps we can somehow disable witness voting and freeze the stake to afford our active and valuable community time to consider fair alternatives but, make no mistake, I don't assume we would be treated with the same decency. 

Let's not give them the opportunity. The time to act is now before the Rising Sun.

### DIFF
--- a/contrib/config-for-broadcaster.ini
+++ b/contrib/config-for-broadcaster.ini
@@ -11,7 +11,7 @@ log-logger = {"name":"p2p","level":"info","appender":"stderr"}
 # Plugin(s) to enable, may be specified multiple times
 plugin = webserver p2p json_rpc witness
 
-plugin = database_api witness_api condenser_api network_broadcast_api rc_api
+plugin = database_api condenser_api network_broadcast_api rc_api
 
 # Defines a range of accounts to track as a json pair ["from","to"] [from,to] Can be specified multiple times
 # account-history-track-account-range =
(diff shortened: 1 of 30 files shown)


techcoderx:
This is the old code that is not updated for the current HF22. The code changes is still on HF20 (approaching HF21).

anthonyadavisii:
> This is the old code that is not updated for the current HF22. The code changes in this PR is still on HF20 (approaching HF21).

Yeah I totally borked the PR on that. Started from scratch and should be good to go. Sure the witnesses got everything undercontrol after hearing them in PAL.

https://github.com/anthonyadavisii/steem

PR #3625, opened 2020-03-09
Proposal hf23 special power down steemit accounts

joticajulian:
My proposal consists in two points:
1. **Steemit Inc accounts lose the possibility to vote** (for witnesses, proposals and posts). This will be applied not in a softfork but in a hardfork, meaning that all witnesses will have written this rule in their code.
2. **Power down of 2 years only for Steemit Inc accounts**. Concretely 1 payment per week during 104 periods. This rule will have effect during 2.5 years, and after that it changes to the normal rule of 3 months of power down.

**The community** will be sure that this stake will not be used to vote for witnesses and compromise the governance. And they will be confortable that the power down of this stake will take much more time (8 times a normal power down) to not affect a lot the price in a hypothetical dump.

**Justin Sun/Steemit Inc** will have freedom to manage his stake, but he cannot use these tokens to influence the governance, and the power down will take more time  (see that during the power down, each week he will less than 1% of the initial stake).

[**More details in this post**](http://steempeak.com/@jga/proposal-hf23-special-power-down-steemit-accounts)

### DIFF
--- a/libraries/chain/sps_evaluator.cpp
+++ b/libraries/chain/sps_evaluator.cpp
@@ -74,8 +74,11 @@ void update_proposal_votes_evaluator::do_apply( const update_proposal_votes_oper
    {
       FC_ASSERT( _db.has_hardfork( STEEM_PROPOSALS_HARDFORK ), "Proposals functionality not enabled until hardfork ${hf}", ("hf", STEEM_PROPOSALS_HARDFORK) );
 
-      const auto& account = _db.get_account( o.voter );
-      FC_ASSERT( account.can_vote, "Account has declined its voting rights." );
+      if( _db.has_hardfork( STEEM_HARDFORK_0_23 ) )
+      {
+         const auto& account = _db.get_account( o.voter );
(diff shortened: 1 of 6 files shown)


libraries/chain/sps_evaluator.cpp
```diff
@@ -74,6 +74,9 @@ void update_proposal_votes_evaluator::do_apply( const update_proposal_votes_oper
    {
       FC_ASSERT( _db.has_hardfork( STEEM_PROPOSALS_HARDFORK ), "Proposals functionality not enabled until hardfork ${hf}", ("hf", STEEM_PROPOSALS_HARDFORK) );
 
+      const auto& account = _db.get_account( o.voter );
+      FC_ASSERT( account.can_vote, "Account has declined its voting rights." );
```
Smittyvb: This should be wrapped in a hardfork check, otherwise an account with declined voting rights who attempts to vote for an SPS proposal before this hardfork takes effect would cause issues.

joticajulian: Good point @Smittyvb. I fixed it.
Thanks for reviewing the code.

PR #3628, opened 2020-04-02
add some seednode

Yrp:
Delete some unavailable seednodes
Add some available seednodes

### DIFF
--- a/libraries/protocol/include/steem/protocol/config.hpp
+++ b/libraries/protocol/include/steem/protocol/config.hpp
@@ -44,7 +44,7 @@
 
 #else // IS LIVE STEEM NETWORK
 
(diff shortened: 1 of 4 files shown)


libraries/protocol/include/steem/protocol/config.hpp
```diff
@@ -44,7 +44,7 @@
 
 #else // IS LIVE STEEM NETWORK
 
-#define STEEM_BLOCKCHAIN_VERSION              ( version(0, 22, 1) )
+#define STEEM_BLOCKCHAIN_VERSION              ( version(0, 22, 5) )
```
Smittyvb, line 47: Why are you trying to change the blockchain version? It doesn't need to be bumped for a change like this.

PR #3700, opened 2025-07-05
Add Ubuntu 22.04 Support for steemd

DoctorLai:
This is the second PR of the DAO Project [Proposal: Fixing Steemd Build Dependencies in the Latest OS](https://steemit.com/sps/@justyy/proposal-fixing-steemd-build-dependencies-in-the-latest-os). Previous PR: #3699 

This PR adds support for Ubuntu 22.04 in steemd, and also resolves compatibility issues introduced for Ubuntu 20.04 as a result of these changes.

The following functionality has been tested and verified on both Ubuntu 20.04 and 22.04:
1. Block generation
2. Block catch-up
3. RPC port access via curl test
4. steemd restart behavior

image

Progress Reports:
- [Proposal Progress: Built and Tested OK on Ubuntu22.04, However...](https://steemit.com/witness-activities/@justyy/proposal-progress-built-and-tested-ok-on-ubuntu22-04-however)
- [Proposal Progress: Latest Changes Break Ubuntu20.04](https://steemit.com/witness-activities/@justyy/proposal-progress-latest-changes-break-ubuntu20-04)
- [Proposal Progress: secp256k1 may break the build](https://steemit.com/witness-activities/@justyy/proposal-progress-secp256k1-may-break-the-build)
- [Proposal Progress: Added Ubuntu 22.04 Support for steemd](https://steemit.com/witness-activities/@justyy/proposal-progress-added-ubuntu-22-04-support-for-steemd)


only-dev-time:
Is there a reason why you absolutely want to use Boost 1.78? Is this the preparation for the final version?
I have also successfully built `steemd` with the standard version 1.71 in Ubuntu20.04. Maybe it would also works with 1.74 in Ubuntu22.04.

DoctorLai:
> Is there a reason why you absolutely want to use Boost 1.78? Is this the preparation for the final version? I have also successfully built `steemd` with the standard version 1.71 in Ubuntu20.04. Maybe it would also works with 1.74 in Ubuntu22.04.

The asio library have been fixed a few problems on 1.78 - but actually, I am thinking we can try even the latest version 1.88. 

In ubuntu20.04 PR - I 've also updated the submodules (websockpp, rocksdb etc) to the latest versions, they don't work on 1.71 as far as I remember.

This PR is not ready yet.. (current progress 92%)

machinemaker007:
1) The get_segment_manager function in the chainbase.hpp file is not implemented yet
2) Why are const removed from many parameters, for example: const __m128i* -> __128i*, The const type is more appropriate. 
3) Why is the type of the parameter outlen in blake related functions changed from uint8_t to size_t, it seems uint8_t is ok.

DoctorLai:
> 1. The get_segment_manager function in the chainbase.hpp file is not implemented yet
> 2. Why are const removed from many parameters, for example: const __m128i* -> __128i*, The const type is more appropriate.
> 3. Why is the type of the parameter outlen in blake related functions changed from uint8_t to size_t, it seems uint8_t is ok.

@machinemaker007  Thanks for reviewing. The lib equihash has been updated to latest see: https://github.com/khovratovich/equihash   
The previous version does not compile (and it is too old).  It is not a complete re-use of the library and hence I didn't spend time like last PR to make it a submodule.
Not sure what you meant by `get_segment_manager`

libraries/chainbase/include/chainbase/chainbase.hpp
```diff
@@ -994,9 +995,7 @@ namespace chainbase {
          }
 
 #ifndef ENABLE_MIRA
-         auto get_segment_manager() -> decltype( ((bip::managed_mapped_file*)nullptr)->get_segment_manager()) {
-            return _segment->get_segment_manager();
-         }
+         auto get_segment_manager() -> decltype(std::declval<bip::managed_mapped_file>().get_segment_manager());
```
machinemaker007: @DoctorLai Why get_segment_manager function's implementation is removed ?

DoctorLai: @machinemaker007  I've removed this - as this line is not used at all in both "ENABLE_MIRA" on/off.

PR #3708, opened 2026-01-19
Ingest plugin

ety001:


only-dev-time:
Nice idea, I hope you have a usecase for this.

I haven't run it or tested it yet. From what I've seen in the code, there doesn't seem to be any differentiation for replay or resync cases. That would slow down the process considerably when the plugin is activated.
In addition, documentation on the code and the usage of the plugin would be necessary.

ety001:
> Nice idea, I hope you have a usecase for this.
> 
> I haven't run it or tested it yet. From what I've seen in the code, there doesn't seem to be any differentiation for replay or resync cases. That would slow down the process considerably when the plugin is activated. In addition, documentation on the code and the usage of the plugin would be necessary.

I'm currently refactoring the SteemDB project. In order to accelerate the database recovery speed during the cold start, I developed this plugin. This plugin directly retrieves the required data from the block_log.

Currently, the plugin is still under testing, and the relevant test code is here.

https://github.com/steemit/steemdb/tree/next/steemdb-sync/test/docker-compose

